PR #1, opened 2019-07-15
Initial proposal for webgpu.h

Kangz:
@kvark PTAL we have to decide on a license (tentatively 3clause bsd for this repo) and the "authors" name.

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -36,6 +36,8 @@ typedef struct WGPUDeviceImpl* WGPUDevice;
 typedef struct WGPUFenceImpl* WGPUFence;
 typedef struct WGPUPipelineLayoutImpl* WGPUPipelineLayout;
 typedef struct WGPUQueueImpl* WGPUQueue;
+typedef struct WGPURenderBundleImpl* WGPURenderBundle;
+typedef struct WGPURenderBundleEncoderImpl* WGPURenderBundleEncoder;
 typedef struct WGPURenderPassEncoderImpl* WGPURenderPassEncoder;
 typedef struct WGPURenderPipelineImpl* WGPURenderPipeline;
 typedef struct WGPUSamplerImpl* WGPUSampler;
@@ -184,6 +186,13 @@ typedef enum WGPUTextureAspect {
     WGPUTextureAspect_Force32 = 0x7FFFFFFF
 } WGPUTextureAspect;
 
+typedef enum WGPUTextureComponentType {
+    WGPUTextureComponentType_Float = 0x00000000,
+    WGPUTextureComponentType_Sint = 0x00000001,
+    WGPUTextureComponentType_Uint = 0x00000002,
+    WGPUTextureComponentType_Force32 = 0x7FFFFFFF
+} WGPUTextureComponentType;
+
 typedef enum WGPUTextureDimension {
     WGPUTextureDimension_1D = 0x00000000,
     WGPUTextureDimension_2D = 0x00000001,
@@ -196,60 +205,54 @@ typedef enum WGPUTextureFormat {
     WGPUTextureFormat_R8Snorm = 0x00000001,
     WGPUTextureFormat_R8Uint = 0x00000002,
     WGPUTextureFormat_R8Sint = 0x00000003,
-    WGPUTextureFormat_R16Unorm = 0x00000004,
-    WGPUTextureFormat_R16Snorm = 0x00000005,
-    WGPUTextureFormat_R16Uint = 0x00000006,
-    WGPUTextureFormat_R16Sint = 0x00000007,
-    WGPUTextureFormat_R16Float = 0x00000008,
-    WGPUTextureFormat_RG8Unorm = 0x00000009,
-    WGPUTextureFormat_RG8Snorm = 0x0000000A,
-    WGPUTextureFormat_RG8Uint = 0x0000000B,
-    WGPUTextureFormat_RG8Sint = 0x0000000C,
-    WGPUTextureFormat_R32Float = 0x0000000D,
-    WGPUTextureFormat_R32Uint = 0x0000000E,
-    WGPUTextureFormat_R32Sint = 0x0000000F,
-    WGPUTextureFormat_RG16Unorm = 0x00000010,
-    WGPUTextureFormat_RG16Snorm = 0x00000011,
-    WGPUTextureFormat_RG16Uint = 0x00000012,
-    WGPUTextureFormat_RG16Sint = 0x00000013,
-    WGPUTextureFormat_RG16Float = 0x00000014,
-    WGPUTextureFormat_RGBA8Unorm = 0x00000015,
-    WGPUTextureFormat_RGBA8UnormSrgb = 0x00000016,
-    WGPUTextureFormat_RGBA8Snorm = 0x00000017,
-    WGPUTextureFormat_RGBA8Uint = 0x00000018,
-    WGPUTextureFormat_RGBA8Sint = 0x00000019,
-    WGPUTextureFormat_BGRA8Unorm = 0x0000001A,
-    WGPUTextureFormat_BGRA8UnormSrgb = 0x0000001B,
-    WGPUTextureFormat_RGB10A2Unorm = 0x0000001C,
-    WGPUTextureFormat_RG11B10Float = 0x0000001D,
-    WGPUTextureFormat_RG32Float = 0x0000001E,
-    WGPUTextureFormat_RG32Uint = 0x0000001F,
-    WGPUTextureFormat_RG32Sint = 0x00000020,
-    WGPUTextureFormat_RGBA16Unorm = 0x00000021,
-    WGPUTextureFormat_RGBA16Snorm = 0x00000022,
-    WGPUTextureFormat_RGBA16Uint = 0x00000023,
-    WGPUTextureFormat_RGBA16Sint = 0x00000024,
-    WGPUTextureFormat_RGBA16Float = 0x00000025,
-    WGPUTextureFormat_RGBA32Float = 0x00000026,
-    WGPUTextureFormat_RGBA32Uint = 0x00000027,
-    WGPUTextureFormat_RGBA32Sint = 0x00000028,
-    WGPUTextureFormat_Depth32Float = 0x00000029,
-    WGPUTextureFormat_Depth24Plus = 0x0000002A,
-    WGPUTextureFormat_Depth24PlusStencil8 = 0x0000002B,
-    WGPUTextureFormat_BC1RGBAUnorm = 0x0000002C,
-    WGPUTextureFormat_BC1RGBAUnormSrgb = 0x0000002D,
-    WGPUTextureFormat_BC2RGBAUnorm = 0x0000002E,
-    WGPUTextureFormat_BC2RGBAUnormSrgb = 0x0000002F,
-    WGPUTextureFormat_BC3RGBAUnorm = 0x00000030,
-    WGPUTextureFormat_BC3RGBAUnormSrgb = 0x00000031,
-    WGPUTextureFormat_BC4RUnorm = 0x00000032,
-    WGPUTextureFormat_BC4RSnorm = 0x00000033,
-    WGPUTextureFormat_BC5RGUnorm = 0x00000034,
-    WGPUTextureFormat_BC5RGSnorm = 0x00000035,
-    WGPUTextureFormat_BC6HRGBUfloat = 0x00000036,
-    WGPUTextureFormat_BC6HRGBSfloat = 0x00000037,
-    WGPUTextureFormat_BC7RGBAUnorm = 0x00000038,
-    WGPUTextureFormat_BC7RGBAUnormSrgb = 0x00000039,
+    WGPUTextureFormat_R16Uint = 0x00000004,
+    WGPUTextureFormat_R16Sint = 0x00000005,
+    WGPUTextureFormat_R16Float = 0x00000006,
+    WGPUTextureFormat_RG8Unorm = 0x00000007,
+    WGPUTextureFormat_RG8Snorm = 0x00000008,
+    WGPUTextureFormat_RG8Uint = 0x00000009,
+    WGPUTextureFormat_RG8Sint = 0x0000000A,
+    WGPUTextureFormat_R32Float = 0x0000000B,
+    WGPUTextureFormat_R32Uint = 0x0000000C,
+    WGPUTextureFormat_R32Sint = 0x0000000D,
+    WGPUTextureFormat_RG16Uint = 0x0000000E,
+    WGPUTextureFormat_RG16Sint = 0x0000000F,
+    WGPUTextureFormat_RG16Float = 0x00000010,
+    WGPUTextureFormat_RGBA8Unorm = 0x00000011,
+    WGPUTextureFormat_RGBA8UnormSrgb = 0x00000012,
+    WGPUTextureFormat_RGBA8Snorm = 0x00000013,
+    WGPUTextureFormat_RGBA8Uint = 0x00000014,
+    WGPUTextureFormat_RGBA8Sint = 0x00000015,
+    WGPUTextureFormat_BGRA8Unorm = 0x00000016,
+    WGPUTextureFormat_BGRA8UnormSrgb = 0x00000017,
+    WGPUTextureFormat_RGB10A2Unorm = 0x00000018,
+    WGPUTextureFormat_RG11B10Float = 0x00000019,
+    WGPUTextureFormat_RG32Float = 0x0000001A,
+    WGPUTextureFormat_RG32Uint = 0x0000001B,
+    WGPUTextureFormat_RG32Sint = 0x0000001C,
+    WGPUTextureFormat_RGBA16Uint = 0x0000001D,
+    WGPUTextureFormat_RGBA16Sint = 0x0000001E,
+    WGPUTextureFormat_RGBA16Float = 0x0000001F,
+    WGPUTextureFormat_RGBA32Float = 0x00000020,
+    WGPUTextureFormat_RGBA32Uint = 0x00000021,
+    WGPUTextureFormat_RGBA32Sint = 0x00000022,
+    WGPUTextureFormat_Depth32Float = 0x00000023,
+    WGPUTextureFormat_Depth24Plus = 0x00000024,
+    WGPUTextureFormat_Depth24PlusStencil8 = 0x00000025,
+    WGPUTextureFormat_BC1RGBAUnorm = 0x00000026,
+    WGPUTextureFormat_BC1RGBAUnormSrgb = 0x00000027,
+    WGPUTextureFormat_BC2RGBAUnorm = 0x00000028,
+    WGPUTextureFormat_BC2RGBAUnormSrgb = 0x00000029,
+    WGPUTextureFormat_BC3RGBAUnorm = 0x0000002A,
+    WGPUTextureFormat_BC3RGBAUnormSrgb = 0x0000002B,
+    WGPUTextureFormat_BC4RUnorm = 0x0000002C,
+    WGPUTextureFormat_BC4RSnorm = 0x0000002D,
+    WGPUTextureFormat_BC5RGUnorm = 0x0000002E,
+    WGPUTextureFormat_BC5RGSnorm = 0x0000002F,
+    WGPUTextureFormat_BC6HRGBUfloat = 0x00000030,
+    WGPUTextureFormat_BC6HRGBSfloat = 0x00000031,
+    WGPUTextureFormat_BC7RGBAUnorm = 0x00000032,
+    WGPUTextureFormat_BC7RGBAUnormSrgb = 0x00000033,
     WGPUTextureFormat_Force32 = 0x7FFFFFFF
 } WGPUTextureFormat;
 
@@ -360,6 +363,7 @@ typedef struct WGPUBindGroupLayoutBinding {
     bool dynamic;
     bool multisampled;
     WGPUTextureViewDimension textureDimension;
+    WGPUTextureComponentType textureComponentType;
 } WGPUBindGroupLayoutBinding;
 
 typedef struct WGPUBlendDescriptor {
@@ -444,6 +448,18 @@ typedef struct WGPURasterizationStateDescriptor {
     float depthBiasSlopeScale;
     float depthBiasClamp;
 } WGPURasterizationStateDescriptor;
+
+typedef struct WGPURenderBundleDescriptor {
+    void const * nextInChain;
+} WGPURenderBundleDescriptor;
+
+typedef struct WGPURenderBundleEncoderDescriptor {
+    void const * nextInChain;
+    uint32_t colorFormatsCount;
+    WGPUTextureFormat const * colorFormats;
+    WGPUTextureFormat const * depthStencilFormat;
+    uint32_t sampleCount;
+} WGPURenderBundleEncoderDescriptor;
 
 typedef struct WGPURenderPassDepthStencilAttachmentDescriptor {
     WGPUTextureView attachment;
@@ -654,6 +670,7 @@ typedef WGPUComputePipeline (*WGPUProcDeviceCreateComputePipeline)(WGPUDevice de
 typedef WGPURenderPipeline (*WGPUProcDeviceCreateRenderPipeline)(WGPUDevice device, WGPURenderPipelineDescriptor const * descriptor);
 typedef WGPUPipelineLayout (*WGPUProcDeviceCreatePipelineLayout)(WGPUDevice device, WGPUPipelineLayoutDescriptor const * descriptor);
 typedef WGPUQueue (*WGPUProcDeviceCreateQueue)(WGPUDevice device);
+typedef WGPURenderBundleEncoder (*WGPUProcDeviceCreateRenderBundleEncoder)(WGPUDevice device, WGPURenderBundleEncoderDescriptor const * descriptor);
 typedef WGPUSampler (*WGPUProcDeviceCreateSampler)(WGPUDevice device, WGPUSamplerDescriptor const * descriptor);
 typedef WGPUShaderModule (*WGPUProcDeviceCreateShaderModule)(WGPUDevice device, WGPUShaderModuleDescriptor const * descriptor);
 typedef WGPUTexture (*WGPUProcDeviceCreateTexture)(WGPUDevice device, WGPUTextureDescriptor const * descriptor);
@@ -668,6 +685,20 @@ typedef void (*WGPUProcQueueSubmit)(WGPUQueue queue, uint32_t commandCount, WGPU
 typedef void (*WGPUProcQueueSignal)(WGPUQueue queue, WGPUFence fence, uint64_t signalValue);
 typedef WGPUFence (*WGPUProcQueueCreateFence)(WGPUQueue queue, WGPUFenceDescriptor const * descriptor);
 
+// Procs of RenderBundleEncoder
+typedef void (*WGPUProcRenderBundleEncoderSetPipeline)(WGPURenderBundleEncoder renderBundleEncoder, WGPURenderPipeline pipeline);
+typedef void (*WGPUProcRenderBundleEncoderSetBindGroup)(WGPURenderBundleEncoder renderBundleEncoder, uint32_t groupIndex, WGPUBindGroup group, uint32_t dynamicOffsetCount, uint64_t const * dynamicOffsets);
+typedef void (*WGPUProcRenderBundleEncoderDraw)(WGPURenderBundleEncoder renderBundleEncoder, uint32_t vertexCount, uint32_t instanceCount, uint32_t firstVertex, uint32_t firstInstance);
+typedef void (*WGPUProcRenderBundleEncoderDrawIndexed)(WGPURenderBundleEncoder renderBundleEncoder, uint32_t indexCount, uint32_t instanceCount, uint32_t firstIndex, int32_t baseVertex, uint32_t firstInstance);
+typedef void (*WGPUProcRenderBundleEncoderDrawIndirect)(WGPURenderBundleEncoder renderBundleEncoder, WGPUBuffer indirectBuffer, uint64_t indirectOffset);
+typedef void (*WGPUProcRenderBundleEncoderDrawIndexedIndirect)(WGPURenderBundleEncoder renderBundleEncoder, WGPUBuffer indirectBuffer, uint64_t indirectOffset);
+typedef void (*WGPUProcRenderBundleEncoderInsertDebugMarker)(WGPURenderBundleEncoder renderBundleEncoder, char const * groupLabel);
+typedef void (*WGPUProcRenderBundleEncoderPopDebugGroup)(WGPURenderBundleEncoder renderBundleEncoder);
+typedef void (*WGPUProcRenderBundleEncoderPushDebugGroup)(WGPURenderBundleEncoder renderBundleEncoder, char const * groupLabel);
+typedef void (*WGPUProcRenderBundleEncoderSetVertexBuffers)(WGPURenderBundleEncoder renderBundleEncoder, uint32_t startSlot, uint32_t count, WGPUBuffer const * buffers, uint64_t const * offsets);
+typedef void (*WGPUProcRenderBundleEncoderSetIndexBuffer)(WGPURenderBundleEncoder renderBundleEncoder, WGPUBuffer buffer, uint64_t offset);
+typedef WGPURenderBundle (*WGPUProcRenderBundleEncoderFinish)(WGPURenderBundleEncoder renderBundleEncoder, WGPURenderBundleDescriptor const * descriptor);
+
 // Procs of RenderPassEncoder
 typedef void (*WGPUProcRenderPassEncoderSetPipeline)(WGPURenderPassEncoder renderPassEncoder, WGPURenderPipeline pipeline);
 typedef void (*WGPUProcRenderPassEncoderSetBindGroup)(WGPURenderPassEncoder renderPassEncoder, uint32_t groupIndex, WGPUBindGroup group, uint32_t dynamicOffsetCount, uint64_t const * dynamicOffsets);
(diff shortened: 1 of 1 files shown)


webgpu.h
```diff
@@ -0,0 +1,1025 @@
...
+
+typedef struct wgpuPipelineLayoutDescriptor {
+    const void* nextInChain;
+    wgpuStructureType structureType;
+    uint32_t bindGroupLayoutCount;
+    wgpuBindGroupLayout const * bindGroupLayouts;
+} wgpuPipelineLayoutDescriptor;
+
+typedef struct wgpuPipelineStageDescriptor {
+    const void* nextInChain;
```
kvark: can we be consistent here, i.e. `void const * nextInChain` ?

webgpu.h
```diff
@@ -0,0 +1,1025 @@
...
+    wgpuBlendDescriptor alphaBlend;
+    wgpuBlendDescriptor colorBlend;
+    wgpuColorWriteMask writeMask;
+} wgpuColorStateDescriptor;
+
+typedef struct wgpuComputePipelineDescriptor {
+    const void* nextInChain;
+    wgpuStructureType structureType;
+    wgpuPipelineLayout layout;
+    wgpuPipelineStageDescriptor const * computeStage;
```
kvark: why is this a pointer?

webgpu.h
```diff
@@ -0,0 +1,1025 @@
...
+    wgpuStructureType structureType;
+    wgpuTextureFormat format;
+    wgpuBlendDescriptor alphaBlend;
+    wgpuBlendDescriptor colorBlend;
+    wgpuColorWriteMask writeMask;
+} wgpuColorStateDescriptor;
+
+typedef struct wgpuComputePipelineDescriptor {
+    const void* nextInChain;
+    wgpuStructureType structureType;
```
kvark: Do we really need these in the base descriptors? We could say that the structure types here are always what WebGPU version X.Y defines. We could require only those "nextInChain" things to provide the structure types.

webgpu.h
```diff
@@ -0,0 +1,1025 @@
...
+typedef struct wgpuVertexBufferDescriptor {
+    uint64_t stride;
+    wgpuInputStepMode stepMode;
+    uint32_t attributeCount;
+    wgpuVertexAttributeDescriptor const * attributes;
+} wgpuVertexBufferDescriptor;
+
+typedef struct wgpuRenderPassDescriptor {
+    uint32_t colorAttachmentCount;
+    const wgpuRenderPassColorAttachmentDescriptor* const * colorAttachments;
```
kvark: why is this a pointer to a pointer, as opposed to just a pointer to the first element in an array?

webgpu.h
```diff
@@ -0,0 +1,1025 @@
...
+    wgpuIndexFormat indexFormat;
+    uint32_t bufferCount;
+    wgpuVertexBufferDescriptor const * buffers;
+} wgpuVertexInputDescriptor;
+
+typedef struct wgpuRenderPipelineDescriptor {
+    const void* nextInChain;
+    wgpuStructureType structureType;
+    wgpuPipelineLayout layout;
+    wgpuPipelineStageDescriptor const * vertexStage;
```
kvark: why is this a pointer?

webgpu.h
```diff
@@ -0,0 +1,1025 @@
...
+    wgpuPipelineStageDescriptor const * fragmentStage;
+    wgpuVertexInputDescriptor const * vertexInput;
```
kvark: why is this a pointer? (looks like a larger discussion is pending here, I'll stop noting these)

webgpu.h
```diff
@@ -0,0 +1,1025 @@
...
+                                           uint64_t dataLength,
+                                           void* userdata);
+typedef void (*wgpuFenceOnCompletionCallback)(wgpuFenceCompletionStatus status, void* userdata);
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// Procs of BindGroup
+typedef void (*wgpuProcBindGroupReference)(wgpuBindGroup bindGroup);
```
kvark: Does this have to be a part of the API header? Looks like an implementation detail to me.

webgpu.h
```diff
@@ -0,0 +1,1025 @@
...
+    WGPU_VERTEX_FORMAT_UINT4 = 0x00000019,
+    WGPU_VERTEX_FORMAT_INT = 0x0000001A,
+    WGPU_VERTEX_FORMAT_INT2 = 0x0000001B,
+    WGPU_VERTEX_FORMAT_INT3 = 0x0000001C,
+    WGPU_VERTEX_FORMAT_INT4 = 0x0000001D,
+    WGPU_VERTEX_FORMAT_FORCE32 = 0x7FFFFFFF
+} wgpuVertexFormat;
+
+typedef enum {
+    WGPU_BUFFER_USAGE_BIT_NONE = 0x00000000,
```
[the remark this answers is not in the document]
Kangz: Because AFAIK that's the best practice, for example Vulkan has this (although now I realize that they use the VkFlags type):

```c
typedef enum VkImageUsageFlagBits {
    VK_IMAGE_USAGE_TRANSFER_SRC_BIT = 0x00000001,
    VK_IMAGE_USAGE_TRANSFER_DST_BIT = 0x00000002,
    VK_IMAGE_USAGE_SAMPLED_BIT = 0x00000004,
    VK_IMAGE_USAGE_STORAGE_BIT = 0x00000008,
    VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT = 0x00000010,
    VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT = 0x00000020,
    VK_IMAGE_USAGE_TRANSIENT_ATTACHMENT_BIT = 0x00000040,
    VK_IMAGE_USAGE_INPUT_ATTACHMENT_BIT = 0x00000080,
    VK_IMAGE_USAGE_FLAG_BITS_MAX_ENUM = 0x7FFFFFFF
} VkImageUsageFlagBits;
typedef VkFlags VkImageUsageFlags;
```

Would you like the header to use the same scheme as Vulkan with a `wgpuFlags` type?

webgpu.h
Kangz: Done

webgpu.h
Kangz: Because this descriptor needs to be extended and could be swapped for a differently typed descriptor (linked to [1])

webgpu.h
Kangz: Because this descriptor needs to be extended and could be swapped for a differently typed descriptor (linked to [1])

webgpu.h
Kangz: Because this descriptor needs to be extended and could be swapped for a differently typed descriptor (linked to [1])

webgpu.h
Kangz: Because this descriptor needs to be extended and could be swapped for a differently typed descriptor (linked to [1])

webgpu.h
Kangz: [1]

The `structureType` is useful in the base descriptor because it allows extensions to provide a completely alternative descriptor for some functions / parent descriptors. It also allows making WebGPU 2 descriptors by just making structureType be `WGPU_STRUCTURE_TYPE_WEBGPU2_DEFAULT`.

All your comments below about things being pointers are valid cosmetic concerns that happen when we require base descriptors to be swappable like this, but I think it is important in the long run. (like Vulkan forgot to do this for some descriptors and had to add them back in some VK_KHR_maintenance extensions, search for "2KHR" in vulkan.h)

webgpu.h
Kangz: Every WebGPU implementation will need to implement something like this to avoid freeing data in use by the GPU, and having a refcount helps composing together software that uses webgpu.h

webgpu.h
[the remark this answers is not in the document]
kvark: ah, hmm. We *do* need some way to disambiguate real "enum" things from bitflags, so it sounds like we need both a "Flags" suffix and a separate `wgpuFlags` typedef...

webgpu.h
```diff
@@ -0,0 +1,1025 @@
...
+typedef struct wgpuShaderModuleImpl* wgpuShaderModule;
+typedef struct wgpuTextureImpl* wgpuTexture;
+typedef struct wgpuTextureViewImpl* wgpuTextureView;
+
+typedef enum {
+    WGPU_ADDRESS_MODE_REPEAT = 0x00000000,
+    WGPU_ADDRESS_MODE_MIRROR_REPEAT = 0x00000001,
+    WGPU_ADDRESS_MODE_CLAMP_TO_EDGE = 0x00000002,
+    WGPU_ADDRESS_MODE_FORCE32 = 0x7FFFFFFF
+} wgpuAddressMode;
```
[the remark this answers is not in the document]
Kangz: Sure, what naming conventions would you like for enums, bitfields, structs, callbacks and API function pointers? We can discuss it here and it will be easy to change the code generator for this initial version.

webgpu.h
kvark: Ok, thanks for clarification.
Do *we* need to allow extensions to expect a completely different type? Do you have a good case in mind that would demonstrate why this is useful to us?

webgpu.h
```diff
@@ -0,0 +1,1025 @@
...
+    WGPU_TEXTURE_USAGE_BIT_SAMPLED = 0x00000004,
+    WGPU_TEXTURE_USAGE_BIT_STORAGE = 0x00000008,
+    WGPU_TEXTURE_USAGE_BIT_OUTPUT_ATTACHMENT = 0x00000010,
+    WGPU_TEXTURE_USAGE_BIT_FORCE32 = 0x7FFFFFFF
+} wgpuTextureUsageBit;
+
+
+typedef struct wgpuBindGroupBinding {
+    uint32_t binding;
+    wgpuBuffer buffer;
```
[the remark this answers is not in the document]
Kangz: `GPUBindingResource` is a sum type that doesn't exist in C/C++ (unions really don't work well for C++ and it implementing a wire with a union type would require doing BGL validation on the wire). Here all the various pieces of `GPUBindingResource` are inlined with exactly one of `buffer`, `sampler` and `textureView` being non-null to pass validation.

webgpu.h
kvark: We don't have those procs, and we don't free data in use by GPU.

webgpu.h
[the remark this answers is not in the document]
kvark: But C++ does have unions, and all we need is just to tag them.
Not sure what "BGL" is about for validation on the wire, but it sounds like an implementation detail?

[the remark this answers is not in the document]
Kangz: BGL is shorthand for Bind Group Layout.

Imagine we have the following:

```cpp
struct WGPUBufferBinding {
    WGPUBuffer buffer;
    uint64_t offset;
    uint64_t size;
};

union WGPUBindingResource {
    WGPUBufferBinding buffer;
    WGPUSampler sampler;
    WGPUTextureView textureView;
};

struct WGPUBingGroupBinding {
    uint32_t binding;
    WGPUBindingResource resource;
};
```

In a remoting implementation, the wire format is likely using number handles in different namespaces to serialize `WGPUSampler`, `WGPUBuffer` and `WGPUTextureView` so it can't just memcpy the `WGPUBindGroupBinding` on the wire and it needs to know what member of the union is the relevant one. So there are three choices I can see:

 1. The client of the remoting implementation keeps track of the `WGPUBindGroupLayout` so it can match `WGPUBindGroupBindings` with the corresponding entry and serialize the correct member of the union. This is really sad given how much the API is remotable without keeping state.
 2. A tag is added next to the union that describes which union member to use but this is not as nice, see below.
 3. The union fields are all inlined in `WGPUBindGroupBinding` with only one of the resources being allowed to be non-null (this is checked on the server side).

I'd strongly like to avoid 1) because it is too implicit (it risks crashes when the application uses the wrong resource type) and adds client-side state tracking. And I'd rather have 3) than 2) because the client code is much nicer:

```c
/* Option 2) */
WGPUBindGroupBinding binding = {
    .binding = 0,
    .bindingType = WGPUBindGroupBindingUnionType_Buffer
    .resource = {
        .buffer = {
            .buffer = myBuffer,
            .offset = myOffset,
            .size = mySize,
        }
    }
};

/* Option 3) */
WGPUBindGroupBinding binding = {
    .binding = 0,
    .buffer = myBuffer,
    .offset = myOffset,
    .size = mySize
};
```

webgpu.h
Kangz: I'm thinking about extensions that allow importing external resources, they could completely replace the descriptors with their own. Note that in C with designated initializers developers never need to specify the `WGPU_STRUCTURE_TYPE_DEFAULT` because it is 0 (and in C++ the structure type would have a default value).

webgpu.h
Kangz: Sorry I wasn't clear. I expect every WebGPU implementation will have internal refcounting to avoid freeing data on the GPU (and I see wgpu-rs have ref_count even though I'm not familiar enough with Rust to understand all of what's happening there). So why not expose this to the application since it helps compose things together?

This is also slightly closer to the JS semantics with GC keeping the object alive as long as there is a reference to it.

webgpu.h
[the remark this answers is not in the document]
Kangz: Do we care about potential collisions between bitflags and enums? The only one I can think of is if we add a `WGPUShaderStage` enum at some point.

webgpu.h
kvark: Ok, you want external resources. Can't use express them via the pNext chain? It doesn't look like it requires having the structure type in the base structs anyway.

webgpu.h
kvark: We haven't considered exposing the refcounts to the user:
  - for JS, it already knows when to call destroy()
  - for Rust, we use ownership and borrowing instead of refcounts on the API side

Perhaps, you could do it as an extension?

webgpu.h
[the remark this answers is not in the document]
kvark: It's just cleaner on the API side when you can instantly see if stuff is supposed to be added together, or only unique value is expected.

webgpu.h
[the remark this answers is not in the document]
kvark: Let's start the type names with WGPU, i.e. `WGPUAddressMode`. I know it's not pretty, but the alternatives seem to be worse.

webgpu.h
Kangz: Since WebGPU objects are internally synchronized doesn't it make sense to have them implement `Clone + Sync` and this would kinda require a refcount? (I don't know Rust too well so maybe this is stupid)

kvark: Only `Device` in WebGPU is truly internally synchronized.
In general, if something is internally synchronized, that means `Sync` is fine, but `Clone` has nothing to do with it. If a user wants to share the `Device` between multiple threads, they just wrap it in `Arc<Device>` and go from there.

webgpu.h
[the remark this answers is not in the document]
kvark: I was talking about option (2), but thank you for clearly describing this!
It looks like both unions and inlined representations are ugly, but the latter is at least more straightforward. Let's proceed with (3) then, we can work with it.

webgpu.h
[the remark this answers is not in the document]
Kangz: I thought you said the convention in Rust is to not have any suffix for bitfields? I think I'd be ok keeping the `Bit` suffix as such:
```c
typedef enum WGPUTextureUsageBit {
    WGPUTextureUsageBit_Foo = 0x00000001
    //    VK_IMAGE_USAGE_FLAG_BITS_MAX_ENUM = 0x7FFFFFFF
} WGPUTextureUsageBit;
typedef WGPUFlags WGPUTextureUsageBits;
```

and having the following in C++ (with special sauce bitmaskable enum templates):

```cpp
namespace wgpu {
enum class TextureUsageBit : uint32_t {
    Foo = 0x00000001
};
}
```

webgpu.h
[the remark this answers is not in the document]
Kangz: I'll give it a try.

webgpu.h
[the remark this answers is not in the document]
kvark: It's fine for rust because it's a separate type :) so you can't mix up enums with flags.
But for cases (like C) where it's the same type, Flags suffix appears necessary.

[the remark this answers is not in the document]
kvark: `WGPUTextureUsageBit_Foo` - can it be `WGPUTextureUsage_Foo` instead?

webgpu.h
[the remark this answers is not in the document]
Kangz: Updating the casing, PTAL. I think it looks ok.

webgpu.h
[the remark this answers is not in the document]
Kangz: I'm not sure I understand, you'd like to type to have `Bit` or `Bits` but not the enum names?. Can you write down the code you're thinking about for a sample bit flag?

webgpu.h
Kangz: I got confirmation from Chris Forbes that there is no Vulkan extension that replaces the initial struct (so far).

I don't have a super good case in mind but am still very slightly leaning in favor of having `structureType` for initial structure.

Pros:

 - Structures are more uniform and tooling doesn't need to know the difference between initial structures and not initial.
 - We can replace the initial structure maybe one day for an extension or Webgpu 2
 - (more like the absence of a cons): no added developer friction thanks to C designated initializers, C++ and Rust defaults.

Cons:

 - Slightly larger types and one additional validation check.
 - Makes a couple structures pointed require one more indirection.

webgpu.h
```diff
@@ -21,1002 +21,1004 @@
...
-    WGPU_TEXTURE_USAGE_BIT_OUTPUT_ATTACHMENT = 0x00000010,
-    WGPU_TEXTURE_USAGE_BIT_FORCE32 = 0x7FFFFFFF
-} wgpuTextureUsageBit;
+    WGPUTextureUsageBit_None = 0x00000000,
+    WGPUTextureUsageBit_CopySrc = 0x00000001,
+    WGPUTextureUsageBit_CopyDst = 0x00000002,
+    WGPUTextureUsageBit_Sampled = 0x00000004,
+    WGPUTextureUsageBit_Storage = 0x00000008,
+    WGPUTextureUsageBit_OutputAttachment = 0x00000010,
+    WGPUTextureUsageBit_Force32 = 0x7FFFFFFF
```
kvark: this is not a bit :)

webgpu.h
```diff
@@ -21,1002 +21,1004 @@
...
-    wgpuProcDeviceCreateSampler deviceCreateSampler;
-    wgpuProcDeviceCreateShaderModule deviceCreateShaderModule;
-    wgpuProcDeviceCreateTexture deviceCreateTexture;
-    wgpuProcDeviceCreateBufferMappedAsync deviceCreateBufferMappedAsync;
-    wgpuProcDeviceReference deviceReference;
-    wgpuProcDeviceRelease deviceRelease;
-
-    wgpuProcFenceGetCompletedValue fenceGetCompletedValue;
-    wgpuProcFenceOnCompletion fenceOnCompletion;
-    wgpuProcFenceRefe
```
kvark: can we move all the refcounting stuff into a separate header, even if it's hosted alongside webgpu.h?

webgpu.h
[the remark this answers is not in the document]
kvark: The shortest approach would be
```cpp
typedef enum WGPUTextureUsage {
    WGPUTextureUsage_Foo = 0x00000001
} WGPUTextureUsage;
typedef WGPUFlags WGPUTextureUsageFlags;
```

Do you think that could work for us?

webgpu.h
kvark: Thanks for investigation!

I don't think there is a compelling evidence about the pros here:
  - tooling work isn't too complex - just a matter of an extra bit that you know beforehand
  - replacing an initial structure is not used in Vulkan, as you mentioned, so we'll unlikely need it either. I don't see how anything we'd want couldn't be done via `pNext` anyway?
  - Rust doesn't have defaults for *fields*.

Think about all the redundancy and clutter this would be adding to the header and docs... How about we come back to this question when it's time for V1 release?

webgpu.h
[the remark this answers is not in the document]
Kangz: Ah thanks for the explanation, yeah I think that would work, I'll change it.

webgpu.h
Kangz: Sounds good. Note that this remove the need for `WGPUStructureType` at this point so I'll remove it.

webgpu.h
Kangz: Will be fixed when the flags match what you described :)

webgpu.h
Kangz: Unified them instead to not be per device but on `WGPUObject` which is `void*`

kvark:
Thank you for addressing my concerns!
The last one I got unresolved is - I think the header doesn't need all the logic related to the proc tables and the refcounting. Dawn should be just putting this into a separate header.
@grovesNL - would you have a moment to check if you spot any issues?

kvark:
Had a bit of discussion on the call about refcount exposure.
Pros:
  - both implementations already do that, exposing costs us nothing
  - some user would be happy to use that instead of wrapping staff in `Arc` and such

Cons:
  - destruction is slightly more complicated: there are still explicit destroy_() methods on textures and buffers, but actual handles only go out of scope upon refcount decrease
  - WASM shim would have to emulate refcounts (which means it would affect all the WASM clients?!).

I'm slightly concerned about the "exposing costs us nothing" aspect: it's not like we can straight expose our internal refcounts. Our implementation would likely still need to track the user refcounts separately from the internal ones (because we'd need to know when the user one reaches 0, at least), at which point we end up with just implementing an *extra* refcounting exposed to the user, which is nothing like "exposing costs us nothing" implies.

webgpu.h
```diff
@@ -0,0 +1,861 @@
...
+typedef WGPURenderPipeline (*WGPUProcDeviceCreateRenderPipeline)(WGPUDevice device, WGPURenderPipelineDescriptor const * descriptor);
+typedef WGPUPipelineLayout (*WGPUProcDeviceCreatePipelineLayout)(WGPUDevice device, WGPUPipelineLayoutDescriptor const * descriptor);
+typedef WGPUQueue (*WGPUProcDeviceCreateQueue)(WGPUDevice device);
+typedef WGPUSampler (*WGPUProcDeviceCreateSampler)(WGPUDevice device, WGPUSamplerDescriptor const * descriptor);
+typedef WGPUShaderModule (*WGPUProcDeviceCreateShaderModule)(WGPUDevice device, WGPUShaderModuleDescriptor const * descriptor);
+typedef WGPUTexture (*WGPUProcDeviceCreateTexture)(WGPUDevice device, WGPUTextureDescriptor const * descriptor);
+typedef void (*WGPUProcDeviceCreateBufferMappedAsync)(WGPUDevice device, WGPUBufferDescriptor const * descriptor, WGPUBufferCreateMappedCallback callback, void * userdata);
+
+// Procs of Fence
+typedef uint64_t (*WGPUProcFenceGetCompletedValue)(WGPUFence fence);
```
grovesNL: Are fences exposed intentionally?

webgpu.h
```diff
@@ -0,0 +1,861 @@
...
+WGPU_EXPORT void wgpuRenderPassEncoderSetScissorRect(WGPURenderPassEncoder renderPassEncoder, uint32_t x, uint32_t y, uint32_t width, uint32_t height);
+WGPU_EXPORT void wgpuRenderPassEncoderSetVertexBuffers(WGPURenderPassEncoder renderPassEncoder, uint32_t startSlot, uint32_t count, WGPUBuffer const * buffers, uint64_t const * offsets);
+WGPU_EXPORT void wgpuRenderPassEncoderSetIndexBuffer(WGPURenderPassEncoder renderPassEncoder, WGPUBuffer buffer, uint64_t offset);
+WGPU_EXPORT void wgpuRenderPassEncoderEndPass(WGPURenderPassEncoder renderPassEncoder);
+
+// Methods of Texture
+WGPU_EXPORT WGPUTextureView wgpuTextureCreateDefaultView(WGPUTexture texture);
+WGPU_EXPORT WGPUTextureView wgpuTextureCreateView(WGPUTexture texture, WGPUTextureViewDescriptor const * descriptor);
+WGPU_EXPORT void wgpuTextureDestroy(WGPUTexture texture);
+
```
grovesNL: I think we're missing instance creation and platform-specific surface creation

webgpu.h
```diff
@@ -0,0 +1,861 @@
...
+    WGPUTextureFormat format;
+    WGPUBlendDescriptor alphaBlend;
+    WGPUBlendDescriptor colorBlend;
+    WGPUColorWriteMaskFlags writeMask;
+} WGPUColorStateDescriptor;
+
+typedef struct WGPUComputePipelineDescriptor {
+    void const * nextInChain;
+    WGPUPipelineLayout layout;
+    WGPUPipelineStageDescriptor const * computeStage;
```
grovesNL: This could probably be `WGPUPipelineStageDescriptor computeStage` to match `vertexStage`

Kangz: Done

webgpu.h
Kangz: Yes, it is [part of the WebGPU API](https://gpuweb.github.io/gpuweb/#fence).

webgpu.h
Kangz: Yep, that's because this is something where I expect the native API to be significantly different from the Web API and would need its own PR and discussions. Dawn and wgpu-rs might want to expose diverging features, for example Dawn provides a list of all the (API, adapter) combinations it can support which wgpu-rs might not want to do since AFAIK you compile only one backend at a time.

Kangz:
> I'm slightly concerned about the "exposing costs us nothing" aspect: it's not like we can straight expose our internal refcounts. Our implementation would likely still need to track the user refcounts separately from the internal ones (because we'd need to know when the user one reaches 0, at least), at which point we end up with just implementing an extra refcounting exposed to the user, which is nothing like "exposing costs us nothing" implies.

Why do you need to keep the internal refcount separate from the internal one? We were doing this initially in Dawn but merged both refcounts because with RAII wrappers the application can't mistakenly deref too much.

webgpu.h
grovesNL: Alright, sounds good

kvark:
@Kangz 
> Why do you need to keep the internal refcount separate from the internal one? We were doing this initially in Dawn but merged both refcounts because with RAII wrappers the application can't mistakenly deref too much.

How do you rely on RAII wrappers if it's the C API being exposed?

Kangz:
We have a semi-idiomatic C++ wrapper for the C API, you can find Dawn's generated version [here](https://cs.chromium.org/chromium/src/out/Debug/gen/third_party/dawn/dawn/dawncpp.h?q=dawncpp&sq=package:chromium&g=0&l=16). We'd eventually contribute a similar wrapper to this repo.

kvark:
Sure, wrappers in high level languages can have this guarantee provided. But we are exposing the C API, we aren't *requiring* the users to use the wrappers. In the C API we can't guarantee that the issued refcount bumps/decs are sane.

Kangz:
What's the goal of having separate refcounts? I don't think we should validate the refcounting: if an application does too many derefs and the internal refcount is 0, then there will be a use-after-free and a potential segfault. We can't validate derefs and guarantee 100% that the application can't segfault itself on the first invalid deref. So you might as well have a single internal+external refcount. (in Dawn I'm thinking that maybe in debug we could split them, but even then that's only useful for automated testing).

kvark:
It comes do to what is the "Valid Usage" of the API encoded in the header. Providing a malformed pointer in place of a struct is surely a crash, but at least it's very localized: you know exactly when and how things went wrong. The "validitity" of a refcount decrease operation isn't local: it depends on all the previous ups and downs issued. So it's possible to describe, but it's much less desirable to expose an API that is *almost* safe, minus this one case where you'd not be able to tell instantly what the problem is (because it's somewhere deep in the logic of your app, out of a sudden).

Of all the build targets we need to support, look at where exposing refcounts is useful:
  1. Debug native builds - not useful, since the implementations would want to track user refcounts separately.
  2. Release native builds - may be useful.
  3. Web builds - not useful, since WASM shim would have to be implementing them.

Weighting these concerns, I feel like exposing the refcounts would bring more trouble than they are worth?

Kangz:
> It comes do to what is the "Valid Usage" of the API encoded in the header.

I don't think we should strive for Rust's level of safety in a C API because it doesn't have any support for linear types that would allow encoding such constraints.

> The "validitity" of a refcount decrease operation isn't local: it depends on all the previous ups and downs issued. So it's possible to describe, but it's much less desirable to expose an API that is almost safe, minus this one case where you'd not be able to tell instantly what the problem is (because it's somewhere deep in the logic of your app, out of a sudden).

Same thing when there is no refcounting happening: if you use a destroyed object, you might still have correct behavior (or a validation error instead of a crash) because the object is still internally referenced by the WebGPU implementation.

> Debug native builds - not useful, since the implementations would want to track user refcounts separately.

Graphics APIs are optimized for the Release usecases so I don't think we should worry about this.

> Web builds - not useful, since WASM shim would have to be implementing them.

The shim would have to implement the refcount, but it doesn't mean it's useless. The same arguments about refcounts being useful for software composability apply.

Kangz:
Removed the procTable as was requested, and wrapped the API function declarations in a block so they can be skipped. We probably want a getProcAddress-like function in a follow-up PR.

I tried implementing the "unified" refcount today to get Dawn closer to the proposed header. Turns out unified refcount is worse than having a reference/release for each object because it essentially forces implementations to do a dynamic dispatch based on the object type when `WGPUReference` makes teh refcount reach 0, which might not be idiomatic C or Rust.

For the WASM shim this isn't a problem because AFAIK the reference types are all pointing into a single reference table so objects are already unified there. The cases where we would like to do some type-specific work is if we want to call `GPUTexture/GPUBuffer.destroy` to free GPU memory when the refcount reaches 0.

Then the question is do we want refcount reaching 0, (or explicit destroy) to act as if `GPUBuffer/GPUTexture.destroy` is called? I think we want to separate the two concepts for several reasons:

 - The Javascript API separates the freeing of GPU memory from the lifetime of the state-tracking object: if you lose the last reference to a texture you can still submit command buffers using it, irrespective of whether your `GPUTexture` has been garbage collected or not.
 - This actually means that an API with only combined destruction of the state tracking and freeing of the GPU memory cannot be used to implement WebGPU in a Web engine unless it keeps track of the GPU object graph on the content process side (a big no-no for our implementation).
 - Often-enough people will want to create WGPU objects on the fly for the commands they are recording (like a tiny UBO, ad-hoc sampler or bindgroup) and they don't want to have a wgpu_native `PendingResource`-like structure on their side to keep them alive until the `wgpuQueueSubmit` call.

So irrespective of refcount vs. no refcount, we should split the CPU object lifetime management from the equivalent of `GPUBuffer/GPUTexture.destroy`.

Speaking of Web engines, they will need some form of refcounting of WGPU objects as soon as they implement sharing objects between workers.

kainino0x:
> For the WASM shim this isn't a problem because AFAIK the reference types are all pointing into a single reference table

The table is implemented in JS at least for now, so we can implement it however we want. Emscripten's GL object tables are per-type.

kvark:
Thanks for the extensive responses!

> I don't think we should strive for Rust's level of safety in a C API

It's not as much about safety as it is about a general "good API". When the user gets undesired behavior, a good API makes it obvious to what needs to be fixed. That's what the "locality" gives you: user searches for a place where they destroy the object and try to figure out if it's not the right place to destroy.

> if you use a destroyed object, you might still have correct behavior (or a validation error instead of a crash) because the object is still internally referenced by the WebGPU implementation.

We should strive to hide the implementation details from the user, if we want the API to be portable. The way I see it, if the user calls `destroy_xxx(handle)` - this has a clear (implementation-independent!) semantics of the `handle` being lost. Any use of `handle` by the user is a definite validation error after that. There is no room for crashes or "may work" scenarios.

> Graphics APIs are optimized for the Release usecases so I don't think we should worry about this.

The question here is not about optimization. It's about what API we expose for lifetime management.

> The shim would have to implement the refcount, but it doesn't mean it's useless. The same arguments about refcounts being useful for software composability apply.

The shim should be as light as possible though. Users who don't need refcounting (or want them implemented differently, say, by introducing weak references, thread-local counters, etc) shouldn't be paying the implementation cost for them. If we implement refcounting in the WASM shim, will it be strictly better than anything the users may do on their side?

As for the reasons to separate "destroy" from refcounting semantics, this one is least clear to me:
 >This actually means that an API with only combined destruction of the state tracking and freeing of the GPU memory cannot be used to implement WebGPU in a Web engine unless it keeps track of the GPU object graph on the content process side (a big no-no for our implementation).

I'm confused because there appears to be some context implied here that needs to be specified, e.g. for "combined destruction of the state tracking and freeing of the GPU memory".

Let me try to bring some clarity. About "destroy" - it doesn't make sense to actually destroy anything at this stage, since the implementation may have dependent work in flight. Forcing the user to know whether there is work in flight (e.g. with fences) is an option, but that's not the direction WebGPU group has taken. So "destroy" simply means "I don't want this object any more, please free associated resources as soon as you can". This is pretty much exactly the same semantics as reaching the refcount of 0.

So at the end of the day, we are trying to figure out what of the two APIs would work better:
  1) `destroy_xxx(), destroy_yyy(), destory_zzz()` methods that logically invalidate the handles
  2) `add_ref_xxx(), remove_ref_xxx(), add_ref_yyy(), remove_ref_yyy(), add_ref_zzz(), remove_ref_zzz()` with an optional `destroy_xxx()` to match WebGPU API.

The (2) can't be specified to invalidate the handles at any point, since the condition of reaching 0 is run-time dependent. The (2) has more APIs, potentially more redundant ones ("destroy" versus "remove_ref"), which are making it more difficult for the users to deal with lifetime issues (and correspondingly easier to leak objects/memory).

kainino0x:
FWIW, I noticed that refcounting allows the API to more closely match WebGPU:

```
const t = device.createTexture(...);
commands.push(makeSomeCommands(t));
// t.destroy() here would cause submit to fail later

...

queue.submit(commands);
// no need to destroy
```

```c
WGPUTexture t = wgpuDeviceCreateTexture(device, ...);
commands.push_back(makeSomeCommands(t));
wgpuTextureRelease(t);
// Destroying here would cause submit to fail later, but OK to release
// because command buffer still holds a ref
...

wgpuQueueSubmit(queue, commands.size(), commands.data());
```

```c
WGPUTexture t = wgpuDeviceCreateTexture(device, ...);
commands.push_back(makeSomeCommands(t));
// Destroying here would cause submit to fail later

...

wgpuQueueSubmit(queue, commands.size(), commands.data());
wgpuTextureDestroy(t);
```

kainino0x:
This is under the assumption that destroy destroys eagerly (which I think it is supposed to).

One way to change around that would be to just have Release on everything (no Reference), and then on textures and buffers, also have Destroy which destroys eagerly.

Kangz:
Removed all forms of refcounting and destroy so we can talk about it in a follow up. @kvark PTAL again.

webgpu.h
```diff
@@ -0,0 +1,776 @@
...
+                                           uint64_t dataLength,
+                                           void* userdata);
+typedef void (*WGPUFenceOnCompletionCallback)(WGPUFenceCompletionStatus status, void* userdata);
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// Procs of Buffer
+typedef void (*WGPUProcBufferUnmap)(WGPUBuffer buffer);
```
kvark: Since there is no ability to get a proc address, these typedefs aren't useful. Let's not add them right now?

Kangz: Removed (for now)

Kangz:
Updated to have render bundles and other WebGPU changes. Removed procs.

As far as I can tell there are no more contentious changes here so I'll go ahead and merge it.

webgpu.h
Kangz: Using `Arc` forces the way refcounting is done, which doesn't match what Dawn is doing (it's C++ and the refcount is at an offset while it is at offset 0 for `Arc`) nor what `wgpu-native` because of the `HUB` structure they use.

That said the same `Arc` semantics could be done with a `Sync + Clone` object that uses `webgpu.h`'s refcounting mechanisms to implement `Clone`.

PR #3, opened 2019-08-23
Add WGPU procs and WGPUGetProcAddress.

Kangz:
Having a GetProcAddress mechanism will be important to support
extensions that expose new functions. Procs are defined for all
entry-points because from experience in EGL_KHR_get_all_proc_addresses
developers want to sometimes be able to query all functions through
GetProcAddress.

WGPUGetProcAddress takes an optional WGPUDevice argument so that, like
in Vulkan, the procs can be specialized for a given device.

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -633,6 +633,107 @@ typedef void (*WGPUBufferMapWriteCallback)(WGPUBufferMapAsyncStatus status,
                                            void* userdata);
 typedef void (*WGPUFenceOnCompletionCallback)(WGPUFenceCompletionStatus status, void* userdata);
 
+typedef void (*WGPUProc)();
+
+#if !defined(WGPU_SKIP_PROCS)
+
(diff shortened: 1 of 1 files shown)


kvark:
> Having a GetProcAddress mechanism will be important to support
extensions that expose new functions.

It would help me if we had a more concrete (strawman) extension to discuss.

AFAIK, the `GetProcAddress` is in Vulkan because it was designed from the start as a layered API: so that there could be 3rd party layers between the app and the driver, such as the validation layers. Whether WebGPU-native should be layered or not is an open question that I don't think we discussed.

Let me think a bit on how we can use that...

Kangz:
For example an extension could add a new methods to encoders like `RenderPassEncoder::MultiDrawIndirect` or new objects with a corresponding `Device` creation method like `ResourceAliasHeap` we discussed in the group previously.

PR #4, opened 2019-08-24
Add enums and function definitions for error handling

austinEng:
Based on https://github.com/gpuweb/gpuweb/blob/master/design/ErrorHandling.md
Note that the device lost callback and uncaptured error callback are separate to match the Javascript API.

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -121,7 +121,7 @@ typedef enum WGPUBufferMapAsyncStatus {
     WGPUBufferMapAsyncStatus_Success = 0x00000000,
     WGPUBufferMapAsyncStatus_Error = 0x00000001,
     WGPUBufferMapAsyncStatus_Unknown = 0x00000002,
-    WGPUBufferMapAsyncStatus_ContextLost = 0x00000003,
+    WGPUBufferMapAsyncStatus_DeviceLost = 0x00000003,
     WGPUBufferMapAsyncStatus_Force32 = 0x7FFFFFFF
 } WGPUBufferMapAsyncStatus;
 
@@ -144,11 +144,27 @@ typedef enum WGPUCullMode {
     WGPUCullMode_Force32 = 0x7FFFFFFF
 } WGPUCullMode;
 
+typedef enum WGPUErrorFilter {
+    WGPUErrorFilter_None = 0x00000000,
+    WGPUErrorFilter_Validation = 0x00000001,
+    WGPUErrorFilter_OutOfMemory = 0x00000002,
+    WGPUErrorFilter_Force32 = 0x7FFFFFFF
+} WGPUErrorFilter;
+
+typedef enum WGPUErrorType {
+    WGPUErrorType_NoError = 0x00000000,
+    WGPUErrorType_Validation = 0x00000001,
+    WGPUErrorType_OutOfMemory = 0x00000002,
+    WGPUErrorType_Unknown = 0x00000003,
+    WGPUErrorType_DeviceLost = 0x00000004,
+    WGPUErrorType_Force32 = 0x7FFFFFFF
+} WGPUErrorType;
+
 typedef enum WGPUFenceCompletionStatus {
     WGPUFenceCompletionStatus_Success = 0x00000000,
     WGPUFenceCompletionStatus_Error = 0x00000001,
     WGPUFenceCompletionStatus_Unknown = 0x00000002,
-    WGPUFenceCompletionStatus_ContextLost = 0x00000003,
+    WGPUFenceCompletionStatus_DeviceLost = 0x00000003,
     WGPUFenceCompletionStatus_Force32 = 0x7FFFFFFF
 } WGPUFenceCompletionStatus;
 
@@ -660,7 +676,9 @@ typedef void (*WGPUBufferMapWriteCallback)(WGPUBufferMapAsyncStatus status,
                                            void* data,
                                            uint64_t dataLength,
                                            void* userdata);
+typedef void (*WGPUDeviceLostCallback)(const char* message, void* userdata);
 typedef void (*WGPUFenceOnCompletionCallback)(WGPUFenceCompletionStatus status, void* userdata);
+typedef void (*WGPUErrorCallback)(WGPUErrorType type, const char* message, void* userdata);
 
 // Methods of Buffer
 WGPU_EXPORT void wgpuBufferUnmap(WGPUBuffer buffer);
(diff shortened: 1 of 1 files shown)


webgpu.h
```diff
@@ -144,6 +144,22 @@ typedef enum WGPUCullMode {
...
 
+typedef enum WGPUErrorFilter {
+    WGPUErrorFilter_None = 0x00000000,
+    WGPUErrorFilter_Validation = 0x00000001,
+    WGPUErrorFilter_OutOfMemory = 0x00000002,
+    WGPUErrorFilter_Force32 = 0x7FFFFFFF
+} WGPUErrorFilter;
+
+typedef enum WGPUErrorType {
+    WGPUErrorType_None = 0x00000000,
```
kainino0x: Maybe "NoError" for clarity

webgpu.h
```diff
@@ -144,6 +144,22 @@ typedef enum WGPUCullMode {
...
+    WGPUErrorType_Validation = 0x00000001,
+    WGPUErrorType_OutOfMemory = 0x00000002,
+    WGPUErrorType_Unknown = 0x00000003,
+    WGPUErrorType_ContextLost = 0x00000004,
```
kainino0x: I'm not sure this is actually useful. It's only needed because the DeviceLost callback currently receives one. But I don't think we would ever call it with any other value (even WGPUErrorType_None) - we should always call this callback even if the device is being intentionally destroyed. Basically the DeviceLost callback would be a "device destroyed for any reason" callback.

On the Web, we might not want to resolve `device.lost` if the page is in the middle of shutdown. But I think that logic should be at the browser level.

webgpu.h
```diff
@@ -704,6 +721,10 @@ WGPU_EXPORT WGPUSampler wgpuDeviceCreateSampler(WGPUDevice device, WGPUSamplerDe
 WGPU_EXPORT WGPUShaderModule wgpuDeviceCreateShaderModule(WGPUDevice device, WGPUShaderModuleDescriptor const * descriptor);
 WGPU_EXPORT WGPUTexture wgpuDeviceCreateTexture(WGPUDevice device, WGPUTextureDescriptor const * descriptor);
 WGPU_EXPORT void wgpuDeviceCreateBufferMappedAsync(WGPUDevice device, WGPUBufferDescriptor const * descriptor, WGPUBufferCreateMappedCallback callback, void * userdata);
+WGPU_EXPORT void wgpuDeviceSetDeviceLostCallback(WGPUDevice device, WGPUErrorCallback callback, void * userdata);
```
Kangz: unit: sort alphabetically.

webgpu.h
```diff
@@ -144,6 +144,22 @@ typedef enum WGPUCullMode {
     WGPUCullMode_Force32 = 0x7FFFFFFF
 } WGPUCullMode;
 
+typedef enum WGPUErrorFilter {
```
Kangz: Maybe this should be a bitfield?

webgpu.h
Kangz: So just to be sure that enum value would disappear in favor of a specific callback type for device destroyed that's like a Rust `FnOnce() -> ()`? SGTM. Alternatively or in addition we could have a getter for device loss.

This also means that for testing we'll need the equivalent of WebGL's "lose context" extension.

Kangz: Mmmmh good point but that doesn't match the [ErrorHandling](https://github.com/gpuweb/gpuweb/blob/master/design/ErrorHandling.md) explainer. Maybe that can be a small divergence between this header and JS. Translation would do something like this:

 - JS API on C API: ignore device lost (or validate the lost handler is called before any device lost error).
 - C API on JS API: every popErrorScope checks whether the device is lost before calling the callback.

Kangz: Ah, rejection is like the context lost so it makes sense to keep the enum.

However maybe rename context lost to device lost because we aren't in WebGL? ^^

webgpu.h
Kangz: SGTM

webgpu.h
austinEng: We also have
> If there are no error scopes on the stack, `popErrorScope()` rejects.

If we just call the callback with `ERROR_TYPE_VALIDATION`, then for JS on C, then an error due to an empty stack looks the same as a validation error contained within a scope.

Maybe we should have `popErrorScope` return a success boolean?

kainino0x: hm, for memory management it is still probably more convenient for us to call that callback. I guess we could just use UNKNOWN?

In JS we could change it to throw an exception if you pop an empty stack, and the best way to expose that would probably be by returning a bool. If we did that then native apps would have to either do a little extra memory management or just ASSERT and allow it to leak in release.

austinEng: UNKNOWN would work, but the naming is a bit cryptic.

webgpu.h
```diff
@@ -144,11 +144,27 @@ typedef enum WGPUCullMode {
...
+    WGPUErrorFilter_None = 0x00000000,
+    WGPUErrorFilter_Validation = 0x00000001,
+    WGPUErrorFilter_OutOfMemory = 0x00000002,
+    WGPUErrorFilter_Force32 = 0x7FFFFFFF
```
kvark: These `Force32` things are quite unsound. Can we remove them?

webgpu.h
```diff
@@ -144,11 +144,27 @@ typedef enum WGPUCullMode {
...
+} WGPUErrorFilter;
+
+typedef enum WGPUErrorType {
+    WGPUErrorType_NoError = 0x00000000,
```
kvark: Do we need a "NoError" variant at all for an error type? That doesn't sound right.

webgpu.h
kainino0x: Hm, I thought this happened already in the original PR.

Doing this would mean splitting off the typedefs from the enums:

```c
typedef uint32_t WPUEnum;
```
```c
enum {
    WGPUErrorFilter_None = 0x00000000,
    WGPUErrorFilter_Validation = 0x00000001,
    WGPUErrorFilter_OutOfMemory = 0x00000002,
    WGPUErrorFilter_Force32 = 0x7FFFFFFF
}
typedef WGPUEnum WGPUErrorFilter;
```

, which looks OK to me.

Better to do it in another PR though.

webgpu.h
kainino0x: Yeah, I think it's necessary. Maybe we should rename this enum? It's used when the callback is called but an error has not occurred. In C it makes sense to guarantee all once-callbacks will be called exactly once (even if there's no error), in case the caller allocates memory that it wants to free after the callback runs.

webgpu.h
kvark: It doesn't look like `Force32` gives you what you need anyway. It doesn't guarantee that the compiler chooses 32 bits here.
I agree about discussing it in a different PR though

webgpu.h
kvark: ok, something like `WGPUErrorType_None = 0` is fine then

webgpu.h
Kangz: Whoops somehow I thought I had fixed flags to be to something we agreed to. I must have messed something up. I'll revisit that change.

Kangz: Ah, you'd like this for enums as well? I don't think that's a great idea because it loses even more typing information that's useful for code completion for example.

kainino0x: In C I don't think there is actually any typing information here: https://godbolt.org/z/rwMsAV
but maybe some autocompletion engines would use it anyway.

kvark: I'm not following. Removing "Force32" loses even more typing?

kainino0x: I think the typing loss comes from changing the type of WGPUErrorFilter from `enum WGPUErrorFilter` to `uint32_t`.

kvark: Sad indeed, although:
  1. you are going to be mostly using the API via a C++ wrapper, it's just unfortunate to require to to duplicate all the enums
  2. how else would we be able to force it to 32 bits?

webgpu.h
kainino0x: Just remembered, it's also needed for popErrorScope if there was no error inside the scope (where in JS we resolve to null).

webgpu.h
Kangz: `Force32` doesn't guarantee 32bit but that's the mechanism Vulkan uses already. I'll put up a PR and we can ask some folks that would use WebGPU in C (@floooh ?)

kvark: oh snap. So all vulkan headers are technically UB in terms of enum usage ? yikes

PR #5, opened 2019-08-29
Follow WebGPU IDL changes for texture view creation

Kangz:
GPUTexture.createDefaultView has been removed and more defaults have
been added to GPUTextureViewDescriptor where some members can be left
unspecified to get some automatic value.

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -246,68 +246,70 @@ typedef enum WGPUTextureDimension {
 } WGPUTextureDimension;
 
 typedef enum WGPUTextureFormat {
-    WGPUTextureFormat_R8Unorm = 0x00000000,
-    WGPUTextureFormat_R8Snorm = 0x00000001,
-    WGPUTextureFormat_R8Uint = 0x00000002,
-    WGPUTextureFormat_R8Sint = 0x00000003,
-    WGPUTextureFormat_R16Uint = 0x00000004,
-    WGPUTextureFormat_R16Sint = 0x00000005,
-    WGPUTextureFormat_R16Float = 0x00000006,
-    WGPUTextureFormat_RG8Unorm = 0x00000007,
-    WGPUTextureFormat_RG8Snorm = 0x00000008,
-    WGPUTextureFormat_RG8Uint = 0x00000009,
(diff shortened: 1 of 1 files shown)


webgpu.h
```diff
@@ -282,6 +282,7 @@ typedef enum WGPUTextureFormat {
     WGPUTextureFormat_BC6HRGBSfloat = 0x00000031,
     WGPUTextureFormat_BC7RGBAUnorm = 0x00000032,
     WGPUTextureFormat_BC7RGBAUnormSrgb = 0x00000033,
+    WGPUTextureFormat_None = 0xFFFFFFFF,
```
Kangz: @kainino0x what do you think of renaming this "Default" and set it at position 0 so that 0-initializing `WGPUTextureViewDescriptor` does what's expected?

Kangz: Great, let's do it! (sorry I didn't catch that earlier on the Dawn review)

kvark:
I think there are multiple things mixed in here:
  - "None" semantics, for things like a depth-stencil format that is meant to be optional
  - "Auto" semantics, for things like the view creation where we want the format to be automatically derived from the texture

I agree that "0" value would work better here, but I'd like us to straighten up the semantics of this value a little more. Perhaps, we can have two defines for the same value:
```cpp
WGPUTextureFormat_None = 0,
WGPUTextureFormat_Auto = 0,
```
And then document somewhere that only one or the other is applicable in a particular context?

An alternative to this would be to have `*const WGPUTextureFormat` in those places.

kainino0x:
It makes sense to distinguish them, but I don't like giving two names to the same value since we can't validate "incorrect" usage.

In terms of the JS-C mapping, "None" would just be JS `undefined` - a nice 1-1 mapping. So I think it makes sense to have just the one.

kvark:
I don't mind having just one name for this, but when I see `None` I find the semantics being different from "undefined". Could we call it `WGPUTextureFormat_Undefined = 0`?

Kangz:
Updated to use `Undefined` and the value 0, please look only at the second commit since the first one is the re-ordering from #6.

Kangz:
PTAL @kvark !

PR #8, opened 2019-09-06
Add WGPU[Texture|Buffer]Destroy

Kangz:
These are meant to mirror GPUTexture/Buffer.destroy() which frees the
GPU memory associated with these objects and makes them "internally"
null so that any subsequent operations using them will cause a
validation error. (including submitting a prerecorded command buffer).

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -689,6 +689,7 @@ typedef void (*WGPUProc)();
 typedef WGPUProc (*WGPUProcGetProcAddress)(WGPUDevice device, const char* procName);
 
 // Procs of Buffer
+typedef void (*WGPUProcBufferDestroy)(WGPUBuffer buffer);
 typedef void (*WGPUProcBufferMapReadAsync)(WGPUBuffer buffer, WGPUBufferMapReadCallback callback, void * userdata);
 typedef void (*WGPUProcBufferMapWriteAsync)(WGPUBuffer buffer, WGPUBufferMapWriteCallback callback, void * userdata);
 typedef void (*WGPUProcBufferUnmap)(WGPUBuffer buffer);
@@ -779,6 +780,7 @@ typedef void (*WGPUProcRenderPassEncoderSetViewport)(WGPURenderPassEncoder rende
 
 // Procs of Texture
 typedef WGPUTextureView (*WGPUProcTextureCreateView)(WGPUTexture texture, WGPUTextureViewDescriptor const * descriptor);
+typedef void (*WGPUProcTextureDestroy)(WGPUTexture texture);
 
 #endif  // !defined(WGPU_SKIP_PROCS)
 
@@ -787,6 +789,7 @@ typedef WGPUTextureView (*WGPUProcTextureCreateView)(WGPUTexture texture, WGPUTe
 WGPU_EXPORT WGPUProc WGPUGetProcAddress(WGPUDevice device, const char* procName);
 
 // Methods of Buffer
+WGPU_EXPORT void wgpuBufferDestroy(WGPUBuffer buffer);
 WGPU_EXPORT void wgpuBufferMapReadAsync(WGPUBuffer buffer, WGPUBufferMapReadCallback callback, void * userdata);
 WGPU_EXPORT void wgpuBufferMapWriteAsync(WGPUBuffer buffer, WGPUBufferMapWriteCallback callback, void * userdata);
 WGPU_EXPORT void wgpuBufferUnmap(WGPUBuffer buffer);
@@ -877,6 +880,7 @@ WGPU_EXPORT void wgpuRenderPassEncoderSetViewport(WGPURenderPassEncoder renderPa
 
 // Methods of Texture
 WGPU_EXPORT WGPUTextureView wgpuTextureCreateView(WGPUTexture texture, WGPUTextureViewDescriptor const * descriptor);
+WGPU_EXPORT void wgpuTextureDestroy(WGPUTexture texture);
 
 #endif  // !defined(WGPU_SKIP_DECLARATIONS)
 


kvark:
> (including submitting a prerecorded command buffer).

Alright, sounds like we need to talk about this. Isn't this behavior different from the upstream WebGPU behavior? My expectation from destroy() was that it's just a way to avoid relying on GC to release GPU resources. I.e. you call destroy() to say that you are no longer going to be *using the handle directly*, and the implementation can release the GPU resources as soon as it can (but not necessarily instantly!). There can still be indirect references to a resource, e.g. from a pre-recorded command buffer, which should keep its GPU contents alive.

Also, from our previous discussion on the refcounts, it appeared to me that on native we don't need destroy() in the way this PR proposes it:
  - we either have destroy() for every object explicitly (my preference), meaning (again) that the user can't use the handle *directly* any more
  - or have refcounts for objects, and releasing the last user reference means that the implementation can free GPU resources as soon as it can

Kangz:
> There can still be indirect references to a resource, e.g. from a pre-recorded command buffer, which should keep its GPU contents alive.

That's definitely not the intent of `GPUBuffer.destroy` at least in [BufferOperations](https://github.com/gpuweb/gpuweb/blob/master/design/BufferOperations.md) and in the original [PR](https://github.com/gpuweb/gpuweb/pull/96): calling `destroy` actually schedules the destruction after all previous GPU operations have completed and prevents any other operations acting on the buffer, including submitting command buffers that reference the buffer.

Having command buffers and render bundles keeping the buffers alive means we have the same issues with GC being unpredicatble. Would we need to manually destroy all the command buffer, render bundles, pass encoders and bindgroups that reference a buffer before we can be sure the memory has been reclaimed?

kvark:
I don't think the "spec" is clear about this:
> after a call to GPUBuffer.destroy where it is a validation error to do anything with the buffer.

We are missing a proper definition of "do anything with the buffer". One way to interpret this is the way I mentioned above: you can't do anything with this buffer handle *directly*. The way you are suggesting is: you can't do anything with this buffer *indirectly*.

kvark:
Given the confusion, we should bring this up perhaps at the call? It's not clear to me what benefit does the "aggressive destroying" interpretation brings, other than the increased chance that the users will screw themselves up.

Kangz:
For the record there was little discussion about this in the group and the minutes in this [doc](https://docs.google.com/document/d/1W13wnE2LKxCofo3Ysd2DaOgmI3CVk4UFr2YsKRhzmdk) aren't super clear either.

> It's not clear to me what benefit does the "aggressive destroying" interpretation brings

It allows implementations to free the GPU memory without waiting for all objects referencing the `[=buffer=]` directly or indirectly (`GPUBuffer` but also things like `GPUBindGroup` or `GPUCommandBuffer`) to be GCed. This is important because the GC doesn't get "GPU memory pressure events" and might keep all these objects alive for a very long time. Without "aggressive destroying" there is no way for JS applications to free GPU memory explicitly.

> Given the confusion, we should bring this up perhaps at the call?

I'd rather do a spec PR first, and if there is not consensus there, bring it up in the call. I'm confident that the group will agree to "aggressive destroying" because that's what it originally agreed to, in order to give JS apps control over GPU memory usage in the face of the GC.

Kangz:
BTW I'm glad this point of disagreement is now apparent: once resolved it will make the refcount vs. single-destroy discussion simpler.

kvark:
> It allows implementations to free the GPU memory without waiting for all objects referencing the [=buffer=] directly or indirectly

But it doesn't, right? It still has to wait for GPU to be done with the resources... So it feels like a half-measure anyway, it doesn't have a nice and simple meaning.

I think I understand this a little better now. So if we keep them alive for as long as, say, texture views are alive, then we'd still make the destruction depend on the JS engine GC, *unless* we also provide a similar method for the texture views destroying.

Note that for native refcounting/destroy() we know exactly when to free GPU resources without relying on GC anyway.

>I'm confident that the group will agree to "aggressive destroying" because that's what it originally agreed to

My concern here is that we are introducing more unexpected behavior: say, user created a bind group, verified at some point that it's valid. It would be great if they can assume that the bind group is valid forever, since it's immutable. Instead, with "aggressive destroy()" semantics at any point the bind group can suddenly become invalid :/

Kangz:
> But it doesn't, right? It still has to wait for GPU to be done with the resources.

That's kind of true but you can alias a newly created resource with the one pending destruction in Vulkan or D3D12 and we're planning to take advantage of this in Dawn eventually. In that sense the memory is freed right away. Note that even with `WGPUBufferDelete` (the no refcount lifetime management), the same would happen where the implementation would wait until the buffer is finished being used before destroying the `VkBuffer`.

> Note that for native refcounting/destroy() we know exactly when to free GPU resources without relying on GC anyway.

Agreed, assuming the application is able to destroy the complete object graph itself. However if we are to implement the JS API using the C API, we need to expose the semantic of JS's `GPUBuffer.destroy()`.

> My concern here is that we are introducing more unexpected behavior

That's fair, but I don't really see a way around it. Not that these destroy semantics are very useful to symbolize ownership transfers, for example at the end of requestAnimationFrame it is as if `swapchainTexture.destroy()` was called (except it gets presented to the screen). Otherwise the application could race with the compositor/display controller.

kvark:
> That's kind of true but you can alias a newly created resource with the one pending destruction in Vulkan or D3D12 and we're planning to take advantage of this in Dawn eventually. In that sense the memory is freed right away.

That's a very interesting point. You can say that the memory is freed, conceptually, on the queue timeline. But WebGPU has some operations on the device timeline as well, such as `createBufferMapped`. This function will never be able to work with memory of "destroyed" buffers that are still in use by GPU, so they aren't really "destroyed" in the sense you are presenting.

>Note that even with WGPUBufferDelete (the no refcount lifetime management), the same would happen where the implementation would wait until the buffer is finished being used before destroying the VkBuffer.

Yes, of course. The appeal of this semantics to me is that it's easy to define and reason about: once you call "destroy", your handle is no longer valid, that's it. Underlying resource can be held alive for as long as it's needed by the implementation.

> However if we are to implement the JS API using the C API, we need to expose the semantic of JS's `GPUBuffer.destroy()`.

Now this is super controversial. I take this requirement ("to implement the JS API using the C API") as being very specific to Dawn. We aren't doing this, and neither does Apple. Moreover, you said Dawn has some way of modifying the code paths so that C API used by the implementation is not exactly behaving the same as the C API for the user. So I think this requirement is not something you can hold "in court" advocating for the particular `destroy()` behavior.

> example at the end of requestAnimationFrame it is as if swapchainTexture.destroy() was called

It is true, and a good example that we already have this effect (of invalidating things that are seemingly immutable), eh :/

kainino0x:
> But it doesn't, right? It still has to wait for GPU to be done with the resources

Technically yes, but it is guaranteed to destroy in finite time (as long as the app is making forward progress) - and its timing is even predictable (can enqueue an allocation immediately and expect it to happen after the previous destroy really happens) and measurable (via a fence). The GC is not guaranteed to clean up the object in finite time, nor can it be predicted or measured.

Kangz:
> Yes, of course. The appeal of this semantics to me is that it's easy to define and reason about: once you call "destroy", your handle is no longer valid, that's it. Underlying resource can be held alive for as long as it's needed by the implementation.

The problem is that with semantics the application can't make sure GPU memory is made available in finite time because a `GPUBindGroupLayout` could be keeping a ref on the `GPUBuffer`. For the semantic you describe to be useful at all, we would need `.destroy()` on many more WebGPU objects.

> Now this is super controversial. I take this requirement ("to implement the JS API using the C API") as being very specific to Dawn. We aren't doing this, and neither does Apple.

My understanding was that you would implement WebGPU with wgpu-rs / wgpu-native which is basically the Rust API for WebGPU. We're doing the same but in C/C++ because that's the language Chromium uses. We do have many Dawn-specific entrypoints but I would expect everything that appears in the JS IDL to be present in the C header.

More specifically for `GPUBuffer.destroy()`: if we don't expose it in the C header, then WASM will have the same GC unpredictability that we described above.

kvark:
> The problem is that with semantics the application can't make sure GPU memory is made available in finite time because a GPUBindGroupLayout could be keeping a ref on the GPUBuffer

This is basically the same point as @kainino0x made in https://github.com/webgpu-native/webgpu-headers/pull/8#issuecomment-529654843
And I think it's a valid point. It's just that I'd like the concept of "invalidation" to be flashed out in the spec to reflect the fact that any object can technically become invalid at some point (perhaps we already describe that in the event of a lost device?).

> For the semantic you describe to be useful at all, we would need .destroy() on many more WebGPU objects.

You mean in JS? Yeah. I can see how this would be inconvenient.

> We're doing the same but in C/C++

No, there is an important different. We implement WebGPU API in the browser using a C API, but it's not the *same* C API as the users are supposed to use. It's wgpu-remote for us versus wgpu-native: two different crates technically.

Kangz:
> No, there is an important different. We implement WebGPU API in the browser using a C API, but it's not the same C API as the users are supposed to use. It's wgpu-remote for us versus wgpu-native: two different crates technically.

Because of the requirements of both or browsers, WebGPU's IDL is 100% remotable for free and libdawn_wire actually implements exactly webgpu.h (well, dawn.h but that's converging). I see a lot of value in having the C header be able to be backed by a remoting implementation.

Anyways. Do you still have concerns with this PR beyond the need to discuss again of the semantics of .destroy() in the group?

Kangz:
Thanks you! I added discussing the precise semantics of .destroy to the agenda for the F2F.

PR #13, opened 2019-10-16
Basic swapchain API

### LINKED ISSUE #11: Add swapchains!
Right now the header doesn't have any swapchain interfaces. Dawn's swapchain interface is OMG-level of horrible but I heard wgpu-native has something nice. @kvark could you contribute that interface to this repo?

kvark:
Fixes #11

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -73,6 +73,8 @@ typedef struct WGPUSamplerImpl* WGPUSampler;
 typedef struct WGPUShaderModuleImpl* WGPUShaderModule;
 typedef struct WGPUTextureImpl* WGPUTexture;
 typedef struct WGPUTextureViewImpl* WGPUTextureView;
+typedef struct WGPUSurfaceImpl* WGPUSurface;
+typedef struct WGPUSwapChainImpl* WGPUSwapChain;
 
 typedef enum WGPUAddressMode {
     WGPUAddressMode_Repeat = 0x00000000,
@@ -394,6 +396,11 @@ typedef enum WGPUTextureUsage {
 } WGPUTextureUsage;
 typedef WGPUFlags WGPUTextureUsageFlags;
 
+typedef enum WGPUPresentMode {
+  WGPUPresentMode_NoVsync = 0,
+  WGPUPresentMode_Vsync = 1,
+} WGPUPresentMode;
+
 
 typedef struct WGPUBindGroupBinding {
     uint32_t binding;
@@ -681,6 +688,15 @@ typedef struct WGPURenderPipelineDescriptor {
     bool alphaToCoverageEnabled;
 } WGPURenderPipelineDescriptor;
 
+typedef struct WGPUSwapChainDescriptor {
+    void const * nextInChain;
+    char const * label;
+    WGPUTextureUsageFlags usage;
+    WGPUTextureFormat format;
+    uint32_t width;
+    uint32_t height;
+    WGPUPresentMode presentMode;
+} WGPUTextureDescriptor;
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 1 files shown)


webgpu.h
```diff
@@ -750,6 +766,7 @@ typedef WGPURenderPipeline (*WGPUProcDeviceCreateRenderPipeline)(WGPUDevice devi
 typedef WGPUSampler (*WGPUProcDeviceCreateSampler)(WGPUDevice device, WGPUSamplerDescriptor const * descriptor);
 typedef WGPUShaderModule (*WGPUProcDeviceCreateShaderModule)(WGPUDevice device, WGPUShaderModuleDescriptor const * descriptor);
 typedef WGPUTexture (*WGPUProcDeviceCreateTexture)(WGPUDevice device, WGPUTextureDescriptor const * descriptor);
+typedef WGPUSwapChain (*WGPUProcDeviceCreateSwapChain)(WGPUDevice device, WGPUSurface surface, WGPUSwapChainDescriptor const * descriptor);
```
Kangz: nit: this isn't in alphabetical order and would be just before CreateTexture. Your PR is scary because it adds stuff in an order which isn't the same one generated by our code generator so making PRs will become a bit more complicated for us.

webgpu.h
```diff
@@ -901,6 +923,15 @@ WGPU_EXPORT void wgpuRenderPassEncoderSetViewport(WGPURenderPassEncoder renderPa
 WGPU_EXPORT WGPUTextureView wgpuTextureCreateView(WGPUTexture texture, WGPUTextureViewDescriptor const * descriptor);
 WGPU_EXPORT void wgpuTextureDestroy(WGPUTexture texture);
 
+// Methods of SwapChain
+WGPU_EXPORT WGPUTextureView wgpuSwapChainGetCurrentTextureView(WGPUSwapChain swapChain);
+WGPU_EXPORT void wgpuSwapChainPresent(WGPUSwapChain swapChain);
+
+// Surface creation
```
Kangz: What do you think of using descriptors for these too for extensibility? Also I'm thinking this might need an "Instance" argument somehow, which isn't something we have in this header for now.

webgpu.h
```diff
@@ -901,6 +923,15 @@ WGPU_EXPORT void wgpuRenderPassEncoderSetViewport(WGPURenderPassEncoder renderPa
...
+WGPUSurface wgpuCreateSurfaceFromMetalLayer(void *layer);
+WGPUSurface wgpuCreateSurfaceFromWindowsHwnd(void *_hinstance, void *hwnd);
```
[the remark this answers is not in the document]
grovesNL: We should be able to remove the `_` prefix.

It was just added there temporarily because we don't use `hinstance` internally in wgpu yet :) (the prefix avoids an unused parameter warning in Rust)

kvark:
Do we need more eyes on this? @grovesNL would you want to review these PRs here as well?

webgpu.h
Kangz: That's pretty cool, although I'm not sure what to do when a new platforms comes by (Fuchsia?). Would it be an extension on the descriptor?

Alternatively we could use the sType argument like Vulkan to be able to pass several differently shaped structures.

(also there are not proc definitions for these functions).

kainino0x: IMO, an sType would be pretty nice here.

Kangz: Looks good, last comment is that the WGPUProcCreateSurface is missing. (and we can take care of the instance argument separately, when we resolve whether to have an instance or not)

webgpu.h
```diff
@@ -797,6 +836,10 @@ typedef void (*WGPUProcRenderPassEncoderSetStencilReference)(WGPURenderPassEncod
 typedef void (*WGPUProcRenderPassEncoderSetVertexBuffers)(WGPURenderPassEncoder renderPassEncoder, uint32_t startSlot, uint32_t count, WGPUBuffer const * buffers, uint64_t const * offsets);
 typedef void (*WGPUProcRenderPassEncoderSetViewport)(WGPURenderPassEncoder renderPassEncoder, float x, float y, float width, float height, float minDepth, float maxDepth);
 
+// Procs of SwapChain
+typedef WGPUTextureView (*WGPUProcSwapChainGetCurrentTextureView)(WGPUSwapChain swapChain);
```
kainino0x: From the F2F, I thought we were going to go with Texture instead of TextureView here?

kvark: From the unofficial F2F, @Kangz noted they'd be fine with restricting the native API to be render-only, so returning `TextureView` makes sense. Let me know if you want to reconsider that.

Our implementation would for sure prefer the swapchain to be render-only.

webgpu.h
```diff
@@ -394,6 +396,11 @@ typedef enum WGPUTextureUsage {
 } WGPUTextureUsage;
 typedef WGPUFlags WGPUTextureUsageFlags;
 
+typedef enum WGPUPresentMode {
+  WGPUPresentMode_NoVSync = 0,
+  WGPUPresentMode_VSync = 1,
```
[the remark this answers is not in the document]
kvark: It seems to me that "v-sync" term is well recognized.

webgpu.h
kvark: I wasn't sure if instance-level entry points need a proc in your implementation. It's not like you can query those addresses anyway.

webgpu.h
kainino0x: Gotcha, this is fine with me.

Kangz: Yeah the main argument for WebGPU keeping `getTexture` is because we need to be able to do readbacks on the Web anyway. But in native we want to take advantage of "renderOnly" textures that can be more optimized.

Kangz:
I think this is good to merge and we can iterate. I'll integrate in dawn's code generator and make an update PR.

PR #17, opened 2019-10-29
Changes to swapchain after integration in the generator

Kangz:


webgpu.h
```diff
@@ -209,6 +215,14 @@ typedef enum WGPUPrimitiveTopology {
     WGPUPrimitiveTopology_Force32 = 0x7FFFFFFF
 } WGPUPrimitiveTopology;
 
+typedef enum WGPUSType {
+    WGPUSType_Invalid = 0x00000000,
```
Kangz: Made this a regular enum and added the "0" value is invalid so that people using the C API know they can't leave it undefined. (we could have good error messages on invalid sType that say you forgot to set it).

webgpu.h
```diff
@@ -566,6 +571,41 @@ typedef struct WGPUStencilStateFaceDescriptor {
     WGPUStencilOperation passOp;
 } WGPUStencilStateFaceDescriptor;
 
+typedef struct WGPUSurfaceDescriptor {
+    void const * nextInChain;
+    char const * label;
+} WGPUSurfaceDescriptor;
+
+typedef struct WGPUSurfaceDescriptorFromMetalLayer {
+    void const * nextInChain;
```
Kangz: Added nextInChain to sub descriptors too.

webgpu.h
```diff
@@ -566,6 +571,41 @@ typedef struct WGPUStencilStateFaceDescriptor {
...
+    WGPUSType sType;
+    void * hinstance;
+    void * hwnd;
+} WGPUSurfaceDescriptorFromWindowsHwnd;
+
+typedef struct WGPUSurfaceDescriptorFromWindowsXlib {
+    void const * nextInChain;
+    WGPUSType sType;
+    void * display;
```
Kangz: Made `display` a simple `void*`. Any reason it was a `void*const*`?

webgpu.h
```diff
@@ -566,6 +571,41 @@ typedef struct WGPUStencilStateFaceDescriptor {
     WGPUStencilOperation passOp;
 } WGPUStencilStateFaceDescriptor;
 
+typedef struct WGPUSurfaceDescriptor {
+    void const * nextInChain;
+    char const * label;
+} WGPUSurfaceDescriptor;
+
+typedef struct WGPUSurfaceDescriptorFromMetalLayer {
```
Kangz: Any reason this is from MetalLayer and not more generally from a CALayer? It seems like wgpu implementations could create a child CAMetalLayer if needed so that users don't need to know about Metal at all.

Kangz: Ack, we can always add another surface type if someone really really needs to pass a CALayer.

PR #22, opened 2019-11-12
Fix u32 dynamic offset in prototypes

Kangz:


kainino0x:
accidental comment-and-close?

PR #26, opened 2019-11-26
Add getBindGroupLayout and fix typos

Kangz:
PTAL!

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -767,7 +767,7 @@ typedef void (*WGPUProc)();
 #if !defined(WGPU_SKIP_PROCS)
 
 typedef WGPUInstance (*WGPUProcCreateInstance)(WGPUInstanceDescriptor const * descriptor);
-typedef WGPUProc (*WGPUProcGetProcAddress)(WGPUDevice device, const char* procName);
+typedef WGPUProc (*WGPUProcGetProcAddress)(WGPUDevice device, char const * procName);
 
 // Procs of Adapter
 typedef void (*WGPUProcAdapterRequestDevice)(WGPUAdapter adapter, WGPUDeviceDescriptor const * descriptor, WGPURequestDeviceCallback callback, void * userdata);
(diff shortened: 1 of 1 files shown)


webgpu.h
```diff
@@ -800,6 +800,9 @@ typedef void (*WGPUProcComputePassEncoderPushDebugGroup)(WGPUComputePassEncoder
 typedef void (*WGPUProcComputePassEncoderSetBindGroup)(WGPUComputePassEncoder computePassEncoder, uint32_t groupIndex, WGPUBindGroup group, uint32_t dynamicOffsetCount, uint32_t const * dynamicOffsets);
 typedef void (*WGPUProcComputePassEncoderSetPipeline)(WGPUComputePassEncoder computePassEncoder, WGPUComputePipeline pipeline);
 
+// Procs of ComputePipeline
+typedef WGPUBindGroupLayout (*WGPUProcComputePipelineGetBindGroupLayout)(WGPUComputePipeline computePipeline, uint32_t group);
```
kvark: the working group hasn't agreed on that ability in https://github.com/gpuweb/gpuweb/issues/446

Kangz: Looking at the [minutes](https://docs.google.com/document/d/18CnUBiiUKW11zGsM_RuAkaIBv-O4s3txFuGqcQIb6sw/) what I get from them is that it was agreed upon with just a discussion left on whether the same JS objects are returned or not.

kvark: There is not even a PR open to the spec! We shouldn't rush in here before the upstream is done.

I don't think the discussion is done either, e.g. was this addressed?
> I do wonder though, how would this play with WSL shaders? Overall, since we have to write down in the specification the deriving rules from the shading language, this PR might be blocked until we settle on what that language is.

Kangz: > There is not even a PR open to the spec! We shouldn't rush in here before the upstream is done.

I guess we had a different understanding of how accepted this was. I'll temporarily remove it from what's generated for this webgpu.h.

> MM: sounds fine. Believe it is possible to create a pipeline layout from WSL shader.

That's from the minutes, but I guess it is a "likely ok but need to check"?

kvark: Man, I wish people documented their decisions in the appropriate discussions instead of waiting till the last minute to voice them on a call/f2f :/

Kangz:
Closing and moved the typo fixes to #27

PR #27, opened 2019-11-26
Adapter properties

Kangz:
PTAL only at the second commit for this PR.

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -80,12 +80,30 @@ typedef struct WGPUSwapChainImpl* WGPUSwapChain;
 typedef struct WGPUTextureImpl* WGPUTexture;
 typedef struct WGPUTextureViewImpl* WGPUTextureView;
 
+typedef enum WGPUAdapterType {
+    WGPUAdapterType_DiscreteGPU = 0x00000000,
+    WGPUAdapterType_IntegratedGPU = 0x00000001,
+    WGPUAdapterType_CPU = 0x00000002,
+    WGPUAdapterType_Unknown = 0x00000003,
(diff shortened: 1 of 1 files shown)


webgpu.h
```diff
@@ -767,9 +794,10 @@ typedef void (*WGPUProc)();
 #if !defined(WGPU_SKIP_PROCS)
 
 typedef WGPUInstance (*WGPUProcCreateInstance)(WGPUInstanceDescriptor const * descriptor);
-typedef WGPUProc (*WGPUProcGetProcAddress)(WGPUDevice device, const char* procName);
+typedef WGPUProc (*WGPUProcGetProcAddress)(WGPUDevice device, char const * procName);
 
 // Procs of Adapter
+typedef bool (*WGPUProcAdapterGetProperties)(WGPUAdapter adapter, WGPUAdapterProperties * properties);
```
Kangz: I chose to have `AdapterGetProperties` return a boolean so that it can return `false` when it doesn't support one of the extensions used in `properties`. We should figure out how we want error handling before `WGPUDevice` to happen.

Kangz:
Thanks!

webgpu.h
kvark: This isn't the way Vulkan does it, I think. Instead, the STYPE things not known are just ignored, and the spec is written in such a way that leaving their contents to be zeroed is equivalent to the defaults. I mean, this function isn't the right place to check for supported extensions anyway, we shouldn't try.

Kangz: I see, so you'd rather return `void` here?

kvark: yeah

Kangz: #28

PR #31, opened 2020-01-21
Add WGPUSurfaceDescriptorFromHTMLCanvasId

Kangz:


Kangz:
Thanks for the explanation. That sounds good, and in emscripten we could even have this single surface type, and first check getElementById and if nothing is found look in the internal array. (or have some sort of prefix that means the internal array).

kainino0x:
If we did that I'd probably prefer to stick with the way `findCanvasEventTarget`* (not `findEventTarget`, sorry) works for consistency, and call it target instead of id.

kainino0x:
(I still prefer matching getElementById exactly, so let's merge)

PR #32, opened 2020-01-22
Add GetSurfacePreferredFormat

Kangz:
PTAL!

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -797,6 +797,7 @@ typedef void (*WGPUErrorCallback)(WGPUErrorType type, char const * message, void
 typedef void (*WGPUFenceOnCompletionCallback)(WGPUFenceCompletionStatus status, void * userdata);
 typedef void (*WGPURequestAdapterCallback)(WGPUAdapter result, void * userdata);
 typedef void (*WGPURequestDeviceCallback)(WGPUDevice result, void * userdata);
+typedef void (*WGPUSurfaceGetPreferredFormatCallback)(WGPUTextureFormat format, void * userdata);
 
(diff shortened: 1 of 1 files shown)


webgpu.h
```diff
@@ -858,6 +859,7 @@ typedef void (*WGPUProcFenceOnCompletion)(WGPUFence fence, uint64_t value, WGPUF
 
 // Procs of Instance
 typedef WGPUSurface (*WGPUProcInstanceCreateSurface)(WGPUInstance instance, WGPUSurfaceDescriptor const * descriptor);
+typedef void (*WGPUProcInstanceGetSurfacePreferredFormat)(WGPUInstance instance, WGPUSurface surface, WGPUAdapter adapter, WGPUGetSurfacePreferredFormatCallback callback, void * userdata);
```
kvark: I also thought, after talking to you, that having this on `Instance` is unnecessary, if it could be a method on either the surface or the adapter. Sorry about not communicating this earlier! I'd have a minor preference towards having this on the adapter, since it has other methods.

kainino0x: I have a slight preference for having it on Surface, because in WebGPU it's on GPUCanvasContext which is kind of equivalent.

kainino0x: (and not on Adapter, since it's not on Adapter in WebGPU)

kvark: That's fine by me. I suggested to put it on an adapter because Vulkan has it on a physical device: `vkGetPhysicalDeviceSurfaceFormatsKHR`

Kangz: Put it on the surface, but happy to bikeshed more.

PR #34, opened 2020-02-05
webgpu.h: Fix WGPU_WHOLE_SIZE causing linking errors.

Kangz:
The symbol was not marked as static and would cause linking errors
because it would be defined in multiple translation units. Replace it
with a more traditional C-style #define.

Found by @floooh.

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -52,7 +52,7 @@
 #include <stddef.h>
 #include <stdbool.h>
 
-const uint64_t WGPU_WHOLE_SIZE = 0xffffffffffffffffULL; // UINT64_MAX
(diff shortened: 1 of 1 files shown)


kvark:
would adding `static` make it stronger instead?

floooh:
```static``` worked too when I tested, but since this is specifically a fix for the C language, IMHO a #define is the better choice - it's usually only the C++ people who have a problem with the C preprocessor ;)

Kangz:
@kvark given the above, WDYT of using the `#define`?

Kangz:
Thanks for the review!

PR #36, opened 2020-03-19
Add read/write storage textures and storageTextureFormat

Kangz:


Kangz:
Note that this doesn't remove the regular storage texture enum, because it's pretty clear that it will be supported at some point.

webgpu.h
```diff
@@ -113,6 +113,8 @@ typedef enum WGPUBindingType {
     WGPUBindingType_Sampler = 0x00000003,
     WGPUBindingType_SampledTexture = 0x00000004,
     WGPUBindingType_StorageTexture = 0x00000005,
```
kvark: Would be great at least to avoid mixing the stuff we add in native with the base stuff from the Web, i.e. move it outside like
```
WGPUBindingType_StorageTexture = 0x000000010;
```

webgpu.h
```diff
@@ -470,6 +472,7 @@ typedef struct WGPUBindGroupLayoutBinding {
     bool multisampled;
     WGPUTextureViewDimension textureDimension;
```
kvark: might as well rename this to `viewDimension`

Kangz:
Closing in favor of https://github.com/webgpu-native/webgpu-headers/pull/38

PR #37, opened 2020-03-19
Make present modes more Vulkan-like.

Kangz:
The previous present modes couldn't make the difference between double
and triple buffering. This also changes the enums to be more Vulkan-like
because the names make more semantic sense.


Kangz:
I don't think so, it's very Vulkan-specific so I think we can get away with not having it for now.

Kangz:
Reordered to match https://github.com/gfx-rs/wgpu/blob/master/ffi/wgpu.h and merging with @kvark in TBR since it matches wgpu.h.

PR #42, opened 2020-04-06
Add occlusion queries

Kangz:
PTAL! I changed the name of arguments to resolve slightly.


Kangz:
Thanks for the review!

PR #43, opened 2020-04-08
Fix a -Wstrict-prototype

Kangz:


Kangz:
This is a trivial fix, so I'll just land it.

PR #49, opened 2020-05-20
Add pipeline statistics queries

Kangz:
PTAL!

Two concerns:

 - Do we want pipeline statistics on `WGPURenderBundle`? Currently the spec says yes, but we might want to change this.
 - Do can pipeline statistics queries be nested? If not, then `EndPipelineStatisticsQuery` should not take any arguments.

Opened https://github.com/gpuweb/gpuweb/issues/794 for this.

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -227,6 +227,15 @@ typedef enum WGPULoadOp {
     WGPULoadOp_Force32 = 0x7FFFFFFF
 } WGPULoadOp;
 
+typedef enum WGPUPipelineStatisticsName {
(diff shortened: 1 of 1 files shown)


kvark:
Not sure about nesting. That's something to ask the main investigator in WebGPU group :)

haoxli:
Yes, they are not supported on bundles, but we can nest query operations for a same queryset or querysets with different types.

Kangz:
PTAL again, updated to remove pipeline statistics queries from render bundles, and removed the argument in `EndPipelineStatisticsQuery`.

Kangz:
Thanks!

PR #50, opened 2020-05-20
Add `writeBuffer` and `writeTexture`.

Kangz:
There are slight differences compared to the WebGPU spec:
 - The order of arguments is different (there will be an equivalent
   WebGPU spec PR).
 - `writeBuffer` doesn't take a `sourceOffset` since it can be done
   with pointer arithmetic.
 - `writeTexture` takes an extra `dataSize` that will let implementations
   validate that there is no OOB. This is also useful for remoting
   implementations.
 - `WGPUBufferCopyView` doesn't inherit from `WGPUTextureDataLayout`
   since inheritance doesn't exist in C. Instead it is reordered to
   contain `WGPUTextureDataLayout` "inline".

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -488,14 +488,6 @@ typedef struct WGPUBlendDescriptor {
     WGPUBlendFactor dstFactor;
 } WGPUBlendDescriptor;
 
-typedef struct WGPUBufferCopyView {
-    WGPUChainedStruct const * nextInChain;
-    WGPUBuffer buffer;
-    uint64_t offset;
-    uint32_t bytesPerRow;
-    uint32_t rowsPerImage;
-} WGPUBufferCopyView;
-
 typedef struct WGPUBufferDescriptor {
     WGPUChainedStruct const * nextInChain;
     char const * label;
@@ -686,6 +678,13 @@ typedef struct WGPUSwapChainDescriptor {
     WGPUPresentMode presentMode;
 } WGPUSwapChainDescriptor;
 
+typedef struct WGPUTextureDataLayout {
+    WGPUChainedStruct const * nextInChain;
+    uint64_t offset;
+    uint32_t bytesPerRow;
+    uint32_t rowsPerImage;
+} WGPUTextureDataLayout;
+
 typedef struct WGPUTextureViewDescriptor {
     WGPUChainedStruct const * nextInChain;
     char const * label;
@@ -718,6 +717,12 @@ typedef struct WGPUBindGroupLayoutDescriptor {
     uint32_t entryCount;
     WGPUBindGroupLayoutEntry const * entries;
 } WGPUBindGroupLayoutDescriptor;
+
+typedef struct WGPUBufferCopyView {
+    WGPUChainedStruct const * nextInChain;
+    WGPUTextureDataLayout layout;
+    WGPUBuffer buffer;
+} WGPUBufferCopyView;
 
 typedef struct WGPUColorStateDescriptor {
     WGPUChainedStruct const * nextInChain;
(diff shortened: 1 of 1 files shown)


Kangz:
See https://github.com/gpuweb/gpuweb/pull/795 I might have to do another reordering of arguments in this PR if the gpuweb PR is landed as is.

webgpu.h
```diff
@@ -689,6 +689,13 @@ typedef struct WGPUSwapChainDescriptor {
     WGPUPresentMode presentMode;
 } WGPUSwapChainDescriptor;
 
+typedef struct WGPUTextureDataLayout {
```
kvark, line 692: how are we going to ensure these structs are actually nested? i.e. can we just do
```cpp
typedef struct WGPUBufferCopyView {
    WGPUTextureDataLayout layout;
    WGPUBuffer buffer;
} WGPUBufferCopyView;
```

webgpu.h
```diff
@@ -909,6 +916,8 @@ typedef void (*WGPUProcQuerySetDestroy)(WGPUQuerySet querySet);
 typedef WGPUFence (*WGPUProcQueueCreateFence)(WGPUQueue queue, WGPUFenceDescriptor const * descriptor);
 typedef void (*WGPUProcQueueSignal)(WGPUQueue queue, WGPUFence fence, uint64_t signalValue);
 typedef void (*WGPUProcQueueSubmit)(WGPUQueue queue, uint32_t commandCount, WGPUCommandBuffer const * commands);
+typedef void (*WGPUProcQueueWriteBuffer)(WGPUQueue queue, WGPUBuffer destination, uint64_t offset, size_t size, void const * data);
+typedef void (*WGPUProcQueueWriteTexture)(WGPUQueue queue, WGPUTextureCopyView destination, void const * data, size_t dataSize, WGPUTextureDataLayout const * dataLayout, WGPUExtent3D const * writeSize);
```
kvark, line 920: It's a bit controversial to have the dataSize here, but looks to be a good thing

webgpu.h
Kangz: That would work. I'm not looking forward to doing another progressive migration :P

webgpu.h
```diff
@@ -722,6 +721,12 @@ typedef struct WGPUBindGroupLayoutDescriptor {
     WGPUBindGroupLayoutEntry const * entries;
 } WGPUBindGroupLayoutDescriptor;
 
+typedef struct WGPUBufferCopyView {
+    WGPUChainedStruct const * nextInChain;
```
kvark, line 725: there is a bit of a problem that both structures contain `nextInChain` fields, is that intentional?

Kangz:
Addressed the comment, and reordered the arguments to `writeTexture` to match https://github.com/gpuweb/gpuweb/pull/795

webgpu.h
Kangz: Somewhat, because either is a "dictionary" and can be extended. That shouldn't be a problem in general, `WGPURenderPipelineDescriptor` has multiple extension points too.

Kangz:
This is not an exactly trivial PR so I'll wait for @austinEng's review before merging.

austinEng:
> Addressed the comment, and reordered the arguments to `writeTexture` to match [gpuweb/gpuweb#795](https://github.com/gpuweb/gpuweb/pull/795)

It looks like that PR has the `size` as the last argument of `writeBuffer` and the `data` comes after the `layout` in `writeTexture`

Kangz:
Whoops, fixed the order!

Kangz:
I'll wait for the WebGPU PR, because it suddenly became bikesheddy.

kainino0x:
Sorry ^_^

Kangz:
No worries, it is healthy :)

Kangz:
Updated to match the order of the PR that was landed in WebGPU.

Kangz:
Thanks!

webgpu.h
kainino0x: FYI in Emscripten I'm probably going to ignore it, because it's nontrivial to verify (and I wouldn't really have any way to signal the error except to ASSERT in debug builds).

PR #52, opened 2020-05-20
Add the known changes for mapAsync

Kangz:
While the exact shape of the mapAsync calls is still under discussion,
the change to have getMappedRange and
GPUBufferDescriptor.mappedAtCreation are agreed upon.

getMappedRange is split into getMappedRange and getConstMappedRange to
provide better type safety in native.

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -501,6 +501,7 @@ typedef struct WGPUBufferDescriptor {
     char const * label;
     WGPUBufferUsageFlags usage;
     uint64_t size;
+    bool mappedAtCreation;
 } WGPUBufferDescriptor;
 
 typedef struct WGPUColor {
@@ -524,12 +525,6 @@ typedef struct WGPUComputePassDescriptor {
     WGPUChainedStruct const * nextInChain;
     char const * label;
 } WGPUComputePassDescriptor;
-
-typedef struct WGPUCreateBufferMappedResult {
-    WGPUBuffer buffer;
-    uint64_t dataLength;
-    void * data;
-} WGPUCreateBufferMappedResult;
 
 typedef struct WGPUDeviceDescriptor {
     WGPUChainedStruct const * nextInChain;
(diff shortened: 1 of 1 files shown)


webgpu.h
```diff
@@ -843,6 +838,8 @@ typedef void (*WGPUProcAdapterRequestDevice)(WGPUAdapter adapter, WGPUDeviceDesc
 
 // Procs of Buffer
 typedef void (*WGPUProcBufferDestroy)(WGPUBuffer buffer);
+typedef void * const (*WGPUProcBufferGetConstMappedRange)(WGPUBuffer buffer);
```
austinEng, line 841: I think this should be:
```suggestion
typedef void const * (*WGPUProcBufferGetConstMappedRange)(WGPUBuffer buffer);
```

webgpu.h
```diff
@@ -843,6 +838,8 @@ typedef void (*WGPUProcAdapterRequestDevice)(WGPUAdapter adapter, WGPUDeviceDesc
 
 // Procs of Buffer
 typedef void (*WGPUProcBufferDestroy)(WGPUBuffer buffer);
+typedef void const * (*WGPUProcBufferGetConstMappedRange)(WGPUBuffer buffer);
```
kvark, line 841: This is going to be the same function returning the same thing, and C isn't protecting against any writes into this memory. The utility of returning `const *` here in an extra method (that doesn't exist in the upstream API) seems pretty limited.

kainino0x: In native we can expose real mappings here, where writes to the mapping have undefined results. By making the pointer const, we avoid a very hidden footgun. Even though the caller can const_cast (or c cast) the pointer, such code would be more immediately suspicious.

kvark: Are you going to have the same function pointer for both of these signatures? Or is it truly a separate function on your side?

Kangz: It also serve as validation that `GetMappedRange` isn't called for `mapReadAsync`, and any validation error would make the function return nullptr. In Rust you could return something like `Option<&[u8]>` or `Option<&mut [u8]>` (with the appropriate lifetime ofc).

Kangz: Nice race conditions :) The validation specific to `GetMappedRange` would mean it would be two different entrypoints.

Kangz:
Great! Thank you!

PR #53, opened 2020-06-02
Second part of mapAsync

Kangz:
PTAL!

I'm not 100% happy with this but I guess people will make helper mapReadAsync and mapWriteAsync methods in their wrapper.

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -437,6 +437,13 @@ typedef enum WGPUColorWriteMask {
 } WGPUColorWriteMask;
 typedef WGPUFlags WGPUColorWriteMaskFlags;
 
+typedef enum WGPUMapMode {
+    WGPUMapMode_Read = 0x00000001,
+    WGPUMapMode_Write = 0x00000002,
+    WGPUMapMode_Force32 = 0x7FFFFFFF
+} WGPUMapMode;
+typedef WGPUFlags WGPUMapModeFlags;
+
 typedef enum WGPUShaderStage {
     WGPUShaderStage_None = 0x00000000,
     WGPUShaderStage_Vertex = 0x00000001,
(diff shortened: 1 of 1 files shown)


webgpu.h
```diff
@@ -427,6 +437,13 @@ typedef enum WGPUColorWriteMask {
 } WGPUColorWriteMask;
 typedef WGPUFlags WGPUColorWriteMaskFlags;
 
+typedef enum WGPUMappingFlag {
```
austinEng, line 440: This may not be the right place to bikeshed, but I didn't see the PR to add this to the spec yet.

I think we should not have the "Flag" suffix here as we don't have it on WGPUShaderStage, WGPUTextureUsage, WGPUBufferUsage, etc. It also makes the `WGPUMappingFlagFlags` typedef kinda ugly.

Maybe WGPUMapMode and WGPUMapModeFlags ?

webgpu.h
```diff
@@ -845,8 +864,7 @@ typedef void (*WGPUProcAdapterRequestDevice)(WGPUAdapter adapter, WGPUDeviceDesc
 typedef void (*WGPUProcBufferDestroy)(WGPUBuffer buffer);
 typedef void const * (*WGPUProcBufferGetConstMappedRange)(WGPUBuffer buffer);
 typedef void * (*WGPUProcBufferGetMappedRange)(WGPUBuffer buffer);
-typedef void (*WGPUProcBufferMapReadAsync)(WGPUBuffer buffer, WGPUBufferMapReadCallback callback, void * userdata);
-typedef void (*WGPUProcBufferMapWriteAsync)(WGPUBuffer buffer, WGPUBufferMapWriteCallback callback, void * userdata);
+typedef void (*WGPUProcBufferMapAsync)(WGPUBuffer buffer, WGPUMappingFlagFlags flags, size_t offset, size_t size, WGPUBufferMapWriteCallback callback, void * userdata);
```
austinEng, line 867: I believe `offset` should be `uint64_t`. Actually now that I think about it, that's a pretty good reason for both read and write to have the `offset` and `size` args. If the CPU can't address all of the GPU memory then it would be impossible to map the whole buffer, and then `getMappedRange()` at the start of it, and at the end of it.

webgpu.h
```diff
@@ -227,6 +227,15 @@ typedef enum WGPULoadOp {
     WGPULoadOp_Force32 = 0x7FFFFFFF
 } WGPULoadOp;
 
+typedef enum WGPUPipelineStatisticsName {
```
kvark, line 230: unrelated code!

webgpu.h
kvark: nobody should operate on staging buffers that can't be addressed on CPU in full, that would be quite crazy :)

Kangz:
Addressed comments (except `size_t`)

Kangz:
I just realized I had forgotten to update the callbacks. I don't think that's controversial so merging this.

PR #58, opened 2020-07-02
Updates from latest spec

Kangz:
- Adds minBufferBindingSize
 - Adds timestamp queries
 - Fix a missing const* for WGPUTextureDataLayout in writeTexture

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -255,6 +255,7 @@ typedef enum WGPUPrimitiveTopology {
 typedef enum WGPUQueryType {
     WGPUQueryType_Occlusion = 0x00000000,
     WGPUQueryType_PipelineStatistics = 0x00000001,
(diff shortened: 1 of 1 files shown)


webgpu.h
```diff
@@ -497,6 +498,7 @@ typedef struct WGPUBindGroupLayoutEntry {
     WGPUTextureViewDimension viewDimension;
     WGPUTextureComponentType textureComponentType;
     WGPUTextureFormat storageTextureFormat;
+    uint64_t minBufferBindingSize;
```
kvark, line 501: let's put it right after `hasDynamicOffset` since they both apply to buffers

Kangz: Done

Kangz:
Thanks!

PR #62, opened 2020-07-15
Add offset/size to getMappedRange

kainino0x:
and rename flags -> mode to match upstream

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -863,9 +863,9 @@ typedef void (*WGPUProcAdapterRequestDevice)(WGPUAdapter adapter, WGPUDeviceDesc
 
 // Procs of Buffer
 typedef void (*WGPUProcBufferDestroy)(WGPUBuffer buffer);
-typedef void const * (*WGPUProcBufferGetConstMappedRange)(WGPUBuffer buffer);
-typedef void * (*WGPUProcBufferGetMappedRange)(WGPUBuffer buffer);
-typedef void (*WGPUProcBufferMapAsync)(WGPUBuffer buffer, WGPUMapModeFlags flags, size_t offset, size_t size, WGPUBufferMapCallback callback, void * userdata);
+typedef void const * (*WGPUProcBufferGetConstMappedRange)(WGPUBuffer buffer, size_t offset, size_t size);
+typedef void * (*WGPUProcBufferGetMappedRange)(WGPUBuffer buffer, size_t offset, size_t size);
+typedef void (*WGPUProcBufferMapAsync)(WGPUBuffer buffer, WGPUMapModeFlags mode, size_t offset, size_t size, WGPUBufferMapCallback callback, void * userdata);
 typedef void (*WGPUProcBufferUnmap)(WGPUBuffer buffer);
 
 // Procs of CommandEncoder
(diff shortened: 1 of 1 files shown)


webgpu.h
```diff
@@ -863,9 +863,9 @@ typedef void (*WGPUProcAdapterRequestDevice)(WGPUAdapter adapter, WGPUDeviceDesc
 
 // Procs of Buffer
 typedef void (*WGPUProcBufferDestroy)(WGPUBuffer buffer);
-typedef void const * (*WGPUProcBufferGetConstMappedRange)(WGPUBuffer buffer);
-typedef void * (*WGPUProcBufferGetMappedRange)(WGPUBuffer buffer);
-typedef void (*WGPUProcBufferMapAsync)(WGPUBuffer buffer, WGPUMapModeFlags flags, size_t offset, size_t size, WGPUBufferMapCallback callback, void * userdata);
+typedef void const * (*WGPUProcBufferGetConstMappedRange)(WGPUBuffer buffer, size_t offset, size_t size);
```
Kangz, line 866: I remember we discussed this already and agreed that in "native" it doesn't make sense to have offset and size because you can easily do pointer manipulations (whereas it's a pain to do it in Javascript)

kainino0x:
What about invalidating regions for flushing?

Kangz:
Maybe you don't get that in emscripten.

kvark:
@Kangz 
> Maybe you don't get that in emscripten.

But this would be a perf concern if you suddenly flush a whole more than necessary. Let's not deviate from the Web spec unless it's necessary. Do you recall where the conversation happened that concluded we don't need the offset/size?

Kangz:
I think it was in one of our biweekly meetings. The thing is that in a C API the size argument is completely useless, and I suspect folks will actually shoot themselves in the foot by requesting a size that's too small. It will work in native but fail in emscripten. (same concern for MapAsync arguments, but that should be called in less places.)

I think in native we can say that the MapAsync range is always completely flushed, and that's fine. Alternatively we could have an Unmap with a dirty range like D3D12.

Essentially GetMappedRange is how you interact with memory, and it makes sense that it would be different between JS and a C API.

Kangz:
Note that wgpu-rs also has the [argument-less GetMappedRange](https://wgpu.rs/doc/wgpu/struct.BufferSlice.html#method.get_mapped_range).

kvark:
> The thing is that in a C API the size argument is completely useless, and I suspect folks will actually shoot themselves in the foot by requesting a size that's too small.

It's useful in the same sense as it was designed: to signify which ranges need to be flushed. This still holds. And yes, C users can shoot themselves in both feet :) We can tell them to use write_buffer instead ;)

> I think in native we can say that the MapAsync range is always completely flushed, and that's fine. Alternatively we could have an Unmap with a dirty range like D3D12.

Why would this be fine? We are losing the feature for no reason.

> Note that wgpu-rs also has the argument-less GetMappedRange.

Sorry, looks like you got fooled a bit by the type system. It has no arguments because it's a method of `BufferSlice`, which itself has offset and size in it.

Kangz:
> Sorry, looks like you got fooled a bit by the type system. It has no arguments because it's a method of BufferSlice, which itself has offset and size in it.

Isn't the slice where the mapping happens as well? Like you'd need to do:

```rust
let buffer : Buffer = ...;
let slice = buffer.slice(some_bound);
slice.map_async(MapMode::Write).await;
let buffer_view = slice.get_mapped_range();
```

Then it means you always get a buffer view for the full range that was mapped.

kvark:
It's *just* a slice, it doesn't have to be the same slice. See [our belt implementation](https://github.com/gfx-rs/wgpu-rs/blob/32de700ef908029cc5f8ac3b0befd7439f5bd7bd/src/util/belt.rs#L105) for example.

Kangz:
Ok let's do it.

kainino0x:
In a C++ wrapper [with C++20 :( ] we could return a std::span to make it safer.

PR #73, opened 2020-12-15
Readonly/Writeonly -> ReadOnly/WriteOnly

### LINKED ISSUE #72: Readonly vs ReadOnly
If we take a direct casing conversion from JS/IDL's `"readonly-storage"` to camel case `ReadonlyStorage` then it looks maybe a little funny. Should we case it as `ReadOnlyStorage`?

Note this also comes up again soon in WGPUStorageTextureAccess.

Should we keep consistency with IDL (`Readonly`), break with it (`ReadOnly`), or change the IDL (`"read-only-storage"`)?

cc @toji

kainino0x:
These are the only cases for now; header hasn't caught up with latest spec yet.

Fixes #72

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -111,13 +111,13 @@ typedef enum WGPUBackendType {
 typedef enum WGPUBindingType {
     WGPUBindingType_UniformBuffer = 0x00000000,
     WGPUBindingType_StorageBuffer = 0x00000001,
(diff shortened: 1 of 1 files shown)


kvark (approved):

Kangz (approved):
LGTM although there's going to be the big update with the BGLEDesc change.

Issue #45, opened 2020-04-09
Have all enums have "undefined" as value 0?


Kangz:
This would allow zero-initializing to be as close as an empty dictionnary as possible in JS. It would also help if the JS API decide in after version 1 to make some enum members optional for example when adding defaults.

kvark:
I don't think it's worth mimicking the empty dictionary here, but I do buy the argument about making some stuff optional. How about this compromise: we switch all enum values to start with `1`, but only add `Undefined = 0` where they can truly be passed as optionals? That would be forward compatible.

Kangz:
Great idea. I'll do that, but it will take some time because there's a lot of reordering to do :)

kainino0x:
Should we:
1. have "undefined" variants for enums which are sometimes used with defaults (for example TextureDimension which defaults to "2d" in TextureDescriptor) - and implement the IDL's defaulting in webgpu.h implementations?
2. or only have them for enums which we use as nullable (for example LoadOp in depthLoadOp/stencilLoadOp)?

I'm tentatively in favor of 2. We can switch to 1 later.

I'm not exactly sure what Dawn does right now, but I'm pretty sure we don't implement the IDL's defaulting at least in most cases.

Kangz:
In Dawn we try to implement the IDL defaulting as much as possible. Though maybe we missed a couple things. In general it's nice if defaulting can be equivalent to zero-initialization as much as possible, so IMHO option 1 would be ideal, with Undefined in every single enum to be future proof when other additions to the header use the enum.

kainino0x:
OK with me. I do like having the defaulting there, though because of the various sentinel UNDEFINED values zero-initialization unfortunately doesn't work in general (https://github.com/webgpu-native/webgpu-headers/issues/158#issuecomment-1165034163). But it does make it somewhat easier to use zero-initialization and initialize only the fields which don't default to 0 (probably going to be common practice even after we add INIT macros).

kainino0x:
Another question came up when I started implementing this: what should we do with output/status enums?
- status enums: should success be 0 or should we leave 0 unallocated?
- status enums that have multiple success values (WGPUWaitStatus)

kainino0x:
More questions. Do we actually need these? AFAICT neither of these really needs to have default values.
- WGPUSType_Invalid?
[...]

PR #75, opened 2021-01-12
Add WGPUFeatureName enum and wgpuAdapterGetFeatures

### LINKED ISSUE #74: How to pass features in the device descriptor?
In the JS API,  the device descriptor takes a sequence of strings which are the feature names. The list of features (as strings) can also be queried on the adapter.

In the C API I see two options to represent this. (actually three, but the last one seems bad)
 1. List of null-terminated strings. This is how Vulkan does device extensions. We'd have a list of feature names somewhere and ask people not to make extensions with conflicting names.
 2. List of enums - we'd ask people making extensions to make a PR to webgpu-headers to reserve their enum values. This is what SPIR-V does for OpCodes introduced by an extension. Emscripten or whatever translates C -> JS would need to have a table mapping the enums to strings for the JS API.
 3. A struct where each feature is a boolean member. This is not easily extensible.
 
@kvark @Kangz what do you think?

I have a slight preference for (1) since it's the most direct translation to JS, so it would be less maintenance in the C->JS layer. And, I don't think it's that important to optimize passing the list of features since it only happens on device creation.

Edit: Nevermind, I remembered we have string enums in JS anyway so there need to be enum tables maintained anyway. I prefer (2) then.

austinEng:
wgpuAdapterGetFeatures when passed nullptr for |features| returns
the number of features the adapter has. When passed non-null, it
serializes the list of enums of available features into the pointer.

A list of features can then be passed in the WGPUDeviceDescriptor to
request a device with those features enabled.

Fixes #74

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -444,6 +444,16 @@ typedef enum WGPUBufferUsage {
 } WGPUBufferUsage;
 typedef WGPUFlags WGPUBufferUsageFlags;
 
+typedef enum WGPUFeatureName {
+  WGPUFeatureName_DepthClamping = 0x00000000,
+  WGPUFeatureName_Depth24UnormStencil8 = 0x00000001,
+  WGPUFeatureName_Depth32FloatStencil8 = 0x00000002,
+  WGPUFeatureName_PipelineStatisticsQuery = 0x00000003,
+  WGPUFeatureName_TextureCompressionBC = 0x00000004,
+  WGPUFeatureName_TimestampQuery = 0x00000005,
+  WGPUFeatureName_Force32 = 0x7FFFFFFF
+} WGPUFeatureName;
+
 typedef enum WGPUColorWriteMask {
     WGPUColorWriteMask_None = 0x00000000,
     WGPUColorWriteMask_Red = 0x00000001,
@@ -557,6 +567,8 @@ typedef struct WGPUComputePassDescriptor {
 
 typedef struct WGPUDeviceDescriptor {
     WGPUChainedStruct const * nextInChain;
+    uint32_t nonGuaranteedFeaturesCount;
+    WGPUFeatureName const * nonGuaranteedFeatures;
 } WGPUDeviceDescriptor;
 
 typedef struct WGPUExtent3D {
@@ -881,6 +893,7 @@ typedef WGPUProc (*WGPUProcGetProcAddress)(WGPUDevice device, char const * procN
 
 // Procs of Adapter
 typedef void (*WGPUProcAdapterGetProperties)(WGPUAdapter adapter, WGPUAdapterProperties * properties);
+typedef uint32_t (*WGPUProcAdapterGetFeatures)(WGPUAdapter adapter, WGPUFeatureName * features);
 typedef void (*WGPUProcAdapterRequestDevice)(WGPUAdapter adapter, WGPUDeviceDescriptor const * descriptor, WGPURequestDeviceCallback callback, void * userdata);
 
 // Procs of Buffer
@@ -1022,6 +1035,7 @@ WGPU_EXPORT WGPUProc wgpuGetProcAddress(WGPUDevice device, char const * procName
 
 // Methods of Adapter
 WGPU_EXPORT void wgpuAdapterGetProperties(WGPUAdapter adapter, WGPUAdapterProperties * properties);
+WGPU_EXPORT uint32_t wgpuAdapterGetFeatures(WGPUAdapter adapter, WGPUFeatureName * features);
 WGPU_EXPORT void wgpuAdapterRequestDevice(WGPUAdapter adapter, WGPUDeviceDescriptor const * descriptor, WGPURequestDeviceCallback callback, void * userdata);
 
 // Methods of Buffer


webgpu.h
```diff
@@ -444,6 +444,16 @@ typedef enum WGPUBufferUsage {
 } WGPUBufferUsage;
 typedef WGPUFlags WGPUBufferUsageFlags;
 
+typedef enum WGPUFeatureName {
```
kainino0x, line 447: nit: should we reserve 0 for undefined even though I'm pretty sure we'll never use it? It at least means default-initialization will error instead of enabling DepthClamping.

austinEng: That would make sense to me. Would the same argument apply for `WGPUTextureDimension_1D`, `WGPUTextureComponentType_Float`, (everything else?)

kainino0x: I think so, I think we agreed to do that... somewhere.

austinEng: Ok, we can do that in a separate PR if that's the decision. I think *maybe* it's not because we wanted zero-initialization of other structs ot have reasonable defaults...? maybe.

Kangz: We discussed it in https://github.com/webgpu-native/webgpu-headers/issues/45 but didn't make the change quite yet.

webgpu.h
```diff
@@ -881,6 +894,7 @@ typedef WGPUProc (*WGPUProcGetProcAddress)(WGPUDevice device, char const * procN
 
 // Procs of Adapter
 typedef void (*WGPUProcAdapterGetProperties)(WGPUAdapter adapter, WGPUAdapterProperties * properties);
+typedef uint32_t (*WGPUProcAdapterGetFeatures)(WGPUAdapter adapter, WGPUFeatureName * features);
```
Kangz, line 897: Are the features retrieved as a separate entrypoint or do we augment `WGPUAdapterProperties`? I kind of had the idea that `WGPUAdapterProperties` would be a structure that contains everything, and could also be augmented with a chain, so that the application would look into the chain similarly to what the implementation does.

austinEng: Wouldn't we need to still have a separate entrypoint to find out how many features there are so the application can allocate enough memory to hold the list? So it would sort of be the same thing but populating the list inside a chained struct. 

Overall I like it separate to match the JS API better.

Kangz: The `WGPUAdapterProperties` could have a `featureNameCount` and a `featureNames` members. The pointers would have the same lifetime as the `WGPUAdapter`. But I don't have a super strong opinion at the moment.

austinEng: Ah I see. I do see that we probably don't need to have the two-call mechanism to get the size and populate. The adapter probably needs to store the list anyway.

I still think it should be a separate entrypoint though to match JS better.

```suggestion
typedef void (*WGPUProcAdapterGetFeatures)(WGPUAdapter adapter, WGPUFeatureName ** features, uint32_t * featuresCount);
```

webgpu.h
```diff
@@ -444,6 +444,17 @@ typedef enum WGPUBufferUsage {
 } WGPUBufferUsage;
 typedef WGPUFlags WGPUBufferUsageFlags;
 
+typedef enum WGPUFeatureName {
```
kvark, line 447: could we consider making this a bitmask?

austinEng: Wouldn't that limit us to just 32 features? I think we could easily end up with more than that, especially if vendors/implementations start experimenting with their own internal extensions.

kainino0x: 64 maybe, but I still don't think that's enough.

webgpu.h
Kangz: Sounds good, I'm just we're going to look at the adapter API again in the future when we get closer to a 1.0 webgpu.h

webgpu.h
kvark: Another option is to follow Vulkan here: have a structure with boolean fields for features.

Kangz: Does this mean that like Vulkan we're going to add a chained substructure that contains a single boolean for each feature? Adding more member to the structure would break the ABI

kvark: If a feature is added, in most cases it would need some API changes as well, so some chained structs are inevitable. Would it make sense to make the feature list an exception from this?

austinEng: What benefit does using the chained structs offer us?

kvark: It's a mechanism we already have to have. So why not use it, instead of trying to invent a separate mechanism for growing the feature lists?

austinEng: Returning a list via chained structs makes the API unnecessarily difficult to use when the concept of a list is fairly straightforward.

We use extension structs in places where additional members are added to descriptors. This makes sense, because while the JS API can simply add additional members for extensions, it's impossible to do the same thing in the C API without breaking ABI compatibility.

It would be a divergence from the JS API to convert the list of features at `GPUAdapter.features` to a chain of structs when the issue of extending without breaking ABI compatibility does not apply. We can do a direct translation of list->list.

webgpu.h
kvark: Writing down `WGPUFeatureName**` seems to be quite weird. I don't think that's close to what any of the native APIs do.

austinEng: Well, we could keep the two-call mechanism then. Either with the previous suggestion or like 

```c
typedef void (*WGPUProcAdapterGetFeatures)(WGPUAdapter adapter, WGPUFeatureName ** features, uint32_t * featuresCount);
```

this is what [vkEnumerateDeviceExtensionProperties](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkEnumerateDeviceExtensionProperties.html) does.

kvark: Another option is to just add `featureCount` field to `WGPUAdapterProperties`.

austinEng: Yes, and Corentin even suggested putting *both* the pointer and the count on `WGPUAdapterProperties`.  I don't really feel that strongly about this, but the count is analogous to asking for `adapter.features.length` in JS and not `adapter.properties.featuresLength`, so it makes more sense to me to keep the pointer to the features list and the size of the features list in the same place.

kainino0x: Maybe this would a bit divergent, but this seems like a very reasonable place to use a null-terminated list. Do we have other places in the C API that output lists?

kvark: I would not expect the other places, no

austinEng: I forgot about this for a long time. I changed this so that support for features is queried individually. I realized it's not as useful to have a list when we'd want a more set-like interface. This better matches GPUSupportedFeatures in the JS API, and my suggested native limit API in #102.

kvark (approved):

webgpu.h
kainino0x: The WebIDL setlike thing is still iterable though.

PR #76, opened 2021-01-25
Updates for upstream WebGPU changes:

Kangz:
- Changes from the flat to the nested "WGPUBindGroupLayoutEntry"
 - Add the index format to wgpuRenderPassEncoderSetIndexBuffer
 - Change maxAnisotropy to uint16_t
 - Add WGPUTextureCopyView::aspect

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -108,19 +108,6 @@ typedef enum WGPUBackendType {
     WGPUBackendType_Force32 = 0x7FFFFFFF
 } WGPUBackendType;
 
-typedef enum WGPUBindingType {
-    WGPUBindingType_UniformBuffer = 0x00000000,
-    WGPUBindingType_StorageBuffer = 0x00000001,
-    WGPUBindingType_ReadOnlyStorageBuffer = 0x00000002,
-    WGPUBindingType_Sampler = 0x00000003,
-    WGPUBindingType_ComparisonSampler = 0x00000004,
-    WGPUBindingType_SampledTexture = 0x00000005,
-    WGPUBindingType_MultisampledTexture = 0x00000006,
-    WGPUBindingType_ReadOnlyStorageTexture = 0x00000007,
-    WGPUBindingType_WriteOnlyStorageTexture = 0x00000008,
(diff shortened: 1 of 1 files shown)


webgpu.h
```diff
@@ -752,12 +786,14 @@ typedef struct WGPUBindGroupDescriptor {
...
-typedef struct WGPUBindGroupLayoutDescriptor {
-    WGPUChainedStruct const * nextInChain;
-    char const * label;
-    uint32_t entryCount;
-    WGPUBindGroupLayoutEntry const * entries;
-} WGPUBindGroupLayoutDescriptor;
+typedef struct WGPUBindGroupLayoutEntry {
+    uint32_t binding;
+    WGPUShaderStageFlags visibility;
+    WGPUBufferBindingLayout buffer;
```
kvark, line 792: Could this be a `union`?

Kangz: It could be it would need a tag, and unions in general are pretty unsafe in C and C++, plus this is pretty cold code. We had some discussion about this topic with @kainino0x and @toji and came to the solution in this PR.

kvark: Another thing I don't understand is each of `WGPUBufferBindingLayout`, `WGPUSamplerBindingLayout`, and `WGPUTextureBindingLayout` being extendable, but `WGPUBindGroupLayoutEntry` itself - not extendable. Would it be easier to make it the opposite?

Kangz: Uh, WGPUBindGroupLayoutEntry should definitely be extendable, that's a mistake. It might be useful to have eachsub-structure are extendable but I see how that could be unnecessary and using space for nothing. We could probably remove it. @toji WDYT?

kainino0x: I think both should be extendable, it's a small overhead (cold code, like you said) and I think it's likely that most extensions would show up in the inner structures.

toji: I don't have anything against it being extendable.

Kangz: Made it extendable.

kvark (approved):

Kangz:
Yay!

PR #78, opened 2021-02-02
Remove Fence in favor of QueueOnSubmittedWorkDone

Kangz:
This has a signaledValue argument contrary to the current WebGPU spec in
preparation for it being added in the future when WebGPU gains
multi-queue support. For now it should be invalid to pass any value other
than 0, when QueueSignal is added, 0 should mean "the latest signaled
value".

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -66,7 +66,6 @@ typedef struct WGPUCommandEncoderImpl* WGPUCommandEncoder;
 typedef struct WGPUComputePassEncoderImpl* WGPUComputePassEncoder;
 typedef struct WGPUComputePipelineImpl* WGPUComputePipeline;
 typedef struct WGPUDeviceImpl* WGPUDevice;
-typedef struct WGPUFenceImpl* WGPUFence;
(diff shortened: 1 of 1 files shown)


kvark:
You are pushing for delaying multi-queue (which is totally reasonable to delay!), and at the same time trying to get the API in here before we get it? Are you trying to reduce the future breakage for people?

Kangz:
Exactly! We cannot do overloads in C so adding the signal value now is to avoid having two different functions in the C API when we add multi-queue.

kvark (approved):
okie-dokie

Kangz:
Thanks, @kainino0x @austinEng WDYT?

austinEng (approved):
Looks good.

It will be interesting to design how Signal is going to work eventually. The old way we used to have it doesn't look like it would let you read the last signaled value and signal a new value, all atomically.

kainino0x (approved):

Kangz:
Yeah we should revisit this when we try to stabilize webgpu.h to make sure it is in line with the future idea for multi-queue. Thanks all!

PR #81, opened 2021-03-08
Renderpipeline descriptor

Kangz:
PTAL at the second commit. +CC @toji

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -164,14 +164,14 @@ typedef enum WGPUCompareFunction {
     WGPUCompareFunction_Force32 = 0x7FFFFFFF
 } WGPUCompareFunction;
 
-typedef enum WGPUCreateReadyPipelineStatus {
-    WGPUCreateReadyPipelineStatus_Success = 0x00000000,
-    WGPUCreateReadyPipelineStatus_Error = 0x00000001,
-    WGPUCreateReadyPipelineStatus_DeviceLost = 0x00000002,
-    WGPUCreateReadyPipelineStatus_DeviceDestroyed = 0x00000003,
-    WGPUCreateReadyPipelineStatus_Unknown = 0x00000004,
-    WGPUCreateReadyPipelineStatus_Force32 = 0x7FFFFFFF
-} WGPUCreateReadyPipelineStatus;
+typedef enum WGPUCreatePipelineAsyncStatus {
+    WGPUCreatePipelineAsyncStatus_Success = 0x00000000,
+    WGPUCreatePipelineAsyncStatus_Error = 0x00000001,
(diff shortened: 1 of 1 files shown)


kvark (commented):
We should figure out a good set of guidelines on what is passed by pointer, what is passed by value with "bool enabled", and what is passed by value with another internal nullability mechanism. Currently with this PR we a mix (of at least the last 2).

Kangz:
Yeah that's in part what triggered this PR. WDYT?

kvark (commented):
So we have at least these 3 things:
  - fragment state
  - depth-stencil state
  - blend state

It would be nice to cover them consistently.
What was the concern with making them pointers?

Kangz:
As usual there's a concern that making the C descriptor very deep (in terms of pointer indirection) makes them very annoying to use. When the default can be done  via 0-initialization, we have an opportunity to reduce that nesting and make it much nicer to write WebGPU.

kvark:
Who is going to write C code for talking to WebGPU though? I could see two groups of users:
  1. direct users via Rust/C++/Python/etc - these would not be affected.
  2. engine writers, for whom this is a translation step, so they aren't going to be affected either, and they are going to need to deal with this difference exactly once in the lifetime.

Kangz:
People use the C API (like sokol-gfx) and at least at the moment Dawn's C++ WebGPU wrapper is a trivial wrapper around the C API so ergonomics are directly linked to the C API's ergonomics. And for C++ it would be unfortunate to have to iterate over the whole render pipeline descriptor to pointerize it for C. There are also engine writers that use C and could replace their RHI with WebGPU, so the C API being usable matters.

Kangz:
@toji @austinEng we talked about this offline with @kvark and I described our problem with the C++ header being very close to the C API so the ergonomics of the C API are important for us. Basically since we already are using a `ComboRenderPipelineDescriptor` we could handle the additional pointerization by making that class smarter. For example a call `descriptor.UseDepthStencilAttachment(wgpu::TextureFormat)` could do the linkage of the pointer at the same time as it sets the format so that it look nice. Since we have to revisit all pipeline descriptors anyway in Dawn, we can do the transform to use the more featureful `RenderPipelineDescriptorBuilder` (or whatever we call it).

For this PR we should "just" figure out what additional things we'd like to be pointers and change. @kvark mentioned fragmentState, depthStencilState and blend. Is there something else?

Kangz:
Added more pointers. Is that sufficient or should `multisample` and `primitive` also be pointers?

kvark (approved):

kvark:
Thank you!
> Is that sufficient or should multisample and primitive also be pointers?

They aren't nullable in the spec, they have defaults. So they are still values, no need for pointers there (phew!).

Kangz:
Great, thank you!

PR #83, opened 2021-04-01
WGPUExtent.depth -> depthOrArrayLayers

Kangz:
PTAL!


Kangz:
Reviewed as part of #85

PR #87, opened 2021-04-21
Compilation info

Kangz:
PTAL!

Do you think that `WGPUCompilationInfo` should be extensible so that the WebGPU library can pass more information to the application in the future?


kvark (approved):

kvark:
Yes, making it extensible sounds reasonable.

Kangz:
Also maybe for WGPUCompilationMessage. @toji WDYT?

toji (approved):
LGTM. And yes, it's probably a decent idea for at least CompilationInfo to be extensible. Feels like making CompilationMessage extensible may just be unnecessary complexity. If worst comes to worst you could always add a parallel array to CompilationInfo to contain additional per-message data.

PR #93, opened 2021-08-23
Updates to follow the specification:

Kangz:
- StoreOp_Clear -> Discard
 - TextureUsage_Sampled -> TextureBinding
 - TextureUsage_Storage -> StorageBinding
 - InputStepMode -> VertexStepMode

Other updates:
 - Put operation first in BlendComponent because it is the operation
   that uses the factors so it makes sense that it would be first.
 - Add missing nextInChain to BGEntry.

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -215,12 +215,6 @@ typedef enum WGPUIndexFormat {
     WGPUIndexFormat_Force32 = 0x7FFFFFFF
 } WGPUIndexFormat;
 
-typedef enum WGPUInputStepMode {
-    WGPUInputStepMode_Vertex = 0x00000000,
-    WGPUInputStepMode_Instance = 0x00000001,
-    WGPUInputStepMode_Force32 = 0x7FFFFFFF
-} WGPUInputStepMode;
-
 typedef enum WGPULoadOp {
     WGPULoadOp_Clear = 0x00000000,
     WGPULoadOp_Load = 0x00000001,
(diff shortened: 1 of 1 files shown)


kvark (approved):

kvark:
Thank you!

webgpu.h
```diff
@@ -529,9 +530,9 @@ typedef struct WGPUBindGroupEntry {
 } WGPUBindGroupEntry;
 
 typedef struct WGPUBlendComponent {
+    WGPUBlendOperation operation;
     WGPUBlendFactor srcFactor;
     WGPUBlendFactor dstFactor;
-    WGPUBlendOperation operation;
 } WGPUBlendComponent;
```
kainino0x, line 536: nit: spec actually has the original order: srcFactor, dstFactor, operation.

I'm going to open a PR with a few other things so I'll include it.

webgpu.h
```diff
@@ -806,11 +807,16 @@ typedef struct WGPUBlendState {
     WGPUBlendComponent alpha;
 } WGPUBlendState;
 
+typedef struct WGPUCompilationInfo {
+    uint32_t messageCount;
+    WGPUCompilationMessage const * messages;
```
kainino0x, line 812: WGPUCompilationMessage isn't defined yet, and this isn't used

webgpu.h
kvark: oops

webgpu.h
Kangz: Whoops

Issue #82, opened 2021-03-31
Determine how to return CompilationInfo


toji:
CompilationInfo is a little unique because it's one of the only places in the API that a structure of data is being returned out of the API. AdapterProperties seems to be the only close analog, but because CompilationInfo is queried with a promise in the JS API it would need to be supplied by a callback here.

So a first stab at this would suggest we'd want an API surface like that following?

```C
typedef enum WGPUCompilationMessageType {
    WGPUCompilationMessageType_Undefined = 0x00000000,
    WGPUCompilationMessageType_Error = 0x00000001,
    WGPUCompilationMessageType_Warning = 0x00000002,
    WGPUCompilationMessageType_Info = 0x00000003,
} WGPUCompilationMessageType;

typedef struct WGPUCompilationMessage {
    WGPUChainedStruct const * nextInChain; // Are these extensible?
    char const * message;
    WGPUCompilationMessageType type;
    uint64_t lineNum;
    uint64_t linePos;
} WGPUCompilationMessage;

typedef struct WGPUCompilationInfo {
    WGPUChainedStruct const * nextInChain; // Are these extensible?
    WGPUCompilationMessage * messages;
    uint32_t messageCount;
} WGPUCompilationInfo ;

typedef void (*WGPUCompilationInfoCallback )(WGPUCompilationInfo* compilationInfo, void * userdata);

WGPU_EXPORT void wgpuShaderModuleGetCompilationInfo(WGPUShaderModule shaderModule, WGPUCompilationInfoCallback callback, void * userdata);
```

Or, recognizing that WGPUCompilationInfo is just a list of messages, if we weren't worried about extensibility of that particular structure the callback signature could simply be:

```C
typedef void (*WGPUCompilationInfoCallback )(WGPUCompilationMessage* messages, uint32_t messageCount, void * userdata);
```

I'm also not sure what sort of guarantees the library gives about lifetime of objects like this, if it's expected that it would only be valid for the duration of the callback or if the library would be expected to keep it alive and owned by the shader module after being queried.

austinEng:
Overall, I think we should keep the extensibility you have proposed. I could definitely imagine extensions like WEBGL_debug_shaders.getTranslatedShaderSource() that might fit nicely here.

[...]

PR #96, opened 2021-08-23
Small updates to match spec more closely

kainino0x:
Also adds the chain of types used by WGPUCompilationInfo (but still no
way to actually use it), and makes them extensible. #82 #87

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -539,9 +539,9 @@ typedef struct WGPUBindGroupEntry {
 } WGPUBindGroupEntry;
 
 typedef struct WGPUBlendComponent {
+    WGPUBlendOperation operation;
     WGPUBlendFactor srcFactor;
     WGPUBlendFactor dstFactor;
-    WGPUBlendOperation operation;
 } WGPUBlendComponent;
 
 typedef struct WGPUBufferBindingLayout {
--- a/webgpu.h
+++ b/webgpu.h
@@ -164,6 +164,13 @@ typedef enum WGPUCompareFunction {
     WGPUCompareFunction_Force32 = 0x7FFFFFFF
 } WGPUCompareFunction;
 
+typedef enum WGPUCompilationMessageType {
+    WGPUCompilationMessageType_Error = 0x00000000,
+    WGPUCompilationMessageType_Warning = 0x00000001,
+    WGPUCompilationMessageType_Info = 0x00000002,
+    WGPUCompilationMessageType_Force32 = 0x7FFFFFFF
+} WGPUCompilationMessageType;
+
 typedef enum WGPUCreatePipelineAsyncStatus {
     WGPUCreatePipelineAsyncStatus_Success = 0x00000000,
     WGPUCreatePipelineAsyncStatus_Error = 0x00000001,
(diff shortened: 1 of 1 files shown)


webgpu.h
```diff
@@ -366,24 +373,25 @@ typedef enum WGPUTextureFormat {
...
-    WGPUTextureFormat_BC5RGSnorm = 0x00000032,
-    WGPUTextureFormat_BC6HRGBUfloat = 0x00000033,
-    WGPUTextureFormat_BC6HRGBFloat = 0x00000034,
-    WGPUTextureFormat_BC7RGBAUnorm = 0x00000035,
-    WGPUTextureFormat_BC7RGBAUnormSrgb = 0x00000036,
+    WGPUTextureFormat_Stencil8 = 0x00000025,
+    WGPUTextureFormat_Depth16unorm = 0x00000026,
+    WGPUTextureFormat_Depth24Plus = 0x00000027,
+    WGPUTextureFormat_Depth24PlusStencil8 = 0x00000028,
+    WGPUTextureFormat_Depth32Float = 0x00000029,
```
kainino0x, line 380: Should we leave any space in the texture format list? Or should we just add new ones to the end? I know holes can be bad for switch performance.

webgpu.h
```diff
@@ -1065,7 +1084,7 @@ typedef void (*WGPUProcRenderPassEncoderWriteTimestamp)(WGPURenderPassEncoder re
 typedef WGPUBindGroupLayout (*WGPUProcRenderPipelineGetBindGroupLayout)(WGPURenderPipeline renderPipeline, uint32_t groupIndex);
 
 // Procs of Surface
-typedef void (*WGPUProcSurfaceGetPreferredFormat)(WGPUSurface surface, WGPUAdapter adapter, WGPUSurfaceGetPreferredFormatCallback callback, void * userdata);
+typedef WGPUTextureFormat (*WGPUProcSurfaceGetPreferredFormat)(WGPUSurface surface, WGPUAdapter adapter);
```
kainino0x, line 1087: Do these need to stay async?

webgpu.h
```diff
@@ -164,6 +164,13 @@ typedef enum WGPUCompareFunction {
     WGPUCompareFunction_Force32 = 0x7FFFFFFF
 } WGPUCompareFunction;
 
+typedef enum WGPUCompilationMessageType {
+    WGPUCompilationMessageType_Error = 0x00000000,
+    WGPUCompilationMessageType_Warning = 0x00000001,
+    WGPUCompilationMessageType_Info = 0x00000002,
```
kainino0x, line 170: Which option should be 0?

Kangz: Probably doesn't matter that much given it is only data passed from the implementation to the application.

webgpu.h
Kangz: I think we want all formats for v1 in a reasonable order then add at the end when new ones come up?

webgpu.h
```diff
@@ -530,9 +539,9 @@ typedef struct WGPUBindGroupEntry {
 } WGPUBindGroupEntry;
 
 typedef struct WGPUBlendComponent {
-    WGPUBlendOperation operation;
     WGPUBlendFactor srcFactor;
     WGPUBlendFactor dstFactor;
+    WGPUBlendOperation operation;
```
Kangz, line 544: nit: I like marginally better before:

```
alphaBlend = {wgpu::BlendOperation::Add, wgpu::BlendFactor::One, wgpu::BlendFactor::DstAlpha}
// is more understandable than
alphaBlend = {wgpu::BlendFactor::One, wgpu::BlendFactor::DstAlpha, wgpu::BlendOperation::Add}

```

webgpu.h
Kangz: Probably not. But the whole swapchain aspect must be redone at some point anyway.

Kangz (approved):

webgpu.h
kvark: does it matter for this enum?

webgpu.h
kvark: technically speaking, we never match on a texture format in hot code where this difference would matter

webgpu.h
kvark: no, it's not async in the spec

kvark (approved):

webgpu.h
kainino0x: agree it probably doesn't matter.

webgpu.h
kainino0x: I'd be happy to change this back, and we can reorder upstream too

webgpu.h
kainino0x: We might internally though. It would be annoying to have to translate to an internal packed enum (but probably ok if we want to block these out.

I was thinking 0x000 for core color formats, 0x100 for depth/stencil formats, 0x200 for extensions, or something like that. I wonder if compilers can handle a few blocks of values well.

Kangz: I have no real opinion here: texture formats should be relatively cold code in Dawn too since we cache a pointer to dawn_native::Format in the texture and could do the same in pipeline if we wanted.

PR #100, opened 2021-09-02
Add drop method to wgpu objects

Ldash4:
It was suggested that I make this PR here: https://github.com/gfx-rs/wgpu-native/pull/128


kvark (approved):
Thank you!
I'm open to a different naming scheme, if anybody has ideas :)

Kangz:
This is mostly a duplicate of https://github.com/webgpu-native/webgpu-headers/pull/94

I don't think we should merge it before we agree on lifetime management between implementations. We feel strongly that refcounting is needed for many objects, and `wgpu` can support that with no cost. `drop` doesn't have that semantic.

kvark:
@Kangz this is not a duplicate. It's exactly half of #94 (unref but not the addref). What's most important, it's the half that we *do* agree on. So I see no reason to block this from merging.

Kangz:
The name changes the semantic significantly. And I find the push to just get stuff landed quite frustrating given we've waited a whole 3 years for wgpu to figure out their lifetime semantics.

kvark:
I don't think the 3 years mark is relevant here. I understand the frustration, but the 3 year discussion was about stuff we couldn't agree on. And here we are talking about something we actually do agree on. If you suggested this 3 years ago, we'd happily merge!
Let's bikeshed the name then. Would you consider anything other than "Release" here, or are you strongly attached to this?

PR #101, opened 2021-09-05
Add PowerPreference enum

radgeRayden:
As mentioned in https://github.com/gfx-rs/wgpu-native/issues/122 and seen in the [spec](https://www.w3.org/TR/webgpu/#adapter-selection).

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -244,6 +244,12 @@ typedef enum WGPUPipelineStatisticName {
     WGPUPipelineStatisticName_Force32 = 0x7FFFFFFF
 } WGPUPipelineStatisticName;
 
+typedef enum WGPUPowerPreference {
+    WGPUPowerPreference_LowPower = 0x00000000,
+    WGPUPowerPreference_HighPerformance = 0x00000001,
+    WGPUPowerPreference_Force32 = 0x7FFFFFFF 
+} WGPUPowerPreference;
+
 typedef enum WGPUPresentMode {
     WGPUPresentMode_Immediate = 0x00000000,
     WGPUPresentMode_Mailbox = 0x00000001,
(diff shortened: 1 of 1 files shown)


kvark (approved):
Thank you!

Kangz (approved):
LGTM, although we'll likely need to reword `WGPURequestAdapterOptions` a bunch once WebGPU v1 is out to match the spec. (also `compatibleSurface` could be an extension as it's not a thing on the Web)

kvark:
> although we'll likely need to reword WGPURequestAdapterOptions a bunch once WebGPU v1 is out to match the spec

What is not matching the spec?

Kangz:
I think there is more adapter request options, and compatible surface
doesn't exist in WebGPU.

austinEng:
Adapter options are pretty different for web vs. native. We may want to have separate extension struct options for both of them.

kvark:
I'm confused by what you folks are saying.

> Adapter options are pretty different for web vs. native

What's the deal with just exposing the adapter options in the way upstream has it?

> I think there is more adapter request options

This PR isn't about bringing the whole adapter options, it's only about power preference.

> compatible surface doesn't exist in WebGPU

Given that a) - it's pretty much required for native, and b) it's harmless for targeting the web, and c) the header is used by native-first applications, it seems to me that there is nothing wrong in exposing the compatible surface just as a field, not as an extension.

Kangz:
> What's the deal with just exposing the adapter options in the way upstream has it?

> This PR isn't about bringing the whole adapter options, it's only about power preference.

Yeah that's why I approved it with no questions. The comments are that we eventually need to decide how Web vs. native vs. common stuff is exposed at that level: we probably don't want them all on the same descriptor.

PR #102, opened 2021-09-08
Add limits

austinEng:
Limits are represented by an enum WGPULimitName.

Limits can be queried using wgpuAdapterGetLimitU32 or with
wgpuAdapterGetLimitU64, depending on the type of the limit.
These functions return false if the implementation does not
recognize the limit name, or if the limit type does not match.

Limits can be requested by passing a list of WGPULimitEntry in
the WGPUDeviceDescriptor. Values are all uint64_t, but the
implementation validates that they are within the range of the
expected limit type.

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -229,6 +229,37 @@ typedef enum WGPUIndexFormat {
     WGPUIndexFormat_Force32 = 0x7FFFFFFF
 } WGPUIndexFormat;
 
+typedef enum WGPULimitName {
+    WGPULimitName_Undefined = 0x00000000,
+    WGPULimitName_MaxTextureDimension1D = 0x00000001,
+    WGPULimitName_MaxTextureDimension2D = 0x00000002,
+    WGPULimitName_MaxTextureDimension3D = 0x00000003,
+    WGPULimitName_MaxTextureArrayLayers = 0x00000004,
+    WGPULimitName_MaxBindGroups = 0x00000005,
+    WGPULimitName_MaxDynamicUniformBuffersPerPipelineLayout = 0x00000006,
+    WGPULimitName_MaxDynamicStorageBuffersPerPipelineLayout = 0x00000007,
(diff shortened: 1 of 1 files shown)


kvark:
Could you expand on the motivation about why the limits aren't just fields in a struct?

austinEng:
> Could you expand on the motivation about why the limits aren't just fields in a struct?

This was my thinking:

1. if we return a struct, we break ABI compatibility every time we add a new limit.
I think new limits are likely as the API expands. Some alternatives, but I think both are messy or relatively difficult to use:
    - we could version our header and put things behind #ifdefs
    - We could do it like Vulkan and have multiple chained limits structs like VkPhysicalDeviceProperties

2. One day we will need limits for things provided by optional features. This API is extensible to support that. Alternatively, we could have a separate struct and limit query for each optional feature.

kvark:
It's an interesting question. With WebGPU 1.0, can we really add limits without breaking it? Adding a limit means that there is a new restriction, and there might be content out there, which was valid before this limit is added, but then stops being valid. So it seems to me that adding limits is *always* a breaking change anyway.

Kangz:
Adding limits to the existing WebGPU v1 would be a backwards incompatible (though TBH it's fairly likely it will happen at least once because we missed something). However we need a mechanism for optional feature limits and further version limits.

kvark:
Further version limits would just be more fields in the struct.
Extension limits can be the chained structs. The same argument applies (as I mentioned in another discussion): if we already have an extension mechanism, why not use it consistently for everything? Inventing a *separate* extension mechanism for one specific part of the API seems like an unnecessary complication.

austinEng:
The argument that adding core limits should be pretty rare once v1 is stable and optional features can use extension structs makes sense to me. I'll make a PR for that.

austinEng:
Opened #104

almarklein:
Isn't this superseded by #104 ?

austinEng:
yes

PR #104, opened 2021-09-09
Add base WGPULimits structs

austinEng:
This struct may have additional chained structs to request additional
limits that are added to the API or are present in optional features.

An adapter's supported limits may be queried using wgpuAdapterGetLimits.
This function populates the provided limits and any chained limit
structs. It is invalid to pass chained limit structs here for features
that are not supported by the adapter.

requiredLimits may be passed in the WGPUDeviceDescriptor to request
specific non-default limits. Since this is a struct, there is no way to
optionally include members. #defines WGPU_LIMIT_{type}_UNDEFINED may be
used to omit a limit from the request and get the default instead.
It is invalid to pass chained limit structs here for features that are
not supported by the adapter.

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -54,6 +54,8 @@
 
 #define WGPU_WHOLE_SIZE (0xffffffffffffffffULL)
 #define WGPU_COPY_STRIDE_UNDEFINED (0xffffffffUL)
+#define WGPU_LIMIT_U32_UNDEFINED (0xffffffffUL)
+#define WGPU_LIMIT_U64_UNDEFINED (0xffffffffffffffffULL)
 
 typedef uint32_t WGPUFlags;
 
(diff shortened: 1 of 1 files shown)


kvark (approved):

webgpu.h
```diff
@@ -1152,6 +1186,7 @@ WGPU_EXPORT WGPUProc wgpuGetProcAddress(WGPUDevice device, char const * procName
 // Methods of Adapter
 WGPU_EXPORT void wgpuAdapterGetProperties(WGPUAdapter adapter, WGPUAdapterProperties * properties);
 WGPU_EXPORT bool wgpuAdapterHasFeature(WGPUAdapter adapter, WGPUFeatureName feature);
+WGPU_EXPORT void wgpuAdapterGetLimits(WGPUAdapter adapter, WGPULimits * limits);
```
Kangz, line 1189: nit: I think we should return a boolean here to say if the call succeeded or not given the validation rules you  described.

Kangz (approved):
LGTM

PR #106, opened 2021-09-14
Add constants to vertex state and fragment state

shrekshao:
Follow up of https://github.com/webgpu-native/webgpu-headers/pull/99

Realized WGPUProgrammableStageDescriptor has no inheritance here as in spec during implementation.
Add the same `constantCount` and `constants` fields to `WGPUVertexState` and `WGPUFragmentState`

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -1001,6 +1001,8 @@ typedef struct WGPUVertexState {
     WGPUChainedStruct const * nextInChain;
     WGPUShaderModule module;
     char const * entryPoint;
+    uint32_t constantCount;
+    WGPUConstantEntry const * constants;
     uint32_t bufferCount;
     WGPUVertexBufferLayout const * buffers;
 } WGPUVertexState;
@@ -1009,6 +1011,8 @@ typedef struct WGPUFragmentState {
     WGPUChainedStruct const * nextInChain;
     WGPUShaderModule module;
     char const * entryPoint;
+    uint32_t constantCount;
+    WGPUConstantEntry const * constants;
     uint32_t targetCount;
     WGPUColorTargetState const * targets;
 } WGPUFragmentState;


webgpu.h
```diff
@@ -1001,6 +1001,8 @@ typedef struct WGPUVertexState {
     WGPUChainedStruct const * nextInChain;
     WGPUShaderModule module;
```
kvark, line 1002: can we put `module`, `entryPoint`, and the constant stuff into a sub-structure?

shrekshao: That is reasonable. But without inheritance here I think we will have quite some change on the implementation side?
something like this?
```c
typedef struct WGPUProgrammableStageBase {
    WGPUShaderModule module;
    char const * entryPoint;
    uint32_t constantCount;
    WGPUConstantEntry const * constants;
} WGPUProgrammableStageBase;

typedef struct WGPUVertexState {
    WGPUProgrammableStageBase programmableStageBase;
    uint32_t bufferCount;
    WGPUVertexBufferLayout const * buffers;
} WGPUVertexState;
```

kainino0x: fwiw I'm OK with either. I don't think it's too much of a burden to duplicate. It depends on whether we want to match the upstream JS-visible API (duplicate the fields) or the upstream WebIDL (use a base struct).

Kangz: imho flattening is better, since that matches the what JS developers type in their code:

```js
const pipeline = device.createPipeline({
   layout: myLayout,
   // ...
});
// and not
const pipeline = device.createPipeline({
   pipelineBase: {
      layout: myLayout,
   },
   // ...
});
```

kvark: JS developers have inheritance. In C, this is composition. It's ok to have slight differences like this given this is an entirely different language. Making C API ergonomic is also less of a priority: we should expect people using it from all sorts of languages (Python, C++, bunch of others).
In this case, a 4 field structure like @shrekshao suggestion would be perfect. Maybe just `WGPUProgrammableStage` instead of `WGPUProgrammableStageBase`?

Kangz: Uuuh we already discussed this somewhat in https://github.com/webgpu-native/webgpu-headers/pull/81 and while the C API can be a bit less ergonomic we shouldn't just make it worse for almost no gain.

The additional pointerization in #81 made sense to express some JS dictionary members being undefined. However I don't think we should nest the inheritance structures from the WebGPU IDL. First this structure might change over time in the spec because it is not user-visible. Second, we already have `label` inlined in every descriptor so we would need to add an `ObjectBaseDescriptor` instead and this looks silly. Third if we want to match the IDL, then we would have to merge some methods: `wgpuRenderBundleDraw` and `wgpuRenderPassEncoderDraw` or all the `SetLabel` and that would require dynamic dispatch in implementations.

Basically it is fine to diverge slightly from the structure of the WebIDL and we should flatten.

kvark: If we are to choose one way consistently across the whole API, then I agree that this would be flattening instead of compositing. Those `ObjectBaseDescriptor` are too annoying.
However, if we are to make a single exception from this, `WGPUProgrammableStage` would be perfect fit.
All in all, I don't want to push strongly here. If you don't feel positive about a substruct, then flat be it.

kainino0x: > JS developers have inheritance.

WebIDL gives us the spec authors inheritance; JS developers can't tell the difference between inheritance and inline duplication.

shrekshao: Thanks for the inputs. Seems the conclusion is to keep the flatten style.

Kangz: Seems like so, let's merge!

Kangz (approved):

PR #107, opened 2021-09-14
Add wgpuDeviceGetLimits; Split WGPURequiredLimits/WGPUSupportedLimits

austinEng:
This is to match GPUDevice.limits in the JS API. In contains the
actual limits exposed on that device, whereas wgpuAdapterGetLimits
returns the best possible limits on an adapter.

This PR also splits the limits into WGPURequiredLimits and WGPUSupportedLimits
where WGPURequiredLimits is const, but WGPUSupportedLimits is non-const so it
can be populated by wgpuAdapterGetLimits/wgpuDeviceGetLimits.

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -567,6 +567,35 @@ typedef struct WGPUAdapterProperties {
     WGPUBackendType backendType;
 } WGPUAdapterProperties;
 
+typedef struct WGPUBaseLimits {
+    uint32_t maxTextureDimension1D;
+    uint32_t maxTextureDimension2D;
+    uint32_t maxTextureDimension3D;
+    uint32_t maxTextureArrayLayers;
+    uint32_t maxBindGroups;
+    uint32_t maxDynamicUniformBuffersPerPipelineLayout;
+    uint32_t maxDynamicStorageBuffersPerPipelineLayout;
+    uint32_t maxSampledTexturesPerShaderStage;
+    uint32_t maxSamplersPerShaderStage;
+    uint32_t maxStorageBuffersPerShaderStage;
+    uint32_t maxStorageTexturesPerShaderStage;
+    uint32_t maxUniformBuffersPerShaderStage;
+    uint64_t maxUniformBufferBindingSize;
+    uint64_t maxStorageBufferBindingSize;
+    uint32_t minUniformBufferOffsetAlignment;
+    uint32_t minStorageBufferOffsetAlignment;
+    uint32_t maxVertexBuffers;
+    uint32_t maxVertexAttributes;
+    uint32_t maxVertexBufferArrayStride;
+    uint32_t maxInterStageShaderComponents;
+    uint32_t maxComputeWorkgroupStorageSize;
+    uint32_t maxComputeInvocationsPerWorkgroup;
+    uint32_t maxComputeWorkgroupSizeX;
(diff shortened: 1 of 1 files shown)


kvark (approved):

kvark:
The subject + body are a bit misleading?

austinEng:
> The subject + body are a bit misleading?

not sure I understand what you're referring to

kvark:
nvm, I misread 😓

kainino0x:
Doesn't WGPULimits.nextInChain need to be non-const so GetLimits can write into it? (Or maybe I'm misunderstanding the constness)

kvark:
Isn't that great? People wouldn't be able to blindly pass adapter Limits down the road! :)

austinEng:
yea that's probably better actually

So in the latest patchset I split them into WGPUSupportedLimits and WGPURequiredLimits. However, I think we could share code more if we wanted to like:

```
struct WGPUSupportedLimits {
  WGPUChainedStruct * nextInChain;
  WGPULimits limits;
};

WGPURequiredLimits {
  WGPULimits limits;
};
```

kvark:
Yeah, repeating all of them is a waste. I find `WGPULimits limits;` to be a better path forward. Maybe also rename `WGPULimits` to `WGPUBaseLimits`.

austinEng:
Ok, updated! @Kangz ?

kvark (approved):

webgpu.h
```diff
@@ -765,6 +764,11 @@ typedef struct WGPURequestAdapterOptions {
     WGPUPowerPreference powerPreference;
 } WGPURequestAdapterOptions;
 
+typedef struct WGPURequiredLimits {
+    WGPUChainedStruct const * nextInChain;
+    WGPUBaseLimits limits;
```
Kangz, line 769: This is slightly verbose but I don't see a better way.

WGPULimitsV1 maybe? I don't know what will be the best name in the long run.

Kangz (approved):

webgpu.h
kvark: no hard feelings about the name, "LimitsV1" is fine

austinEng: or just "WGPULimits" as it was originally? and then in the future we can have WGPULimits2, WGPURequiredLimits2, WGPUSupportedLimits2. This is like how Vulkan has VkPhysicalDeviceFeatures and VkPhysicalDeviceFeatures2

kvark: ok, fiiiiine :)

kainino0x: I think the extended limits struct will probably be one per optional feature, rather than having a limits2/etc? May use limits2 for limits we add later (hopefully none, except for webgpu-compat which would also have its own struct).

And if those structs are small then it would be kind of annoying to define three structs for every single one. Wish we had a way to avoid it. We could have supported-limits allocated by the implementation (with a separate `free` call), instead of having a non-const pointer chain? Tradeoff is apps would have to deal with reinterpret_casting by SType and skipping STypes they don't understand.

austinEng: There is a struct-less proposal in https://github.com/webgpu-native/webgpu-headers/pull/102. It sort of introduces a new way to handle optional features catered specifically to limits, but it does help resolve this multi-struct annoyance.

kvark (approved):

PR #113, opened 2021-11-09
Updates for the latest spec:

Kangz:
- Adds ETC2 and ASTC formats
 - Adds readOnlyDepth/Stencil to the render bundle encoder descriptor.

Also adds WGPU_MAP_WHOLE_SIZE that's a SIZE_MAX because sizes for mapping
calls are size_t instead of uint64_t for other buffers sizes in the header.


kainino0x (approved):

Kangz:
Merging, I think this is very uncontroversial.

PR #114, opened 2021-11-09
Add GetCompilationInfo (and the respective status)

### LINKED ISSUE #82: Determine how to return CompilationInfo

Kangz:
Fixes #82

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -169,6 +169,14 @@ typedef enum WGPUCompareFunction {
     WGPUCompareFunction_Force32 = 0x7FFFFFFF
 } WGPUCompareFunction;
 
+typedef enum WGPUCompilationInfoRequestStatus {
+    WGPUCompilationInfoRequestStatus_Success = 0x00000000,
+    WGPUCompilationInfoRequestStatus_Error = 0x00000001,
+    WGPUCompilationInfoRequestStatus_DeviceLost = 0x00000002,
+    WGPUCompilationInfoRequestStatus_Unknown = 0x00000003,
+    WGPUCompilationInfoRequestStatus_Force32 = 0x7FFFFFFF
+} WGPUCompilationInfoRequestStatus;
+
 typedef enum WGPUCompilationMessageType {
     WGPUCompilationMessageType_Error = 0x00000000,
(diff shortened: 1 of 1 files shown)


kainino0x (approved):

Kangz:
Merging. I think this is uncontroversial too.

PR #140, opened 2022-01-04
Add WGPUSurfaceDescriptorFromWayland

### LINKED ISSUE #139: Support Wayland
Currently only X11 is supported by `WGPUSurfaceDescriptorFromXlib`.
Any planning on supporting Wayland?

rajveermalviya:
Fixes #139

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -344,6 +344,7 @@ typedef enum WGPUSType {
     WGPUSType_ShaderModuleSPIRVDescriptor = 0x00000005,
     WGPUSType_ShaderModuleWGSLDescriptor = 0x00000006,
     WGPUSType_PrimitiveDepthClipControl = 0x00000007,
+    WGPUSType_SurfaceDescriptorFromWaylandSurface = 0x00000008,
     WGPUSType_Force32 = 0x7FFFFFFF
 } WGPUSType;
 
@@ -906,6 +907,12 @@ typedef struct WGPUSurfaceDescriptorFromXlib {
     void * display;
     uint32_t window;
 } WGPUSurfaceDescriptorFromXlib;
+
+typedef struct WGPUSurfaceDescriptorFromWaylandSurface {
+    WGPUChainedStruct chain;
+    void * display;
+    void * surface;
+} WGPUSurfaceDescriptorFromWaylandSurface;
 
 typedef struct WGPUSwapChainDescriptor {
     WGPUChainedStruct const * nextInChain;


webgpu.h
```diff
@@ -344,6 +344,7 @@ typedef enum WGPUSType {
     WGPUSType_ShaderModuleSPIRVDescriptor = 0x00000005,
     WGPUSType_ShaderModuleWGSLDescriptor = 0x00000006,
     WGPUSType_PrimitiveDepthClipControl = 0x00000007,
+    WGPUSType_SurfaceDescriptorFromWayland = 0x00000008,
```
Kangz, line 347: nit: Maybe `FromWaylandSurface`? I wonder if it should be `FromXlibWindow` as well instead of `FromXlib`.

@kvark WDYT?

Kangz (approved):
Thanks for the contribution! It looks good, only a small naming discussion needed. I'll see if I can implement that in Dawn quickly afterwards, it should be easy, unless you want to give it ia try?

webgpu.h
kvark: current seems fine. Technically Vulkan would call it `WaylandSurfaceDescriptor`, etc, but personally I like the uniformity of putting the variations of this structure at the end ("...FromXxx") instead of prefixing them.

kvark (approved):

webgpu.h
Kangz: Sorry, what I meant was "WGPUSType_SurfaceDescriptorFromWaylandSurface" instead of "WGPUSType_SurfaceDescriptorFromWayland". We are making a `WGPUSurface` from a wayland surface, not from wayland itself. It is a very nitpicky comment though.

rajveermalviya: to me, `SurfaceDescriptorFromWaylandSurface` does indeed make more sense, as it aligns better with other SufaceDescriptor structs (`..FromWindowsHWND`, `..FromMetalLayer`, `..FromCanvasHTMLSelector`).

Though `SurfaceDescriptorFromXlib` does seem like an outlier (which is what I looked for current `..FromWayland`).

so anyways, should I change it to `SurfaceDescriptorFromWaylandSurface` ?

Kangz: Yeah I'll make a fix for `SurfaceDescriptorFromXlib` after this. Let's use `SurfaceDescriptorFromWaylandSurface` thanks!

rajveermalviya: updated

PR #146, opened 2022-02-03
emscripten: use uint64_t instead of pointers

caiiiycuk:
We should not use pointers to avoid `signature_mistmatch` errors, because emscripten is 32 bit system, but wgpu returns 64 bit ids.

related to https://github.com/gfx-rs/wgpu-native/issues/169

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -62,6 +62,32 @@
 
 typedef uint32_t WGPUFlags;
 
+#ifdef EMSCRIPTEN
+// We should not use pointers here to avoid signature_mistmatch
+// errors, because emscripten is 32 bit system, but wgpu returns 
+// 64 bit ids.
+typedef uint64_t WGPUAdapter;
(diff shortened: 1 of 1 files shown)


Kangz:
I'm sorry put we're going to need a lot more context here. `webgpu.h` defines the handle types to be pointer-sized, and the header works on both 32bit and 64bit, so what is the reason for special casing emscripten? (especially since it is a 32bit ABI)

caiiiycuk:
> I'm sorry put we're going to need a lot more context here. `webgpu.h` defines the handle types to be pointer-sized, and the header works on both 32bit and 64bit, so what is the reason for special casing emscripten? (especially since it is a 32bit ABI)

I think I misunderstood the purpose of repository. I thought it related to wgpu project. In wgpu-native, implementation of this header returns uint64_t  id on both 32bit and 64 bit systems.

PR #153, opened 2022-03-20
Add WGPUSurfaceDescriptorFromXcbWindow

rajveermalviya:
https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkXcbSurfaceCreateInfoKHR.html

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -359,6 +359,7 @@ typedef enum WGPUSType {
     WGPUSType_PrimitiveDepthClipControl = 0x00000007,
     WGPUSType_SurfaceDescriptorFromWaylandSurface = 0x00000008,
     WGPUSType_SurfaceDescriptorFromAndroidNativeWindow = 0x00000009,
+    WGPUSType_SurfaceDescriptorFromXcbWindow = 0x00000010,
     WGPUSType_Force32 = 0x7FFFFFFF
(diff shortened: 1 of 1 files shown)


Kangz:
LGTM although I don't think it will be possible to use it with OpenGL ES because AFAIK EGL only supports X11 windows and not XCB windows. (and you can't convert XCB -> X11 window). That could be fine.

@kvark or @cwfitzgerald WDYT for wgpu?

rajveermalviya:
as per [this](https://www.phoronix.com/scan.php?page=news_item&px=Mesa-21.0-EGL_EXT_platform_xcb) & [this](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/6474) mesa drivers support it since 21.0, no idea about nvidia though.

cwfitzgerald:
Seems we only support X11 and wayland https://github.com/gfx-rs/wgpu/blob/master/wgpu-hal/src/gles/egl.rs#L563-L569

cwfitzgerald:
Ah yeah, that should be fine, it matches what we get from raw-window-handle https://docs.rs/raw-window-handle/latest/raw_window_handle/struct.XcbHandle.html

Kangz:
Great!

PR #157, opened 2022-05-23
Add nullable annotations

Kangz:
FYI @litherum


austinEng (approved):

litherum:
I was thinking of `__nullable` from https://clang.llvm.org/docs/analyzer/developer-docs/nullability.html ...

Issue #163, opened 2022-07-04
proposal: support for automatic presentation modes


rajveermalviya:
[wgpu](https://github.com/gfx-rs/wgpu) recently made changes to presention api, which removes implicit fallback for unsupported presentation modes.

it now allows three new presentation modes:
 - `FifoRelaxed` 
 - `AutoVsync` : `FifoRelaxed` if available, otherwise fallback to `Fifo`
 - `AutoNoVsync` : `Immediate` or else `Mailbox` or else `Fifo`
 
it also adds a companion method for listing supported presentation modes for a surface.
for webgpu.h it can be:

```c
WGPUPresentMode const * wgpuSurfaceGetSupportedPresentModes(WGPUSurface surface, WGPUAdapter adapter, size_t * count);
```

so can we add these three present modes & this function to webgpu.h ?
 
for more info check:
https://github.com/gfx-rs/wgpu/blob/c36eb9f44bd9462fe176fb859839592131428feb/wgpu-types/src/lib.rs#L2951-L3021

also cc @cwfitzgerald @jimblandy

rajveermalviya:
closing in favor of #197

PR #164, opened 2022-07-17
replace WGPUSwapChain with methods for WGPUSurface

### LINKED ISSUE #88: usability of swapchain in a native context
At Deno, we are currently working on adding swapchain support which would allow to render to a window. Problem is that Deno strives to be browser compatible, but that is difficult as usage of swapchain wasn't designed for a native context.

### LINKED ISSUE #89: Return a texture instead of a view by the swapchain
As a follow-up to #88, we should start aligning this API closer to upstream.
Switching to a texture from a view is one such step.

```cpp
typedef WGPUTextureView (*WGPUProcSwapChainGetCurrentTextureView)(WGPUSwapChain swapChain);
```
Should be:
```cpp
typedef WGPUTexture (*WGPUProcSwapChainGetCurrentTexture)(WGPUSwapChain swapChain);
```

rajveermalviya:
remove WGPUSwapChain as its already been [removed](https://github.com/gpuweb/gpuweb/pull/1810) from webgpu spec.

- replaces `wgpuDeviceCreateSwapChain` with  `wgpuSurfaceConfigure`
- replaces `wgpuSwapChainPresent` with `wgpuSurfacePresent`
- replaces `wgpuSwapChainGetCurrentTextureView` with `wgpuSurfaceGetCurrentTexture`
- adds `wgpuSurfaceUnconfigure`

`wgpu` already supports most of these, except for `wgpuSurfaceUnconfigure` and `WGPUSurfaceConfiguration.viewFormats`

One question:
- should `WGPUSurfaceConfiguration` also have `alphaMode` & `colorSpace` from [`GPUCanvasConfiguration`](https://gpuweb.github.io/gpuweb/#dictdef-gpucanvasconfiguration)?

Fixes #89
Fixes #88


webgpu.h
```diff
@@ -1354,10 +1353,10 @@ typedef void (*WGPUProcShaderModuleSetLabel)(WGPUShaderModule shaderModule, char
 
 // Procs of Surface
 typedef WGPUTextureFormat (*WGPUProcSurfaceGetPreferredFormat)(WGPUSurface surface, WGPUAdapter adapter);
-
-// Procs of SwapChain
-typedef WGPUTextureView (*WGPUProcSwapChainGetCurrentTextureView)(WGPUSwapChain swapChain);
-typedef void (*WGPUProcSwapChainPresent)(WGPUSwapChain swapChain);
+typedef void (*WGPUProcSurfaceConfigure)(WGPUSurface surface, WGPUDevice device, WGPUSurfaceConfiguration const * configuration);
```
Kangz, line 1356: nit: To match the Web API exactly, the Device would be in the WGPUSurfaceConfiguration.

webgpu.h
```diff
@@ -1354,10 +1353,10 @@ typedef void (*WGPUProcShaderModuleSetLabel)(WGPUShaderModule shaderModule, char
...
+typedef void (*WGPUProcSurfaceUnconfigure)(WGPUSurface surface);
+typedef WGPUTexture (*WGPUProcGetCurrentTexture)(WGPUSurface surface, bool * suboptimal);
```
Kangz, line 1358: nit: Maybe it could be a separate call, but also optionally return some optimal parameters like the size of the texture?

webgpu.h
```diff
@@ -1354,10 +1353,10 @@ typedef void (*WGPUProcShaderModuleSetLabel)(WGPUShaderModule shaderModule, char
...
+typedef void (*WGPUProcSurfacePresent)(WGPUSurface surface);
```
Kangz, line 1359: There are other changes @cwfitzgerald made compared to the Web API for getting the supported formats and present mode. Maybe we should replicate some of them, and also re-discuss them in the context of a native API that's shared between projects.

Kangz (commented):
Overall direction to match the Web API sounds good, but I think we need more changes than just this.

webgpu.h
rajveermalviya: #163 mentions `wgpuSurfaceGetSupportedPresentModes`.

for supported formats I think we can have
 ```c
size_t wgpuSurfaceGetSupportedFormats(WGPUSurface surface, WGPUTextureFormats * formats);
```

File: webgpu.h (hunk 2 above)
Comment (rajveermalviya) [reply webgpu.h:1358]: wouldn't separate calls to `wgpuTextureGetWidth` & `wgpuTextureGetHeight` on the returned `WGPUTexture` suffice?

Comment (Kangz) [reply webgpu.h:1358]: I don't think so. The texture you get should have exactly the size defined in `WGPUSurfaceConfiguration` otherwise the application doesn't really control when the resize happens and needs to be ready to resize other rendertargets any time. (plus if they forget to check then they'll get errors leading to rendering a black screen). At least in Dawn we have manual test that the resizing work, and it is implemented with a blit in the backends if it is needed (of course optimal would return false ^^)

File: webgpu.h (hunk 3 above)
Comment (rajveermalviya) [reply webgpu.h:1359]: don't think `wgpu` [provides a way to know that](https://docs.rs/wgpu-core/latest/wgpu_core/hub/struct.Global.html#method.surface_get_supported_formats)

(also we don't have hdr support in `webgpu.h`, yet, [wgpu does though](https://docs.rs/wgpu-types/0.13.2/wgpu_types/enum.AstcChannel.html#variant.Hdr))

Comment (Kangz) [reply webgpu.h:1359]: `AstcChannel` seems unrelated to the swapchain API TBH.

Comment (rajveermalviya) [reply webgpu.h:1359]: Ah yes, apologies!

File: webgpu.h (hunk 1 above)
Comment (rajveermalviya) [reply webgpu.h]: ack

File: webgpu.h (hunk 2 above)
Comment (rajveermalviya) [reply webgpu.h]: ack

File: webgpu.h [hunk 4]
```diff
@@ -575,6 +574,15 @@ typedef enum WGPUVertexStepMode {
     WGPUVertexStepMode_Force32 = 0x7FFFFFFF
 } WGPUVertexStepMode;
 
+typedef enum WGPUSurfaceStatus {
```
rajveermalviya, line 577: Replaced `bool * suboptimal` with an enum.
This matches [`SurfaceStatus`](https://docs.rs/wgpu-types/latest/wgpu_types/enum.SurfaceStatus.html) in wgpu.

But still not sure if we should only have `Good` & `Suboptimal` and report others as errors.

webgpu.h
```diff
@@ -1354,10 +1363,10 @@ typedef void (*WGPUProcShaderModuleSetLabel)(WGPUShaderModule shaderModule, char
 
 // Procs of Surface
 typedef WGPUTextureFormat (*WGPUProcSurfaceGetPreferredFormat)(WGPUSurface surface, WGPUAdapter adapter);
-
-// Procs of SwapChain
-typedef WGPUTextureView (*WGPUProcSwapChainGetCurrentTextureView)(WGPUSwapChain swapChain);
-typedef void (*WGPUProcSwapChainPresent)(WGPUSwapChain swapChain);
+typedef void (*WGPUProcSurfaceConfigure)(WGPUSurface surface, WGPUSurfaceConfiguration const * configuration);
+typedef void (*WGPUProcSurfaceUnconfigure)(WGPUSurface surface);
+typedef WGPUTexture (*WGPUProcSurfaceGetCurrentTexture)(WGPUSurface surface, uint32_t * width, uint32_t * height, WGPUSurfaceStatus * status);
```
Kangz, line 1368: I'm not sure why we need to return all of these by pointer. `width` and `height` is whatever the user configured and is also available with `wgpuTextureGetWidth/Height`. Status could just be `wgpuSurfaceGetStatus`?

@cwfitzgerald wdyt?

webgpu.h
Kangz: I'm not sure what Outdated means in practice, but the enum sounds good.

rajveermalviya: >I'm not sure what Outdated means in practice

in wgpu it is [mapped](https://github.com/gfx-rs/wgpu/blob/f7526ae9604a91d14e58eea0d6342988b4a94e79/wgpu-hal/src/vulkan/instance.rs#L744) to `VK_ERROR_OUT_OF_DATE_KHR`

after thinking more about it, WGPUSurfaceStatus can be
```c
typedef enum WGPUSurfaceStatus {
    WGPUSurfaceStatus_Good = 0x00000000,  // ~ VK_SUCCESS
    WGPUSurfaceStatus_Suboptimal = 0x00000001, // ~ VK_SUBOPTIMAL_KHR
    WGPUSurfaceStatus_Timeout = 0x00000002, // ~ VK_TIMEOUT
    WGPUSurfaceStatus_Force32 = 0x7FFFFFFF
} WGPUSurfaceStatus;
```
removed Lost & Outdated, those can be reported as errors.

webgpu.h
rajveermalviya: for `width` & `height` if I didn't misunderstand, didn't you propose to return them [here](https://github.com/webgpu-native/webgpu-headers/pull/164#discussion_r940317239).

for `status` I think we should return it via pointer to enforce users to explicitly check swapchain status

webgpu.h
cwfitzgerald: Let's tackle this in another PR - maybe after a mini issue-rfc kind of deal - as there is a ton of complexity there that we aren't yet equipped to deal with here. For reference: https://github.com/gfx-rs/wgpu/issues/2869 is our proposal internal to wgpu which would set the stage for the HDR api.

webgpu.h
cwfitzgerald: We return this immediately - I would also like to discorage a glGetError kinda interface as much as we can.

Kangz: I assume that by "this" you mean the status? That'd work, although maybe we should support passing `NULL` if you don't care. The pointer is slightly unfortunate but IDK how to make a C API better then this (Maybe the function returns the status and it is the WGPUTexture that's returned by pointer? Not sure).

lisanlow:
@Kangz @rajveermalviya I am looking forward to the changes in this PR, may I know if there is currently a road block to its completion and is there any plan to get this change into the first version of webgpu release?

mwyrzykowski:
@Kangz it looks like the last commit to swapchain-rework was 6 months ago. Do you know if these changes are still planned for v1?

It would align the web API more directly with WebGPU.h. Currently there is a bit of a disconnect between them with Surface and SwapChain concepts in WebGPU.h but they are bundled together in GPUCanvasContext in the web specification.

austinEng:
@mwyrzykowski, our team is currently more focused on getting V1 of the JavaScript API out. We may not get to reworking the native swapchain API until later.

litherum:
this is maybe a silly question, but is it really true that updating the webgpu-native/webgpu-headers is blocked on Dawn's implementation?

Can't this repo be updated in parallel?

austinEng:
It certainly can - I was simply saying that our team is probably not going to do native swapchain changes just yet.

PR #181, opened 2023-05-16
Assorted updates:

Kangz:
- Remove the return value from PopErrorScope (validation errors now cause promise rejection and not JS errors).
 - Add more mapping failure reasons to help browsers produce nice validation error messages.
 - Remove the unused WGPUTextureComponentType
 - Tag the buffer in SetVertexBuffer as nullable (to unset it).


Kangz:
PTAL!

austinEng (approved):

Issue #179, opened 2023-05-13
No way to refine types/enums/functions for Swift


litherum:
Frameworks can expose C symbols (of course). However, when projected into Swift, those C symbols are usually not very ergonomic. One example is asynchronous functions: In C, they take a function pointer and a `void* userdata`. If the same function was to be exposed in Swift, it should be set up in such a way that it can actually be marked as `async` and use Swift's built in facilities for asynchronous programming.

To handle this, frameworks have a [mechanism for saying](https://developer.apple.com/documentation/swift/improving-objective-c-api-declarations-for-swift) "expose this symbol to C clients, but not to Swift clients" so that the functionality can be wrapped and exposed in a more ergonomic Swift wrapper. That mechanism looks like this:
```
typedef struct WGPUFooImpl* WGPUFoo NS_REFINED_FOR_SWIFT;
...
typedef enum WGPUBar {
    WGPUBar_Undefined = 0x00000000
} WGPUBar NS_REFINED_FOR_SWIFT;
...
WGPU_EXPORT void wgpuFooBaz() NS_REFINED_FOR_SWIFT;
```

However, the declarations in WebGPU.h are just:

```
typedef struct WGPUFooImpl* WGPUFoo;
...
typedef enum WGPUBar {
    WGPUBar_Undefined = 0x00000000
} WGPUBar;
...
WGPU_EXPORT void wgpuFooBaz();
```

There is no way for us to stick the `NS_REFINED_FOR_SWIFT` attribute on these functions/types.

It would be useful if WebGPU.h was modified, to do something like:
```
#if !defined(WGPU_TYPE_ATTRIBUTE)
#define WGPU_TYPE_ATTRIBUTE
#endif

#if !defined(WGPU_ENUM_ATTRIBUTE)
#define WGPU_ENUM_ATTRIBUTE
#endif

#if !defined(WGPU_FUNCTION_ATTRIBUTE)
#define WGPU_FUNCTION_ATTRIBUTE
#endif
...
typedef struct WGPUFooImpl* WGPUFoo WGPU_TYPE_ATTRIBUTE;
...
typedef enum WGPUBar {
    WGPUBar_Undefined = 0x00000000
} WGPUBar WGPU_ENUM_ATTRIBUTE;
...
WGPU_EXPORT void wgpuFooBaz() WGPU_FUNCTION_ATTRIBUTE;
```

[...]

PR #182, opened 2023-05-16
Add overridable attributes to all types (defaulting to empty) 

Kangz:
@litherum PTAL at the second commit, would this work for you?

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -48,6 +48,22 @@
 #    define WGPU_EXPORT
 #endif  // defined(WGPU_SHARED_LIBRARY)
 
+#if !defined(WGPU_OBJECT_ATTRIBUTE)
+#define WGPU_OBJECT_ATTRIBUTE
+#endif
+#if !defined(WGPU_ENUM_ATTRIBUTE)
+#define WGPU_ENUM_ATTRIBUTE
+#endif
+#if !defined(WGPU_STRUCTURE_ATTRIBUTE)
+#define WGPU_STRUCTURE_ATTRIBUTE
+#endif
+#if !defined(WGPU_FUNCTION_ATTRIBUTE)
+#define WGPU_FUNCTION_ATTRIBUTE
+#endif
+#if !defined(WGPU_NULLABLE)
+#define WGPU_NULLABLE
+#endif
+
 #include <stdint.h>
 #include <stddef.h>
 #include <stdbool.h>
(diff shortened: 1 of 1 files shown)


austinEng:
while you're at it; you may consider adding a WGPU_NULLABLE attribute too?

Kangz:
I'll add `WGPU_NULLABLE` if we decide to move forward after the discussions in #179

litherum:
+1 to nullable <3

Yes, this patch looks perfect.

litherum (approved):

Kangz:
PTAL again, I had to bundle a couple changes that would have been annoying to split up.

webgpu.h
```diff
@@ -723,29 +818,29 @@ typedef struct WGPUCompilationMessage {
...
 typedef struct WGPUComputePassTimestampWrite {
     WGPUQuerySet querySet;
     uint32_t queryIndex;
     WGPUComputePassTimestampLocation location;
-} WGPUComputePassTimestampWrite;
+} WGPUComputePassTimestampWrite WGPU_STRUCTURE_ATTRIBUTE;
 
 typedef struct WGPUConstantEntry {
     WGPUChainedStruct const * nextInChain;
     char const * key;
```
kainino0x, line 831: we might also want a NONNULL for all of the pointers that aren't nullable?

webgpu.h
```diff
@@ -48,6 +48,22 @@
...
+#define WGPU_ENUM_ATTRIBUTE
+#endif
+#if !defined(WGPU_STRUCTURE_ATTRIBUTE)
+#define WGPU_STRUCTURE_ATTRIBUTE
+#endif
+#if !defined(WGPU_FUNCTION_ATTRIBUTE)
+#define WGPU_FUNCTION_ATTRIBUTE
+#endif
+#if !defined(WGPU_NULLABLE)
+#define WGPU_NULLABLE
```
kainino0x, line 64: Do we want the header to use the clang extension (and maybe other extensions) by default if available, or just leave this for people who want to take advantage of it?

webgpu.h
```diff
@@ -639,18 +723,29 @@ typedef enum WGPUTextureUsage {
...
+typedef void (*WGPUCreateRenderPipelineAsyncCallback)(WGPUCreatePipelineAsyncStatus status, WGPURenderPipeline pipeline, char const * message, void * userdata) WGPU_FUNCTION_ATTRIBUTE;
+typedef void (*WGPUDeviceLostCallback)(WGPUDeviceLostReason reason, char const * message, void * userdata) WGPU_FUNCTION_ATTRIBUTE;
+typedef void (*WGPUErrorCallback)(WGPUErrorType type, char const * message, void * userdata) WGPU_FUNCTION_ATTRIBUTE;
+typedef void (*WGPUProc)(void) WGPU_FUNCTION_ATTRIBUTE;
+typedef void (*WGPUQueueWorkDoneCallback)(WGPUQueueWorkDoneStatus status, void * userdata) WGPU_FUNCTION_ATTRIBUTE;
+typedef void (*WGPURequestAdapterCallback)(WGPURequestAdapterStatus status, WGPUAdapter adapter, char const * message, void * userdata) WGPU_FUNCTION_ATTRIBUTE;
+typedef void (*WGPURequestDeviceCallback)(WGPURequestDeviceStatus status, WGPUDevice device, char const * message, void * userdata) WGPU_FUNCTION_ATTRIBUTE;
 
 typedef struct WGPUChainedStruct {
     struct WGPUChainedStruct const * next;
```
kainino0x, line 741: These should be nullable

kainino0x (approved):
LGTM w/ non-blocking comments

webgpu.h
```diff
@@ -149,11 +149,14 @@ typedef enum WGPUBufferBindingType {
...
     WGPUBufferMapAsyncStatus_Success = 0x00000000,
-    WGPUBufferMapAsyncStatus_Error = 0x00000001,
+    WGPUBufferMapAsyncStatus_ValidationError = 0x00000001,
     WGPUBufferMapAsyncStatus_Unknown = 0x00000002,
     WGPUBufferMapAsyncStatus_DeviceLost = 0x00000003,
     WGPUBufferMapAsyncStatus_DestroyedBeforeCallback = 0x00000004,
     WGPUBufferMapAsyncStatus_UnmappedBeforeCallback = 0x00000005,
+    WGPUBufferMapAsyncStatus_MappingAlreadyPending = 0x00000006,
+    WGPUBufferMapAsyncStatus_OffsetOutOfRange = 0x00000007,
+    WGPUBufferMapAsyncStatus_SizeOutOfRange = 0x00000008,
```
kainino0x, line 247: We did this for Dawn but Austin(?) has pointed out that the mapAsync callback should maybe just have a `const char * message` instead, so not sure we should land it here yet

Kangz: This has already landed in https://github.com/webgpu-native/webgpu-headers/pull/181 but we could discuss further and change it!

webgpu.h
Kangz: That's would be a lot of noise, but maybe? What do others think of adding a `WGPU_NONNULL`

kainino0x: @litherum ?

webgpu.h
kainino0x: SG, we'll worry about it later.

kainino0x:
Rebased past #181, let's do the rest as followup

Issue #188, opened 2023-06-03
ExternalTexture support


litherum:
There's not really any reason why external textures can't exist in webgpu-headers. They have platform-specific parts, but then again so does the `WGPUSurface`. I expect a similar design as `WGPUSurface` can be used to describe the external texture descriptors.

Kangz:
We could try to revive https://github.com/webgpu-native/webgpu-headers/pull/192 or simila but for now we are trying to stabilize a v1 of the header so support for external texture would silghtly get in the way. IMHO we can look at it later. @ygdrasil-io what are you looking to use them for in native?

ygdrasil-io:
> We could try to revive [#192](https://github.com/webgpu-native/webgpu-headers/pull/192) or simila but for now we are trying to stabilize a v1 of the header so support for external texture would silghtly get in the way. IMHO we can look at it later. [@ygdrasil-io](https://github.com/ygdrasil-io) what are you looking to use them for in native?

I am working on a [Kotlin binding for wgpu](https://github.com/wgpu4k/wgpu4k), and I would like to move the external texture management code from my library to wgpu-native. However, to avoid going in any random direction, I would prefer if there were already a consensus on the native specifications.

PR #192, opened 2023-06-05
External texture

Kangz:
FYI @litherum @cwfitzgerald

The external texture descriptor is pretty Dawn-specific but the rest seems like it could be shared easily. WDYT?


litherum:
- Should probably say that this fixes https://github.com/webgpu-native/webgpu-headers/issues/188
- I think most of `WGPUExternalTextureDescriptor` is Dawn-specific, so it probably should be moved into a chained struct, rather than being in the top-level `WGPUExternalTextureDescriptor` struct. For comparison, our [implementation](https://github.com/WebKit/WebKit/blob/main/Source/WebGPU/WebGPU/WebGPUExt.h#L93) of `WGPUExternalTextureDescriptor` is much simpler. Also, for us, we can use the platform-specific `CVPixelBufferRef` which includes a lot of the metadata within itself.

Issue #197, opened 2023-07-18
proposal: swapchain & presentation rework


rajveermalviya:
# Context

Currently the WGPUSwapChain API diverts from the js spec; this proposal tries to match the js spec while also maintaining the functionality that is required in the native context.

Keep in mind this proposal is mostly what `wgpu-rs` already exposes so it tries to match it.

### Update present mode enum to add automatic modes

```c
// matches wgpu
// https://docs.rs/wgpu-types/latest/wgpu_types/enum.PresentMode.html
typedef enum WGPUPresentMode {
	WGPUPresentMode_Fifo = 0x00000000, // default
	// VK_PRESENT_MODE_FIFO_RELAXED_KHR
	WGPUPresentMode_FifoRelaxed = 0x00000001, // new
	WGPUPresentMode_Immediate = 0x00000002,
	WGPUPresentMode_Mailbox = 0x00000003,
	WGPUPresentMode_Force32 = 0x7FFFFFFF
} WGPUPresentMode;
```

### Remove `WGPUSwapChain` (and merge it's interface into Surface api)

In the js spec this was done [a while ago](https://github.com/gpuweb/gpuweb/pull/1810).
Fixes: https://github.com/webgpu-native/webgpu-headers/issues/88

```c
// matches wgpu
// https://docs.rs/wgpu-types/latest/wgpu_types/enum.CompositeAlphaMode.html
typedef enum WGPUCompositeAlphaMode {
	// Chooses between Opaque & Inherit whichever is available.
	WGPUCompositeAlphaMode_Auto = 0x00000000,
	WGPUCompositeAlphaMode_Opaque = 0x00000001,
	WGPUCompositeAlphaMode_PreMultiplied = 0x00000002,
	WGPUCompositeAlphaMode_UnPreMultiplied = 0x00000003,
	WGPUCompositeAlphaMode_Inherit = 0x00000004,
	WGPUCompositeAlphaMode_Force32 = 0x7FFFFFFF
} WGPUCompositeAlphaMode;

// Replaces WGPUSwapChainDescriptor
typedef struct WGPUSurfaceConfiguration {
	WGPUChainedStruct const * nextInChain;
	WGPUDevice device;
	WGPUTextureFormat format;
	WGPUTextureUsage usage;
	size_t viewFormatCount;
	WGPUTextureFormat const * viewFormats;
	WGPUCompositeAlphaMode alphaMode;

	uint32_t width;
	uint32_t height;
[...]

PR #203, opened 2023-07-29
swapchain & presentation rework

rajveermalviya:
closes: #197

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -98,7 +98,6 @@ typedef struct WGPURenderPipelineImpl* WGPURenderPipeline WGPU_OBJECT_ATTRIBUTE;
 typedef struct WGPUSamplerImpl* WGPUSampler WGPU_OBJECT_ATTRIBUTE;
 typedef struct WGPUShaderModuleImpl* WGPUShaderModule WGPU_OBJECT_ATTRIBUTE;
 typedef struct WGPUSurfaceImpl* WGPUSurface WGPU_OBJECT_ATTRIBUTE;
-typedef struct WGPUSwapChainImpl* WGPUSwapChain WGPU_OBJECT_ATTRIBUTE;
 typedef struct WGPUTextureImpl* WGPUTexture WGPU_OBJECT_ATTRIBUTE;
 typedef struct WGPUTextureViewImpl* WGPUTextureView WGPU_OBJECT_ATTRIBUTE;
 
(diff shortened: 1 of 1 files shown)


Kangz (approved):
LGTM though I'd like @kainino0x to look at it as well.

It's going to be a pain to update Dawn ^^

webgpu.h
```diff
@@ -496,6 +507,16 @@ typedef enum WGPUStoreOp {
     WGPUStoreOp_Force32 = 0x7FFFFFFF
 } WGPUStoreOp WGPU_ENUM_ATTRIBUTE;
 
+typedef enum WGPUSurfaceGetCurrentTextureError {
+	WGPUSurfaceGetCurrentTextureError_NoError = 0x00000000,
```
kainino0x, line 511: We currently have 1 other thing called "NoError" and 6 called "Success". Maybe we should standardize on "Success"? And "Status" instead of "Error"?

(Actually, I don't even know what the other NoError could possibly be used for...)

webgpu.h
```diff
@@ -284,6 +285,15 @@ typedef enum WGPUCompilationMessageType {
     WGPUCompilationMessageType_Force32 = 0x7FFFFFFF
 } WGPUCompilationMessageType WGPU_ENUM_ATTRIBUTE;
 
+typedef enum WGPUCompositeAlphaMode {
+	WGPUCompositeAlphaMode_Auto = 0x00000000,
```
kainino0x, line 289: nit: replace all tabs with 4 spaces

kainino0x (approved):
LGTM but I did just think of one more thing to bikeshed

webgpu.h
rajveermalviya: ack

webgpu.h
kainino0x: > (Actually, I don't even know what the other NoError could possibly be used for...)

Never mind, it's used by PopErrorScope when there's no error in the scope. That name makes sense.

rajveermalviya: > Maybe we should standardize on "Success"

:+1: 

>  "Status" instead of "Error"?

If we go for `Status` instead of `Error` then imo we should remove DeviceLost from here, apps should be able to catch it via DeviceLostCallback anyway.

kainino0x: That's a good point, I wonder if we should also remove it from WGPUErrorType (PopErrorScope) and maybe the others? In line with https://github.com/gpuweb/gpuweb/issues/1629

kainino0x: Meeting: Decided to change to `Status` and `Status_Error`, but keep DeviceLost for now. New issue about DeviceLost: #209

rajveermalviya: ack

kainino0x (approved):

kainino0x:
Going to go ahead and land this; whoever ends up doing the next Dawn->webgpu-headers update (probably @lokokung) you'll probably want to put this into our generator (tagged "upstream") to preserve these. It'll move them around, but that's fine.

Issue #84, opened 2021-04-01
Issues with C / C++ bool type.


Kangz:
This is an issue that someone (graphite?) raised on Discord but didn't file here. The `bool` types in C/C++ are kinda weird and can not match, especially when put back to back. `webgpu.h` shouldn't rely on this type, and instead do something like Vulkan's `VkBool32`.

Beyley:
is there an ETA for a fix landing in this repository? movement seems to be happening chromium side, but i dont see an open PR over here, should i open a PR for adding a typedef for WGPUBool?

lokokung:
Yea this should be coming within the next week. I am trying to land the effective changes on Chromium side to make sure it will work there. I can create the corresponding PR here once that's progressed a bit more.

PR #210, opened 2023-08-04
Adds WGPUBool typedef.

lokokung:
#84

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -66,7 +66,6 @@
 
 #include <stdint.h>
 #include <stddef.h>
-#include <stdbool.h>
 
 #define WGPU_ARRAY_LAYER_COUNT_UNDEFINED (0xffffffffUL)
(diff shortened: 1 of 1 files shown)


kainino0x (commented):
Any reason not to bundle with the other changes to remove stdbool.h and usages of `bool`?

lokokung:
Whoops, no, there's no reason it wasn't included. Updated now!

Beyley (approved):
looks good from my end! should help smooth out bindings from C# (since `bool` in C# is generally marshalled as a 32-bit type for compat with the windows BOOL type, see [this page](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.unmanagedtype?view=net-7.0) UnmanagedType.Bool is 32-bit)

kainino0x (approved):
LGTM!

PR #218, opened 2023-08-21
Use WGPUTextureUsageFlags in WGPUSurfaceConfiguration

dj2:
The TextureUsage is a Bitmask, update the usage in SurfaceConfiguration to use the Flags typedef to be consistent with out uses of TextureUsage.

Follow-up to #203


kainino0x (approved):
LGTM, fyi @rajveermalviya in case you run into this while implementing

PR #219, opened 2023-08-21
Make WGPURenderPassColorAttachment extendable.

dj2:
Update WGPURenderPassColorAttachment as extendable to match the current usage in Dawn which extends the structure.


kainino0x (approved):
LGTM, there's no other way to extend this so I'm just going to call this good and land it.

PR #221, opened 2023-08-21
Add depthSlice to WGPURenderPassColorAttachment

### LINKED ISSUE #215: Adding depthSlice to WGPURenderPassColorAttachment
This is getting added to core in the JS API.

Should we directly add it in these headers (no chained struct) given that we're not at the point of ABI stability yet?

dj2:
Fixes #215

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -1234,6 +1234,7 @@ typedef struct WGPUProgrammableStageDescriptor {
 typedef struct WGPURenderPassColorAttachment {
     WGPUChainedStruct const * nextInChain;
     WGPU_NULLABLE WGPUTextureView view;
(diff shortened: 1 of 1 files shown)


kainino0x (commented):
I'm not sure if we should land this yet. It's not 100% clear from https://github.com/gpuweb/gpuweb/issues/4251 whether the API shape will end up changing before it gets landed (for example we might have a baseDepthSlice and depthSliceCount or something, not sure).

Maybe we should just tag this one as "dawn" in dawn.json for now and let it bake a bit longer.

dj2:
Sounds good, will close this and update dawn.json instead. Thanks.

PR #222, opened 2023-08-21
Add new WGPULimit value

dj2:
Add the maxBindGroupsPlusVertexBuffers entry to WGPULimits.


kainino0x (approved):
LGTM, it's in the JS API so it should be here.

PR #236, opened 2023-10-19
s/PipelineStatisticsQuery/ChromiumExperimentalPipelineStatisticsQuery/g

beaufortfrancois:
Following https://dawn-review.googlesource.com/c/dawn/+/156823, this PR renames the feature name `WGPUFeatureName_PipelineStatisticsQuery` to `WGPUFeatureName_ChromiumExperimentalPipelineStatisticsQuery`

@kainino0x Please review.

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -340,7 +340,7 @@ typedef enum WGPUFeatureName {
     WGPUFeatureName_DepthClipControl = 0x00000001,
     WGPUFeatureName_Depth32FloatStencil8 = 0x00000002,
     WGPUFeatureName_TimestampQuery = 0x00000003,
-    WGPUFeatureName_PipelineStatisticsQuery = 0x00000004,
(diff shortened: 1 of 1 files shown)


beaufortfrancois:
@rajveermalviya This feature is renamed so that developers who use this feature are aware it is still experimental and that it may break in the future.

beaufortfrancois:
That would work for me. @kainino0x what do you think?

kainino0x:
If we're going to mark this experimental, then I agree it shouldn't be in this header. We're trying to stabilize this header soonish, which means we won't be able to remove enum variants.

As for whether it should be experimental - from the standpoint of this header, yes, it probably should be, in case it changes.

beaufortfrancois:
I've just uploaded https://github.com/webgpu-native/webgpu-headers/pull/238 to address your feedback

beaufortfrancois:
I'm closing this PR as I expect https://github.com/webgpu-native/webgpu-headers/pull/238 to be merged soon.

PR #245, opened 2023-11-09
Xlib's Window changed to u64, conforming to X.h

radgeRayden:
As per https://github.com/gfx-rs/wgpu-native/issues/304. The type definition for the X Window handles can be found at https://gitlab.freedesktop.org/xorg/proto/xorgproto/-/blob/master/include/X11/X.h?ref_type=heads#L96. `Window` is of type `XID` = `u64`.

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -1096,7 +1096,7 @@ typedef struct WGPUSurfaceDescriptorFromXcbWindow {
 typedef struct WGPUSurfaceDescriptorFromXlibWindow {
     WGPUChainedStruct chain;
     void * display;
-    uint32_t window;
+    uint64_t window;
 } WGPUSurfaceDescriptorFromXlibWindow WGPU_STRUCTURE_ATTRIBUTE;
 
 typedef struct WGPUSurfaceTexture {


kainino0x:
I found that this changed in this PR, and also was reported as a problem there:
https://github.com/webgpu-native/webgpu-headers/pull/30#issuecomment-574833262

> `Window` is of type `XID` = `u64`.

AFAICT it _can_ also be u32:

```c
#ifndef _XSERVER64
#  ifndef _XTYPEDEF_XID
#    define _XTYPEDEF_XID
typedef unsigned long XID;
// ...
#else
#  include <X11/Xmd.h>
#  ifndef _XTYPEDEF_XID
#    define _XTYPEDEF_XID
typedef CARD32 XID;
```
where I'm pretty sure CARD32 is always `uint32_t`:
```c
# ifdef LONG64
typedef unsigned long CARD64;
typedef unsigned int CARD32;
# else
typedef unsigned long long CARD64;
typedef unsigned long CARD32;
# endif
```

Anyway, this is a mess. I don't think we can match the type, because it requires knowing whether X was built with _XSERVER64 enabled. 

The conclusion is the same, though - we need to make this a `uint64_t`, I think.

kainino0x (approved):

Kangz (approved):

PR #256, opened 2024-01-03
Add const qualifier to SurfaceCapabilities lists

eliemichel:
The end user of GetCapabilities should not be allowed to mutate capabilities as far as I understand it.

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -1030,11 +1030,11 @@ typedef struct WGPUStorageTextureBindingLayout {
 typedef struct WGPUSurfaceCapabilities {
     WGPUChainedStructOut * nextInChain;
     size_t formatCount;
-    WGPUTextureFormat * formats;
+    WGPUTextureFormat const * formats;
     size_t presentModeCount;
-    WGPUPresentMode * presentModes;
+    WGPUPresentMode const * presentModes;
     size_t alphaModeCount;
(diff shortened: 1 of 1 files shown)


Kangz:
I'm not sure: the structure returned and its memory is owned by the user of the API so they can mutate it if they'd like.

eliemichel:
True, though imho non-const pointers somehow suggest that the user could mutate the actual capabilities by mutating the fields of the `WGPUSurfaceCapabilities` struct.

And I can hardly imagine a use case where one would need to mutate these... but on the other hand why not if we do not need to forbid it.

kainino0x:
Jan 4 meeting:

*   KN: Wondering if these could ever actually be owned by some other object and not allocated just for the WGPUSurfaceCapabilities object?
*   CF: Could have refcounting underneath, then it could be shared owned.
*   CF: Is const\_cast undefined behavior in C++?
    *   KN: In this case I’m pretty certain it would be.
    *   (So it would be OK to return a const pointer to internal data.)
*   …
*   KN: Think they should be const in case of the refcounting situation or something.
    *   CF: Actually doing this requires some tricks so FreeMembers can get at the refcount from just SurfaceCapabilities struct. But possible.
*   **Let’s make them const**

kainino0x (approved):

Issue #207, opened 2023-08-03
Define the API in a structured format like vk.xml or dawn.json


kainino0x:
There are many useful properties of the API that aren't expressible in C, but are still programmatically useful, especially for generating bindings to other languages. Vulkan has `vk.xml`, for example. It would make sense for us to have such a thing too.

Dawn already generates `webgpu.h` and `webgpu_cpp.h` from `dawn.json`. If we do define a file like this here, we should generate `webgpu.h` from it too.

See previous discussion on #169, where knowing "in" pointers from "out" pointers would be useful for C# bindings.

kainino0x:
*   KN: How important to do this now?
*   CF: We (wgpu) don’t need it
*   RM: Language bindings will find it very useful
*   CF: Have heard vk.xml is very hard to parse, from the person who maintains the Vulkan bindings for Rust
*   Write it in another language that bakes down to JSON that automated tools can use?
    *   json5, yaml, javascript, typescript?

kainino0x:
I guess the takeaway is: this would be really nice, we should really do it someday, but it's hard to ever call it urgent. So I don't know when we should work on it actively.

I actually quite like the idea of writing this as a typescript program that outputs JSON. It would provide a lightweight schema though wouldn't be a formal schema for the output JSON, and I don't know how important that will be to tooling (guessing not critical? but there are clever things like https://www.npmjs.com/package/typescript-json-schema).

> For a file like this, would it be exclusively webgpu, or would it also have the ability to contain platform specific information (similar to how dawn.json has the `dawn` and `native` flags)?

IMO we should maintain a file that doesn't have that, but it should be powerful enough to also define implementation-specific extensions using the same format.

dj2:
Ok, I'll drop the branch I started (https://github.com/dj2/webgpu-headers/blob/gen/gen_webgpu.rb) where I did it in Ruby as there doesn't seem to be much desire on that front.

The `vk.xml` file is hard to parse, but a chunk of that is due to the complexities of Vulkan and capabilities and extensions. XML itself is pretty easy to parse and procecess with XPath.

eliemichel:
FTR, I wish we had something like this when I wrote my shallow C++ wrapper: https://github.com/eliemichel/WebGPU-Cpp

[...]

PR #259, opened 2024-01-04
generate header from yaml spec

rajveermalviya:
Add a custom yaml spec as discussed in #207, the details about the format can be found in the `schema.json`.

To test the generator:
1. First make sure you have Go installed.
1. Make some change in `webgpu.yml`
1. Run `make gen`

The generator can also be used to generate a separate implementation specific C header along side `webgpu.h`. The steps to do that is:
1. Create a new yaml file, for example `wgpu.yml`
1. Populate it with minimal required fields, refer to `schema.json` and [this example](https://gist.github.com/rajveermalviya/15200babba779bfc21d837af504a1860).
1. Run `go run ./gen -schema schema.json -yaml wgpu.yml -header wgpu.h`

Another option is to generate a single header from multiple yaml specs, which is also included but disabled currently.
I prefer separate headers because that's what `wgpu-native` uses currently.


gen/cheader.tmpl
```diff
@@ -0,0 +1,234 @@
...
+{{- MComment .Doc 0}}
+typedef enum WGPU{{.Name | PascalCase}} {
+    {{- range $entryIndex, $_ := .Entries}}
+        {{- MComment .Doc 4}}
+        {{- $value := "" -}}
+        {{- $valueCombination := .ValueCombination -}}
+        {{- range $valueIndex, $v := .ValueCombination -}}
+            {{- $v = (printf "WGPU%s_%s" ($entry.Name | PascalCase) ($v | PascalCase)) -}}
+            {{- if IsLast $valueIndex $valueCombination -}}
+                {{- $value = (print $value $v) -}}
```
ben-clayton, line 101: nit: parentheses are unnecessary here and below.

gen/cheader.tmpl
```diff
@@ -0,0 +1,234 @@
...
+    WGPU{{.Name | PascalCase}}_Force32 = 0x7FFFFFFF
+} WGPU{{.Name | PascalCase}} WGPU_ENUM_ATTRIBUTE;
+typedef WGPUFlags WGPU{{.Name | PascalCase}}Flags WGPU_ENUM_ATTRIBUTE;{{"\n"}}
+{{- end}}
+
+{{- if eq .Name "webgpu"}}
+typedef void (*WGPUProc)(void) WGPU_FUNCTION_ATTRIBUTE;{{"" -}}
+{{- end}}
+
+{{- if .FunctionTypes}}
```
ben-clayton, line 123: Same thing about scoped indentation, here and below

gen/cheader.tmpl
```diff
@@ -0,0 +1,234 @@
...
+{{- range .FunctionTypes}}
+{{- MComment .Doc 0}}
+typedef {{FunctionReturns .}} (*WGPU{{.Name | PascalCase}})({{FunctionArgs . nil}}) WGPU_FUNCTION_ATTRIBUTE;
+{{- end}}{{"" -}}
+{{- end}}
+
+{{- if .Objects}}
+{{- range $_, $object := .Objects}}
```
ben-clayton, line 131:
```suggestion
{{- range $object := .Objects}}
```

here and below. If you accept this, you can fix all the occurrences with `{{- range $_,` -> `{{- range `

gen/cheader.tmpl
```diff
@@ -0,0 +1,234 @@
...
+
+{{- if eq .Name "webgpu"}}
+typedef void (*WGPUProc)(void) WGPU_FUNCTION_ATTRIBUTE;{{"" -}}
+{{- end}}
+
+{{- if .FunctionTypes}}
+{{- range .FunctionTypes}}
+{{- MComment .Doc 0}}
+typedef {{FunctionReturns .}} (*WGPU{{.Name | PascalCase}})({{FunctionArgs . nil}}) WGPU_FUNCTION_ATTRIBUTE;
+{{- end}}{{"" -}}
```
ben-clayton, line 127: What is the `{{"" -}}` for? You have this in a few places.

gen/gen.go
```diff
@@ -0,0 +1,218 @@
...
+type Data struct {
+	Name string
+	*Yml
+}
+
+func GenCHeader(data *Data, dst io.Writer) error {
+	t := template.
+		New("").
+		Funcs(template.FuncMap{
+			"SComment":     func(v string, indent int) string { return Comment(v, CommentTypeSingleLine, indent, true) },
```
ben-clayton, line 22: An alternative to maintaining absolute indentations in the template is to have two functions to increment and decrement indentation. I've also used special characters in the template to increment and decrement indentation which is then processed at the end of the template.

Not asking for a change here, just suggesting alternatives for you to consider.

gen/gen.go
```diff
@@ -0,0 +1,218 @@
...
+			"MComment":     func(v string, indent int) string { return Comment(v, CommentTypeMultiLine, indent, true) },
+			"SCommentN":    func(v string, indent int) string { return Comment(v, CommentTypeSingleLine, indent, false) },
+			"MCommentN":    func(v string, indent int) string { return Comment(v, CommentTypeMultiLine, indent, false) },
+			"ConstantCase": ConstantCase,
+			"PascalCase":   PascalCase,
+			"CamelCase":    CamelCase,
+			"CType":        CType,
+			"CValue":       CValue,
+			"BitFlagValue": func(index int) uint64 { return uint64(math.Pow(2, float64(index-1))) },
+			"ParseUint": func(s string, bitSize int) uint64 {
```
ben-clayton, line 32: FYI: It doesn't actually look like this is ever called.

gen/gen.go
```diff
@@ -0,0 +1,218 @@
...
+				v, err := strconv.ParseUint(s, 10, bitSize)
+				if err != nil {
+					panic(err)
```
ben-clayton, line 35: Template functions can return an error, so you don't have to use panic. e.g.

`func(s string, bitSize int) uint64` -> `func(s string, bitSize int) (uint64, error)`

gen/gen.go
```diff
@@ -0,0 +1,218 @@
...
+			},
+			"ArrayType": func(typ string, pointer PointerType) string {
+				matches := arrayTypeRegexp.FindStringSubmatch(typ)
+				if len(matches) == 2 {
+					return CType(matches[1], pointer)
+				}
+				return ""
+			},
+			"Singularize": Singularize,
+			"IsLast": func(i int, s interface{}) bool {
```
ben-clayton, line 50:
```suggestion
			"IsLast": func(i int, s any) bool {
```

gen/gen.go
```diff
@@ -0,0 +1,218 @@
...
+	case strings.HasPrefix(typ, "object."):
+		ctype := "WGPU" + PascalCase(strings.TrimPrefix(typ, "object."))
+		return appendModifiers(ctype, pointerType)
+	default:
+		return ""
+	}
+}
+
+func FunctionArgs(f Function, o *Object) string {
+	var sb strings.Builder
```
ben-clayton, line 163:
```suggestion
	sb := &strings.Builder{}
```

Then you can drop all the `&` on the `Fprintf` calls.

gen/utils.go
```diff
@@ -0,0 +1,135 @@
...
+
+func ConstantCase(v string) string {
+	return strings.ToUpper(v)
+}
+
+func PascalCase(s string) string {
+	var out strings.Builder
+	out.Grow(len(s))
+	nextUpper := true
+	for i := 0; i < len(s); i++ {
```
ben-clayton, line 75:
```suggestion
	for _, c := range s {
```

Which will correctly decode the runes from the string. You can then remove the `rune(c)` from below.

gen/utils.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+	}
+	return out.String()
+}
+
+func CamelCase(s string) string {
+	var out strings.Builder
+	out.Grow(len(s))
+	nextUpper := false
+	for i := 0; i < len(s); i++ {
```
ben-clayton, line 95: ditto

gen/cheader.tmpl
```diff
@@ -0,0 +1,234 @@
...
+#if !defined(WGPU_FUNCTION_ATTRIBUTE)
+#define WGPU_FUNCTION_ATTRIBUTE
+#endif
+#if !defined(WGPU_NULLABLE)
+#define WGPU_NULLABLE
+#endif{{"\n" -}}
+
+{{if eq .Name "webgpu"}}
+#include <stdint.h>
+#include <stddef.h>{{"\n" -}}
```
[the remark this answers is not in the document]
rajveermalviya: I went through a pass of all `{{"\n" -}}` and removed all the ones that were unnecessary or replaceable.
Some are still there but for a reason, `{{"\n" -}}` were there to avoid unneeded newlines emitted from template. For example:

<table>
<tr>
<td>  </td>
<td> Template </td>
<td> Generates </td>
</tr>
<tr>
<td> 1. </td>
<td>

```go
#if !defined(WGPU_NULLABLE)
#define WGPU_NULLABLE
#endif{{"\n" -}}

{{if eq .Name "webgpu"}}
#include <stdint.h>
#include <stddef.h>
{{else}}
#include "webgpu.h"
{{end}}
```

</td>
<td>

```c
#if !defined(WGPU_NULLABLE)
#define WGPU_NULLABLE
#endif

#include <stdint.h>
#include <stddef.h>
```

</td>
</tr>

<tr>
<td> 2. </td>
<td>

```go
#if !defined(WGPU_NULLABLE)
#define WGPU_NULLABLE
#endif

{{if eq .Name "webgpu"}}
#include <stdint.h>
#include <stddef.h>
{{else}}
#include "webgpu.h"
{{end}}
```

</td>
<td>

```c
#if !defined(WGPU_NULLABLE)
#define WGPU_NULLABLE
#endif


#include <stdint.h>
#include <stddef.h>
```

</td>
</tr>

<tr>
<td> 3. </td>
<td>

```go
#if !defined(WGPU_NULLABLE)
#define WGPU_NULLABLE
#endif
{{if eq .Name "webgpu"}}
#include <stdint.h>
#include <stddef.h>
{{else}}
#include "webgpu.h"
{{end}}
```

</td>
<td>

```c
#if !defined(WGPU_NULLABLE)
#define WGPU_NULLABLE
#endif

#include <stdint.h>
#include <stddef.h>
```

</td>
</tr>

<tr>
<td> 4. </td>
<td>

```go
#if !defined(WGPU_NULLABLE)
#define WGPU_NULLABLE
#endif

{{- if eq .Name "webgpu"}}
#include <stdint.h>
#include <stddef.h>
{{else}}
#include "webgpu.h"
{{end}}
```

</td>
<td>

```c
#if !defined(WGPU_NULLABLE)
#define WGPU_NULLABLE
#endif
#include <stdint.h>
#include <stddef.h>
```

</td>
</tr>
</table>

Here only `1.` & `3.`  are producing desired the result, and `1.` allows us to have a newline in template, matching the generated result.

gen/cheader.tmpl
rajveermalviya: Similar reasons to `{{"\n" -}}` - to avoid unnecessary newlines. But now removed all instances.

gen/gen.go
rajveermalviya: `ParseUint` is used for enum & bitflags blocks.

ben-clayton: I see that it's used in the templates, but it doesn't ever appear to be called. You can replace the body with a `panic("!")` and see that it is never called.

ben-clayton (commented):
Go / template readability LGTM. Thank you for all the changes.
Review still needed from project maintainers.

gen/gen.go
rajveermalviya: Correct, it'll only get called if in the yaml an enum/bitflag entry has a custom `value` property, which is optional.
Current `webgpu.yml` doesn't need them currently, but other custom specs may need them.

ben-clayton: 👍🏼 Just checking that this was intentional. SGTM

kainino0x:
Sorry for the delay, I've been having a hard time catching up with work. Thank you Ben for the Go readability review!

One high level comment - I'd like to have a blocking GitHub Actions check that the generated output is up-to-date. This shouldn't be too complicated, just run the generator and then check that the git repository is clean.

kainino0x:
(Oh, and - since we are generated files already - I wonder if we could also generate a JSON version of the YAML file, for easy ingestion by other projects? But let's save this for a future PR.)

gen/gen.go
kainino0x, line 1: I stumbled upon something - I remember you mentioned this would probably make GitHub say the project is written in Go.
Apparently we can do something like
`gen/** -linguist-detectable`
in .gitattributes to avoid this:
https://dev.to/katkelly/changing-your-repo-s-language-in-github-5gjo
https://github.com/github-linguist/linguist/blob/master/docs/overrides.md

schema.json
```diff
@@ -0,0 +1,488 @@
...
+            "description": "The dedicated enum prefix for the implementation specific header to avoid collisions"
+        },
+        "constants": {
+            "type": "array",
+            "items": {
+                "type": "object",
+                "properties": {
+                    "name": {
+                        "type": "string",
+                        "description": "Name of the constant variable/define"
```
kainino0x, line 20: (Can be future work)
Maybe we could use regex patterns to validate these a little more?
https://json-schema.org/understanding-json-schema/reference/string#regexp

These would get peppered all over the place so we'd probably want to link to the definition by reference:
https://stackoverflow.com/a/62585173/1268131

schema.json
```diff
@@ -0,0 +1,488 @@
...
+                    },
+                    "value": {
+                        "type": "string",
+                        "description": "An unsigned integer in decimal or hexadecimal syntax, in string"
```
kainino0x, line 24: Regex patterns are useful here too: `^(0x[0-9A-F]{4}|0|[1-9][0-9]{0,4})$` or something like that

Though perhaps better - since we're using YAML already which supports hexadecimal numbers - could we just accept integers here instead? With range validation:
https://json-schema.org/understanding-json-schema/reference/numeric#range
(Can also be future work)

schema.json
```diff
@@ -0,0 +1,488 @@
...
+                                    "type": "string",
+                                    "description": "Optional property, specifies if a struct member type is a pointer",
+                                    "enum": [
+                                        "immutable",
+                                        "mutable"
+                                    ]
+                                },
+                                "optional": {
+                                    "type": "boolean",
+                                    "description": "Optional property, to indicate if a struct member is optional"
```
kainino0x, line 262: (Can be future work)
For the INIT macros (and generating C++ bindings), some optional members will need to know what their default value is (like WGPU_DEPTH_SLICE_UNDEFINED). Probably we can just put that here, though an alternative could be to move that information into the "type system" and give the `depthSlice` member a special type that knows its default.

schema.json
```diff
@@ -0,0 +1,488 @@
...
+                                    "type": "string",
+                                    "description": "Argument type"
+                                },
+                                "pointer": {
+                                    "type": "string",
+                                    "description": "Optional property, specifies if a function argument type is a pointer",
+                                    "enum": [
+                                        "immutable",
+                                        "mutable"
+                                    ]
```
kainino0x, line 331: (Can be future work)
This is used a bunch of times and I think could be defined by `$ref` as well (link again: https://stackoverflow.com/a/62585173/1268131)

`functions.items.properties.args` and `objects.items.properties.methods.items.properties.args` also could probably be the same, and maybe `structs.items.properties.members` too depending if it can be made to match (dawn does this too)

webgpu.h
```diff
@@ -448,7 +425,7 @@ typedef enum WGPUSType {
     WGPUSType_SurfaceDescriptorFromWaylandSurface = 0x00000008,
     WGPUSType_SurfaceDescriptorFromAndroidNativeWindow = 0x00000009,
     WGPUSType_SurfaceDescriptorFromXcbWindow = 0x0000000A,
-    WGPUSType_RenderPassDescriptorMaxDrawCount = 0x0000000F,
+    WGPUSType_RenderPassDescriptorMaxDrawCount = 0x0000000B,
```
kainino0x, line 428: I think the YAML could specify a value for this so that it doesn't get renumbered? (We're going to need to renumber all these anyway later but it's convenient to minimize the differences for now)

gen/cheader.tmpl
```diff
@@ -0,0 +1,232 @@
...
+{{-   end}}
+    WGPU{{.Name | PascalCase}}_Force32 = 0x7FFFFFFF
+} WGPU{{.Name | PascalCase}} WGPU_ENUM_ATTRIBUTE;
+typedef WGPUFlags WGPU{{.Name | PascalCase}}Flags WGPU_ENUM_ATTRIBUTE;
+{{  end}}
+
+{{- if eq .Name "webgpu"}}
+typedef void (*WGPUProc)(void) WGPU_FUNCTION_ATTRIBUTE;
+{{- end}}
+
```
kainino0x, line 122: nit: this blank line disappears from the generated result, but I think it would be more readable if it were kept

gen/cheader.tmpl
```diff
@@ -0,0 +1,232 @@
...
+{{- range .FunctionTypes}}
+{{-   MComment .Doc 0}}
+typedef {{FunctionReturns .}} (*WGPU{{.Name | PascalCase}})({{FunctionArgs . nil}}) WGPU_FUNCTION_ATTRIBUTE;
+{{- end}}
+
```
kainino0x, line 127: same for this blank line

kainino0x (approved):
I don't have much to say about the Go code. I don't know Go, but it seems good, and if it works, it works :) Looking at the schema, it looks really great!

The changes to webgpu.h look really easy to reproduce in Dawn (except the callback definition order will require a little work), so we can stay in sync until we start using this generator in Dawn (hopefully).

I added some comments for future work, but let's probably do those in another PR.
Just a few little things I'd like to do before landing this, but they're not critical.

schema.json
rajveermalviya: default values are planned - In-order to keep this initial PR minimal, a future PR which includes both the header generation and default values populated in the yaml at once would be better.

gen/main.go
```diff
@@ -0,0 +1,221 @@
...
+// 	for _, f := range functions {
+// 		yml.Functions = append(yml.Functions, f)
+// 	}
+// 	for _, s := range structs {
+// 		yml.Structs = append(yml.Structs, s)
+// 	}
+// 	for _, o := range objects {
+// 		yml.Objects = append(yml.Objects, o)
+// 	}
+// }
```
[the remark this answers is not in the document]
rajveermalviya: removed for now, plan is to add it back later (for finding invalid duplicates across multiple yaml specs).

kainino0x (approved):
LGTM! Nice reduction on the schema, thanks for digging deeper into what we could do with ref.

kainino0x:
I've left the future work comments "unresolved" as a reference.

kainino0x:
I'll go ahead and merge since webgpu.h changes are minimal - in Dawn we can hopefully catch up soon so the sync process won't be made harder.

Issue #252, opened 2023-12-01
Consider having a "count" for all enumerate calls.


Kangz:
At the moment the `EnumerateFeatures` calls only take a pointer to the list of things to write. This is technically ok when the size of the list never changes, but:

 1 - In the future we might have list that change sizes, like for an `EnumerateAdapters`
 1 - Even for fixed-size lists, it is good hygiene in C APIs to always pass the size.

kainino0x:
Dec 14 meeting:

*   CW: I implemented an EnumerateWGSLLanguageFeatures in Dawn. Don’t think this is upstream yet but I assume we want it in the instance.
*   CW: Problem with enumeration, say adapters: you call it once with null, it gives a count, you allocate your array, call it again - and the number of adapters in the system has changed. It may write past the end of the array. So probably need take count as an input in enumerate.
*   RM: … Adding count seems fine
*   **Add a count input. Still return a count output. The return count could be larger than the input count, so you know you didn’t get all of them.**

kainino0x:
Just had a thought - I can't remember if we talked about this already, months ago - why don't we change this so you _don't_ need to make multiple calls? In the worst case, a call-twice API changes size every time you call it, so you need to call it in a loop until you've managed to get everything it's going to output. Which is awful because almost no one is probably going to actually write that loop.

Since we're going to break these anyway, should we refactor them to use FreeMembers instead?

```c
typedef struct WGPUAdapterFeatures { // non-extensible
    size_t featureCount;
    WGPUFeatureName const * features;
} WGPUAdapterFeatures WGPU_STRUCTURE_ATTRIBUTE;

WGPU_EXPORT WGPUStatus wgpuAdapterGetFeatures(WGPUAdapter adapter, WGPUAdapterFeatures * features) WGPU_FUNCTION_ATTRIBUTE;

WGPU_EXPORT void wgpuAdapterFeaturesFreeMembers(WGPUAdapterFeatures features) WGPU_FUNCTION_ATTRIBUTE;
```

Or possibly the same but merged with WGPUSupportedLimits (see also #260).

eliemichel:
The worst case scenario is very unlikely, and afaik other battlefield tested APIs use this kind of call-twice API that lets the end user manage memory in their preferred way? I have no strong opinion here though, both options sound reasonable.

kainino0x:
[...]

PR #262, opened 2024-01-08
Add wgpuInstanceHasWGSLLanguageFeature

beaufortfrancois:
@Kangz Please review

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -1450,7 +1450,6 @@ typedef void (*WGPUProcDeviceRelease)(WGPUDevice device) WGPU_FUNCTION_ATTRIBUTE
 
 // Procs of Instance
 typedef WGPUSurface (*WGPUProcInstanceCreateSurface)(WGPUInstance instance, WGPUSurfaceDescriptor const * descriptor) WGPU_FUNCTION_ATTRIBUTE;
-typedef size_t (*WGPUProcInstanceEnumerateWGSLLanguageFeatures)(WGPUInstance instance, WGPUWGSLFeatureName * features) WGPU_FUNCTION_ATTRIBUTE;
 typedef WGPUBool (*WGPUProcInstanceHasWGSLLanguageFeature)(WGPUInstance instance, WGPUWGSLFeatureName feature) WGPU_FUNCTION_ATTRIBUTE;
 typedef void (*WGPUProcInstanceProcessEvents)(WGPUInstance instance) WGPU_FUNCTION_ATTRIBUTE;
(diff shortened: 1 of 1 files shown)


Kangz (approved):

webgpu.h
```diff
@@ -1675,6 +1686,8 @@ WGPU_EXPORT void wgpuDeviceRelease(WGPUDevice device) WGPU_FUNCTION_ATTRIBUTE;
 
 // Methods of Instance
 WGPU_EXPORT WGPUSurface wgpuInstanceCreateSurface(WGPUInstance instance, WGPUSurfaceDescriptor const * descriptor) WGPU_FUNCTION_ATTRIBUTE;
+WGPU_EXPORT size_t wgpuInstanceEnumerateWGSLLanguageFeatures(WGPUInstance instance, WGPUWGSLFeatureName * features) WGPU_FUNCTION_ATTRIBUTE;
```
kainino0x, line 1689: We may want to make the change we decided on in #252 before landing this. Otherwise we'll add this and then change it soon.

That said, having considered what we'd change here, I'm going to comment there about whether that's actually what we want to do.

beaufortfrancois: When do you think these changes will be made in webgpu.h?

We're going to ship ReadonlyAndReadwriteStorageTextures, Packed4x8IntegerDotProduct, UnrestrictedPointerParameters, and PointerCompositeAccess in Chrome 122 and later and it'd be nice if developers can check whether those WGSL language features are supported or not.

beaufortfrancois: How about I remove `wgpuInstanceEnumerateWGSLLanguageFeatures` from this CL and only add `wgpuInstanceHasWGSLLanguageFeature `? What do you think?

kainino0x: Most often webgpu.h changes after Dawn changes - Dawn implements them in our generator (even if only behind the "upstream" tag) and then we "roll" it into this repository.

Adding just HasWGSLLanguageFeature for now would be fine, then we can land most of this PR and deal with the rest soon after.

I'll also note it is not critical for changes to land here before they are implemented in Emscripten. Emscripten actually tracks the Dawn API and not the upstream webgpu.h API right now, for practical reasons for people cross-targeting Dawn and Emscripten.

Still to avoid downstream projects depending on the wgpuInstanceEnumerateWGSLLanguageFeatures API before we fix it, I think Emscripten would also best implement only HasWGSLLanguageFeatures.

beaufortfrancois: Thanks for the detailed answer @kainino0x 
I've updated this PR and sent a PR for Emscripten at https://github.com/emscripten-core/emscripten/pull/21055

rajveermalviya (approved):

kainino0x (approved):

PR #263, opened 2024-01-10
Add depthSlice to WGPURenderPassColorAttachment

beaufortfrancois:
See https://gpuweb.github.io/gpuweb/#dom-gpurenderpasscolorattachment-depthslice

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -69,6 +69,7 @@
 
 #define WGPU_ARRAY_LAYER_COUNT_UNDEFINED (0xffffffffUL)
 #define WGPU_COPY_STRIDE_UNDEFINED (0xffffffffUL)
+#define WGPU_DEPTH_SLICE_UNDEFINED (0xffffffffUL)
 #define WGPU_LIMIT_U32_UNDEFINED (0xffffffffUL)
 #define WGPU_LIMIT_U64_UNDEFINED (0xffffffffffffffffULL)
 #define WGPU_MIP_LEVEL_COUNT_UNDEFINED (0xffffffffUL)
--- a/webgpu.h
+++ b/webgpu.h
@@ -1213,6 +1213,7 @@ typedef struct WGPUProgrammableStageDescriptor {
 typedef struct WGPURenderPassColorAttachment {
     WGPUChainedStruct const * nextInChain;
     WGPU_NULLABLE WGPUTextureView view;
+    uint32_t depthSlice;
     WGPU_NULLABLE WGPUTextureView resolveTarget;
     WGPULoadOp loadOp;
     WGPUStoreOp storeOp;


beaufortfrancois:
> We also need DEPTH_SLICE_UNDEFINED

Good catch!

Kangz (approved):

rajveermalviya (approved):

kainino0x:
Fixes #215

yig:
With this change, can users still zero-initialize the struct? Will `.depthSlice` be ignored for non-3D attachments? (Dawn currently throws a validation error.)

yig:
[Sorry that I'm late to the discussion. I teach a fall course using WebGPU, so I'm catching up on breaking changes in the last year.]

I see that the original proposal <https://github.com/gpuweb/gpuweb/issues/4251> said about `WGPURenderPassColorAttachment.depthSlice`:

> Default to 0 and ignored for the `view` not created on 3d texture.

However, the spec <https://www.w3.org/TR/webgpu/> states that unless `renderViewDescriptor.dimension` is "3d":

> `this.depthSlice` must not be provided.

Dawn's implementation validates this according to the spec, but then immediately sets it back to 0: <https://github.com/google/dawn/commit/f08036cc67afbadce0a61615eab20629a6c9f187#diff-f2b2b0315ddb05f08b00c60e82a5454417ed7cb2a9a7fb7f150c3752aeecb9a0L1118-L1124>

As a user of webgpu from C, I prefer the original proposal (ignoring this field for 2d textures). It means there are fewer fields to manually set for the common case (2d textures) versus zero-initialization.

yig:
For what it's worth, the spec states about `WGPURenderPassColorAttachment.clearValue`:

> Ignored if [loadOp](https://www.w3.org/TR/webgpu/#dom-gpurenderpasscolorattachment-loadop) is not ["clear"](https://www.w3.org/TR/webgpu/#dom-gpuloadop-clear).

I would definitely not argue for consistency's sake that `clearValue` should have another `UNDEFINED` value created for it (or become a nullable type in C).

kainino0x:
As a general rule it needs to be possible to implement the JS API over the C API (as well as vice versa). So if the JS API's validation needs to be able to distinguish `0` from `undefined`, then we need to be able to express that in C too.

It's technically possible for JS-over-C implementations to inject special validation for those cases, rather than express it through the C API, but we have avoided relying on this pattern where possible.

There's unfortunately a tension between the usability of the C API and matching expressiveness with the JS API. Right now on the Dawn side, we are focusing on making the C++ API try to solve usability problems like this. The idea has also been floated to make nicer-to-use C bindings implemented over `webgpu.h`, but we don't have any efforts planned on that.

yig:
That makes sense. I am suggesting a change to the standard (affecting both JS and C). The standard currently says to ignore `.clearValue` if `.loadOp` is not "clear". My suggestion is to similarly ignore `.depthSlice` if the texture is not 3D.

kainino0x:
Ah, sorry I misunderstood. That's definitely possible, and there are a number of similar strict validation rules in the JS spec that I've been on the fence about pushing to lift. Fortunately, they are non-breaking changes (which is why we chose the strict thing in the first place).

Filed https://github.com/gpuweb/gpuweb/issues/4824

Issue #266, opened 2024-01-22
wgpuAdapterGetProperties is sync but requestAdapterInfo is async


kainino0x:
_Originally posted by @beaufortfrancois in https://github.com/emscripten-core/emscripten/issues/21072#issuecomment-1904177995_

> I assume it's because of the async nature of it. All the code in this file using promises have callback while this one has not.

_Originally posted by @kainino0x in https://github.com/emscripten-core/emscripten/issues/21072#issuecomment-1905001256_

> Hm, right... I forgot that it was async in JS but sync in C.
> 
> I think to make this work we would need to have wgpuInstanceRequestAdapter actually do both requestAdapter AND requestAdapterInfo, then store off the results *in case* wgpuAdapterGetProperties is called later. I'm not sure if this is what we had in mind. It would work, but would make wgpuInstanceRequestAdapter slightly slower, and also would unnecessarily access fingerprintable info that the page may not actually need.
> 
> Alternatives would be:
> - Make it async in C
> - Somehow include in the wgpuInstanceRequestAdapter call whether you want to retrieve the properties or not

beaufortfrancois:
Would it make sense to replace `wgpuAdapterGetProperties` with `wgpuAdapterRequestAdapterInfo` instead?

Kangz:
> Somehow include in the wgpuInstanceRequestAdapter call whether you want to retrieve the properties or not

Seems sensible? At least in C it doesn't seem super necessary to keep the retrieval of info async.

rajveermalviya:
Seems like this is something emscripten layer would have to workaround, because in the native it doesn't make sense to make it async, info callbacks can be especially annoying when enumerating adapters.

> and also would unnecessarily access fingerprintable info that the page may not actually need.

not sure how to avoid that

kainino0x:
> > and also would unnecessarily access fingerprintable info that the page may not actually need.
> 
> not sure how to avoid that

That's the idea of "Somehow include in the wgpuInstanceRequestAdapter call whether you want to retrieve the properties or not": if you don't ask for it in RequestAdapterOptions, then we don't need to add the extra delay or the extra data access.

I think that's a pretty good option. Maybe it would look something like a new field in RequestAdapterOptions that's a bitfield with just 1 bit so far (0 = don't requestAdapterInfo(), 1 = do requestAdapterInfo()). Later on we might have extra bits to ask for more info.

kainino0x:
Feb 1 meeting:

[...]

PR #269, opened 2024-02-08
Add wgpuAdapterRequestAdapterInfo

beaufortfrancois:
As discussed https://github.com/webgpu-native/webgpu-headers/issues/266#issuecomment-1932532969, this PR adds WGPUAdapterRequestAdapterInfo.


kainino0x (commented):
Hm, I forgot we also wanted to remove wgpuAdapterGetProperties.
I also forgot WGPUAdapterProperties has extra info in it.
We'll need a proposal, I'm not sure if we want to move the extra info to an extension of WGPUAdapterInfo (how?) or keep GetProperties for this extra info (but as a native-only extension with no implementation in Emscripten)

kainino0x (approved):
I think this looks about right for now though.

Issue #214, opened 2023-08-17
Allocation of implementation-specific enum values


kainino0x:
We should have a "registry" of allocated ranges to use for enums, especially WGPUSType and WGPUFeatureName
Per #212 the end of the name of the new values will be the name of the implementation ("...Dawn" or "...Wgpu")

In the webgpu.h meeting we agreed to use the top 16 bits for the implementation namespace. I think we can use this for all of the (non-bitflag) enums.

No one really cares about the actual IDs so here's an arbitrary assignment of ranges:

- `0x0000_????`: core
  Suffix: none.
  
  Should be in all implementations, both native and Wasm. Used for features added to WebGPU core after this header stabilizes.
- `0x0001_????`: multi-implementation C-specific extensions
  Suffix: none.
  
  May be native-specific or Wasm-specific, both can go in this block.
- `0x0002_????`: WebGPU Compat
  Suffix: none? TBD.
  
  Special: implementations that don't support compat should ignore these instead of error. This is to be used only for Compat additions, which don't affect semantics of a valid program if a non-compat implementation ignores them.
- `0x0003_????`: wgpu-native
  Suffix: `_WGPU`
- `0x0004_????`: Emscripten
  Suffix: `_Emscripten`
- `0x0005_????`: Dawn
  Suffix: `_Dawn`

---

EDIT: For bitflags (becoming `uint64_t` in #273):

- The least-significant 53 bits are reserved for core - that is, bits defined by the JS API, since it can only represent 53 bits.
- The most-significant 11 bits are for native-only and implementation-specific bits, then overflows into the most-significant end of the least-significant 53 bits if we run out. The allocation of these bits is ad hoc for now.
  - Only bits specified in this header will be stable; implementation-specific bits can overlap or change value for now. Eventually for ABI compatibility we'll probably need a registry of these.

rajveermalviya:
`wgpu-native` now uses the `0x0003_????` namespace

Kangz:
> `0x0001_????`: (reserved)

Shouldn't this be reserved for native-only and not upstream WebGPU (as in the JS spec) features? If we need another reserved block Dawn can move to `0x0004_????` so `0x0002_????` is free.

kainino0x:
Good idea. We don't really need another reserved block so I'll just change what `0x0001_????` is, above.

kainino0x:
[...]

Issue #159, opened 2022-06-03
Flag enum types and flags typedefs


kainino0x:
_Originally posted by @mrshannon in https://github.com/webgpu-native/webgpu-headers/issues/158#issuecomment-1146170716_:

> > Any idea on how to improve WGPUBufferUsage vs. WGPUBufferUsageFlags?
> 
> You technically don't need `WGPUBufferUsageFlags` as `WGPUBufferUsage` will > always be wide enough for the bitwise combination because we have `WGPUBufferUsage_Force32 = 0x7FFFFFFF` which is effectively doing what is done for the C++ headers with:
> ```c++
> enum class BufferUsage : uint32_t {
> ```

This is strictly true, but in Dawn's C++ bindings we also have operator overloads, so when you do `wgpu::BufferUsage::CopySrc | wgpu::BufferUsage::CopyDst` you get a `wgpu::BufferUsage` value. In C `WGPUBufferUsage_CopySrc | WGPUBufferUsage_CopyDst` will give you an int of some kind. So this would force callers of the API to cast back to the appropriate type.

```c
desc.usage = static_cast<WGPUBufferUsage>(WGPUBufferUsage_CopySrc | WGPUBufferUsage_CopyDst);
```

Apparently, the cast is only needed in C++. C seems happy without a cast. So I guess we could define operator overloads for these types behind `#ifdef __cplusplus`...?

kainino0x:
i.e.

```c
#ifdef __cplusplus
WGPUBufferUsage operator|(WGPUBufferUsage a, WGPUBufferUsage b) {
    return static_cast<WGPUBufferUsage>(a | b);
}
```

and so on

kainino0x:
FWIW, we copied this pattern from Vulkan:
https://github.com/webgpu-native/webgpu-headers/pull/1#discussion_r303507976

in any case, whatever we do, we need to make sure it works fully in both C and C++.

Kangz:
Adding operators only in C++ if that's where it causes problems would make sense. We could have something similar to Dawn's [`EnumClassBitmask.h`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/include/dawn/EnumClassBitmasks.h;l=16?q=EnumClassBit&sq=).

> in any case, whatever we do, we need to make sure it works fully in both C and C++.

Yeah right now webgpu.h doesn't work in C because of #84 even.

kainino0x:
[...]

PR #270, opened 2024-02-11
gen: extended headers

rajveermalviya:
This PR finalizes the generations of implementation specific headers from custom yaml specifications aside from `webgpu.yml`.

Apart from the general struct extensions, the PR now allows extending `enums`, `bitflags` and `objects` with extra entires and functions across multiple yaml specifications.

#### Enums

Because C/C++ can't allow extending enums, we need to use casts, the downside being we loose exhaustive switches in C++ over new entries.

```c
#if !defined(__WGPU_EXTEND_ENUM)
#ifdef __cplusplus
#define __WGPU_EXTEND_ENUM(E, N, V) static const E N = E(V)
#else
#define __WGPU_EXTEND_ENUM(E, N, V) static const E N = (E)(V)
#endif
#endif // !defined(__WGPU_EXTEND_ENUM)

__WGPU_EXTEND_ENUM(WGPUFeatureName, WGPUFeatureName_PushConstants, 0x30000000);
__WGPU_EXTEND_ENUM(WGPUFeatureName, WGPUFeatureName_MultiDrawIndirect, 0x30000001);

__WGPU_EXTEND_ENUM(WGPUSType, WGPUSType_ShaderModuleGLSLDescriptor, 0x30000000);
__WGPU_EXTEND_ENUM(WGPUSType, WGPUSType_InstanceExtras, 0x30000001);
```

#### Bitflags

Currently extended bitflag entries are similar to enums, but I think we should change generation of all bitflag entries to be `static const T N = V` instead of enum, because we need to make them 64bit anyway. [Vulkan does the same for 64bit bitflags](https://github.com/KhronosGroup/Vulkan-Headers/blob/5ac36269f50381bdd92a5e1973d8eb041771e59e/include/vulkan/vulkan_core.h#L6538)

```c
__WGPU_EXTEND_ENUM(WGPUShaderStage, WGPUShaderStage_Mesh, 0x00000008);
```

### webgpu_ext.h

Though not needed, but with this PR we can now create a separate common extension header `webgpu_ext.h` and `webgpu_core.h` for core API. So that we can move any native-specific extensions (SPIRV shader) that we currently have in `webgpu.h` to `webgpu_ext.h` and `webgpu_core.h` would try to contain only JS spec API.

### Suffixes

Generator currently doesn't implicitly add any suffixes (`_DAWN`, `_WGPU`) to non-extended identifiers in extension headers, those will need to be added in the corresponding yaml specification explicitly, which I think is what we want because it's flexible that way.

### DIFF
--- a/gen/cheader.tmpl
+++ b/gen/cheader.tmpl
@@ -84,7 +84,7 @@ struct WGPU{{.Name | PascalCase}};
 {{- range $enum := .Enums}}
 {{-   if .Extended}}
 {{-     range $entryIndex, $_ := .Entries}}
-__WGPU_EXTEND_ENUM(WGPU{{ $enum.Name | PascalCase }}, WGPU{{ $enum.Name | PascalCase}}_{{.Name | PascalCase}}, {{EnumValue $.EnumPrefix $enum $entryIndex}});
+__WGPU_EXTEND_ENUM(WGPU{{$enum.Name | PascalCase}}, WGPU{{$enum.Name | PascalCase}}_{{.Name | PascalCase}}, {{EnumValue $.EnumPrefix $enum $entryIndex}});
 {{-     end}}
 {{-   else}}
 {{-     MComment .Doc 0}}
@@ -99,15 +99,21 @@ typedef enum WGPU{{.Name | PascalCase}} {
 {{  end}}
 
 {{- range $bitflag := .Bitflags}}
-{{-   MComment .Doc 0}}
+{{-   if .Extended}}
+{{-     range $entryIndex, $_ := .Entries}}
+__WGPU_EXTEND_ENUM(WGPU{{$bitflag.Name | PascalCase}}, WGPU{{$bitflag.Name | PascalCase}}_{{.Name | PascalCase}}, {{BitflagValue $bitflag $entryIndex}});
+{{-     end}}
+{{-   else}}
+{{-     MComment .Doc 0}}
 typedef enum WGPU{{.Name | PascalCase}} {
-{{-   range $entryIndex, $_ := .Entries}}
-{{-     MComment .Doc 4}}
+{{-     range $entryIndex, $_ := .Entries}}
+{{-       MComment .Doc 4}}
     WGPU{{$bitflag.Name | PascalCase}}_{{.Name | PascalCase}} = {{BitflagValue $bitflag $entryIndex}},
-{{-   end}}
+{{-     end}}
     WGPU{{$bitflag.Name | PascalCase}}_Force32 = 0x7FFFFFFF
 } WGPU{{$bitflag.Name | PascalCase}} WGPU_ENUM_ATTRIBUTE;
 typedef WGPUFlags WGPU{{$bitflag.Name | PascalCase}}Flags WGPU_ENUM_ATTRIBUTE;
+{{-   end}}
 {{  end}}
 
 {{- if eq .Name "webgpu"}}
--- a/gen/cheader.tmpl
+++ b/gen/cheader.tmpl
@@ -187,7 +187,7 @@ typedef WGPUProc (*WGPUProcGetProcAddress)(WGPUDevice device, char const * procN
 {{-     MComment .Doc 0}}
 typedef {{FunctionReturns .}} (*WGPUProc{{$object.Name | PascalCase}}{{.Name | PascalCase}})({{FunctionArgs . $object}}) WGPU_FUNCTION_ATTRIBUTE;
 {{-   end}}
-{{-   if not .IsStruct}}
+{{-   if not (or .IsStruct .Extended)}}
 typedef void (*WGPUProc{{.Name | PascalCase}}Reference)(WGPU{{.Name | PascalCase}} {{.Name | CamelCase}}) WGPU_FUNCTION_ATTRIBUTE;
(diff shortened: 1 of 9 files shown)


rajveermalviya:
Here's a sample with yaml and corresponding generated header.
https://gist.github.com/rajveermalviya/812ee0398ca88c1962469a0978a4845c

kainino0x:
Sorry for the review delay. Could have discussed some of these today, probably. I had a lot to look at here though!

> Currently extended bitflag entries are similar to enums, but I think we should change generation of all bitflag entries to be `static const T N = V` instead of enum, because we need to make them 64bit anyway. [Vulkan does the same for 64bit bitflags](https://github.com/KhronosGroup/Vulkan-Headers/blob/5ac36269f50381bdd92a5e1973d8eb041771e59e/include/vulkan/vulkan_core.h#L6538)

Ah... unfortunate. I dug a little to figure out why this was the case in Vulkan. Clang makes it clear - and this is also the reason Force32 is 0x7fffffff instead of 0xffffffff.

`warning: ISO C restricts enumerator values to range of 'int' (4294967295 is too large) [-Wpedantic]`

C++ seems to have no such complaints for `enum class`, so Dawn should still be able to do the thing we do in C++ to make strongly-typed bitflags from `enum class`.
C++ _also_ seems to have no complains for `enum`, which means we _could_ use `enum` behind `#ifdef __cplusplus`, for added type safety. Doing so would be consistent with #159, where we were going to do the `operator|` thing to make C++ happy with our usage of `enum`s as bitflags. But having separate C and C++ definitions of entire enums is annoying, and also sketchy - it almost certainly breaks ABI compatibility between the C and C++ versions.
So I think we are going to have to scrap the `operator|` thing and just use 64-bit enums. Too bad.

> the downside being we loose exhaustive switches in C++ over new entries

Actually this raises another interesting point - if we extend enums outside the enum definition, then it becomes easy to write a switch that passes the exhaustiveness check, but is not actually exhaustive. Of course, this was always true of bitflags, and generally true if you cast numbers to enums, but those are corner cases.

This is probably fine, it just adds a little bit of a false sense of security. Almost certainly better to get some warnings than no warnings.
Most enums won't be commonly in application code anyway. The enums we do pass out to the application may get extended, but for *most* enums, we should never start returning new values to existing applications unless we enable some feature, so exhaustivity over the core enum is all that really matters.
For reference, I think the only enums where we may suddenly start passing out new values to the application are `FeatureName` (in AdapterEnumerateFeatures, but not DeviceEnumerateFeatures) and `WGSLFeatureName` (in GetInstanceFeatures or whatever we end up calling it). Not `SType`, I think, if we solve #272.

Maybe... after we stabilize, we should just never change the enum definitions and *only* extend them using `static const`? (maybe even rename Force32 to indicate it's there to make the enum extensible/non-exhaustive, as anyone writing exhaustive switches will end up having to type Force32)

(Aside, apparently there is a gcc/clang flag that checks for exhaustive switches even *with* default. I am now tempted to try this in Dawn. https://brevzin.github.io/c++/2019/08/01/enums-default/#the-warning-that-exists)

kainino0x (approved):
LGTM, let me copy that comment to a new issue.

gen/cheader.tmpl
```diff
@@ -37,7 +37,17 @@
 #define WGPU_NULLABLE
 #endif{{"\n" -}}
 
-{{if eq .Name "webgpu"}}
+{{- if ne .Name "webgpu"}}
+#if !defined(__WGPU_EXTEND_ENUM)
+#ifdef __cplusplus
+#define __WGPU_EXTEND_ENUM(E, N, V) static const E N = E(V)
+#else
+#define __WGPU_EXTEND_ENUM(E, N, V) static const E N = (E)(V)
```
kainino0x, line 45: I wonder if we need a `WGPU_ENUM_EXTENDED_VALUE_ATTRIBUTE` here. I'm not really clear on when Swift needs `NS_REFINED_FOR_SWIFT` exactly, but it seems like it probably would need one here in order to refine our enums into Swift enums?

kainino0x: (This does not block landing the PR. I don't think #179 is fully resolved right now anyway.)

rajveermalviya: Let's just defer this to when someone asks for it.

rajveermalviya:
### Namespace indicators

The extension yaml parameter types can use `namespace` field to indicate where the mentioned type comes from. For example when using the `webgpu.yml` types in `wgpu.yml`.

### re: Suffixes

Now generator has a flag `extsuffix` (on by default), which will insert suffixes to the identifiers in appropriate format. The suffix is identified by the yaml file name.

wgpu-native will now use this generator to generate it's `wgpu.h` header. See gfx-rs/wgpu-native#373 for the new `wgpu.yml` & the resulting generated header.

kainino0x:
> The suffix is identified by the yaml file name.

I think it would be better to put the suffix in the yaml file, otherwise I think we wouldn't be able to split webgpu.yml into multiple files for example? Also generally I find it confusing to figure out where the suffix came from when it's implicit from the filename. (Took me a minute to figure this out even though you wrote it down.)

kainino0x (approved):
Seems good, I think you can land this whenever you want

rajveermalviya:
Generator doesn't do anything implicit for extended bitflags yet, it expects custom values specified in yaml to be distinct.

@kainino0x added #214 back on agenda to specify the format of extended bitflags.

rajveermalviya:
Merging, will make another PR for extended `bitflags` when the format is decided, also because it doesn't affect `wgpu.h` (it doesn't extend any core bitflags yet)

PR #277, opened 2024-03-04
gen: Support typedefs

rajveermalviya:
Add support of specifying new typedefs for the header.

The existing `WGPUFlags` & `WGPUBool` don't belong in `webgpu.yml` because they are one of the C-isms which `webgpu.yml` tries to avoid, and instead it provides `bool` & `bitflag` types.

### DIFF
--- a/gen/cheader.tmpl
+++ b/gen/cheader.tmpl
@@ -65,6 +65,10 @@
 typedef uint32_t WGPUFlags;
 typedef uint32_t WGPUBool;
 {{  end}}
+{{- range .Typedefs}}
+{{-   MComment .Doc 0}}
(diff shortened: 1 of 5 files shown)


kainino0x (approved):
LGTM but I think you're saying we wouldn't use this for flags/bool, so what is it used for? (Something in wgpu-native?)

rajveermalviya:
> Something in wgpu-native

Yes, currently there's atleast one where we expose submissionIDs and provide ability to poll/wait on their completion.
Also for core, maybe the proposed WGPUFuture can also be changed to a typedef uint64 instead of current struct wrapper.

PR #297, opened 2024-06-03
Add uncapturedErrorCallbackInfo to requestDevice descriptor

beaufortfrancois:
This PR adds `uncapturedErrorCallbackInfo` to `requestDevice()` descriptor to match recent Dawn changes at https://dawn-review.googlesource.com/c/dawn/+/183162.

It also removes `wgpuDeviceSetUncapturedErrorCallback` that will be deprecated with this new attribute.

Let me know if that looks good to you.

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -721,6 +721,7 @@ typedef void (*WGPUProc)(void) WGPU_FUNCTION_ATTRIBUTE;
 
 typedef void (*WGPUDeviceLostCallback)(WGPUDeviceLostReason reason, char const * message, void * userdata) WGPU_FUNCTION_ATTRIBUTE;
 typedef void (*WGPUErrorCallback)(WGPUErrorType type, char const * message, void * userdata) WGPU_FUNCTION_ATTRIBUTE;
+typedef void (*WGPUUncapturedErrorCallbackInfo)(WGPUErrorCallback callback, void * userdata) WGPU_FUNCTION_ATTRIBUTE;
 
 typedef void (*WGPUAdapterRequestDeviceCallback)(WGPURequestDeviceStatus status, WGPUDevice device, char const * message, WGPU_NULLABLE void * userdata) WGPU_FUNCTION_ATTRIBUTE;
 typedef void (*WGPUBufferMapAsyncCallback)(WGPUBufferMapAsyncStatus status, WGPU_NULLABLE void * userdata) WGPU_FUNCTION_ATTRIBUTE;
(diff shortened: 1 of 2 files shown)


webgpu.h
```diff
@@ -721,6 +721,7 @@ typedef void (*WGPUProc)(void) WGPU_FUNCTION_ATTRIBUTE;
 
 typedef void (*WGPUDeviceLostCallback)(WGPUDeviceLostReason reason, char const * message, void * userdata) WGPU_FUNCTION_ATTRIBUTE;
 typedef void (*WGPUErrorCallback)(WGPUErrorType type, char const * message, void * userdata) WGPU_FUNCTION_ATTRIBUTE;
+typedef void (*WGPUUncapturedErrorCallbackInfo)(WGPUErrorCallback callback, void * userdata) WGPU_FUNCTION_ATTRIBUTE;
```
kainino0x, line 724: WGPUUncapturedErrorCallbackInfo should be a struct, not a callback:
https://source.chromium.org/chromium/chromium/src/+/main:out/mac-Debug/gen/third_party/dawn/webgpu-headers/webgpu.h;l=1787-1791;drc=e7597cf3fea2525614c4c8c6f8044220287389d1

This file should mostly match the "upstream" version of webgpu.h generated by Dawn (`dawn/webgpu-headers/webgpu.h`). I'm sure it's diverged in other ways currently, but you should try to make the two generators match their results. (We will eventually stop having two separate generators for the same file, but not yet.)

`WGPUDeviceLostCallbackInfo` is also already implemented there, and it would make sense to update both together.

Also note: we are eventually going to want the YAML to define these CallbackInfo structs at a higher level (like, instead of a "struct", a new category for callback infos), so that other language bindings can generate things differently. However, I don't think we need to do this now, as the generator should hopefully be able to generate what we need as a "struct" type.

kainino0x (commented):
(Please also cc @lokokung about changes to sync these headers, since he's usually the one who has implemented them in Dawn.)

beaufortfrancois:
@kainino0x Thanks for the review! I've addressed your feedback and will keep @lokokung updated about upcoming changes.

Note that I'll open a separate PR for `WGPUDeviceLostCallbackInfo`

beaufortfrancois:
`WGPUDeviceLostCallbackInfo` currently contains `WGPUCallbackMode mode` which is not yet defined in webgpu.h. I guess this means we need to wait a bit more before adding it to webgpu.h when WGPUFuture are actually defined. What do you think?

kainino0x:
> `WGPUDeviceLostCallbackInfo` currently contains `WGPUCallbackMode mode` which is not yet defined in webgpu.h. I guess this means we need to wait a bit more before adding it to webgpu.h when WGPUFuture are actually defined. What do you think?

Oh yes, fair point. I forgot DeviceLost is promise-based and UncapturedError is not. Makes sense to separate them then.

I am not sure how soon we'll be ready to roll in the promise changes, but I think we're very close. Maybe when we do that we should deal with the generator thing first before rolling all that in, but TBD.

kainino0x (approved):

rajveermalviya (approved):

PR #306, opened 2024-07-02
Adds 'make fix' command to keep yml file(s) organized

lokokung:
- Adds dependency on yq library to help with formatting
- Adds new go script to format the current webgpu.yml file
- Formats the existing webgpu.yml file so that all the arrays of types are sorted alphabetically.
- Note: one down-side of this is that all the newlines are dropped in the yml file. We may be able to add another layer of formatting later on if this is desired.


kainino0x (approved):
This is great! I think the YAML file changes are fine, personally.

I have been wanting to run some checks/lints on the YAML file to keep it tidy, this pretty much fixes that! I'd considered a "strict" syntax as well where we avoid bare strings, but that's probably not necessary.

Another thing I've wanted to check is whether the schema is being validated (e.g. if some bare value gets interpreted as a number where we wanted a string, do we catch that?)

lokokung:
I think that the schema validations happen as a part of the header generation already. Since we call the generation code as a part of the checks, I **think** that transitively we are validating the schema.

Issue #235, opened 2023-10-18
webgpu.h doesn't represent tri-state depthWriteEnabled 


kainino0x:
depthWriteEnabled is tri-state (true/false/undefined) in the JS API (though this is only for validation).

We should probably have a way to express it in C for the benefit of JS->C and C->JS implementations.
WGPUBool is uint32_t, so we could define a constant for a third value.

Chrome/Dawn deal with this using a chained struct:
```c++
depth_write_defined->depthWriteDefined = webgpu_desc->hasDepthWriteEnabled();
```
https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/webgpu/gpu_render_pipeline.cc;l=154;drc=6e807ee8e768d465bbb9152399004cbdb08e40e9

There may be other similar instances, I just noticed this one.

kainino0x:
webgpu.h meeting (oct 26):
- could add a third value to WGPUBool, but this would be very surprising
- could make a new WGPUOptionalBool type with 3 values **<- do this**
- should 0 be Undefined for zero-init? **no, it should be False, so conversions from C++, stdbool, and WGPUBool will work correctly. Make Undefined=2**

kainino0x:
Sorry, last comment was wrong - I forgot `WGPUBool` was `uint32_t` and not an `enum`. So one more conversion is allowed without warning

- In C, WGPUBool -> enum WGPUOptionalBool, stdbool.h bool -> enum WGPUOptionalBool, and stdbool.h bool -> WGPUBool are all allowed without warning.
- In C++, conversion from WGPUBool is an error, conversion from C++ bool is an error.

PR #308, opened 2024-07-08
Define depthWriteEnabled as WGPUOptionalBool

beaufortfrancois:
FIX https://github.com/webgpu-native/webgpu-headers/issues/235

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -407,6 +407,13 @@ typedef enum WGPUMipmapFilterMode {
     WGPUMipmapFilterMode_Force32 = 0x7FFFFFFF
 } WGPUMipmapFilterMode WGPU_ENUM_ATTRIBUTE;
 
+typedef enum WGPUOptionalBool {
+    WGPUOptionalBool_Undefined = 0x00000000,
+    WGPUOptionalBool_False = 0x00000001,
+    WGPUOptionalBool_True = 0x00000002,
+    WGPUOptionalBool_Force32 = 0x7FFFFFFF
+} WGPUOptionalBool WGPU_ENUM_ATTRIBUTE;
+
 typedef enum WGPUPowerPreference {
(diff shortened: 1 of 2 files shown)


beaufortfrancois:
@kainino0x Can you review?

webgpu.h
```diff
@@ -407,6 +407,13 @@ typedef enum WGPUMipmapFilterMode {
     WGPUMipmapFilterMode_Force32 = 0x7FFFFFFF
 } WGPUMipmapFilterMode WGPU_ENUM_ATTRIBUTE;
 
+typedef enum WGPUOptionalBool {
+    WGPUOptionalBool_Undefined = 0x00000000,
+    WGPUOptionalBool_False = 0x00000001,
+    WGPUOptionalBool_True = 0x00000002,
```
kainino0x, line 413: Should be False=0, True=1, Undefined=2
per https://github.com/webgpu-native/webgpu-headers/issues/235#issuecomment-1791098185

beaufortfrancois: As raised in https://issues.chromium.org/issues/42241167#comment6, I've noticed Dawn asserts that "undefined" values must be 0 in `dawn_json_generator.py`. Shall we loosen this restriction in https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/generator/dawn_json_generator.py;l=147;drc=83d3e7fc97d66a79c944c24b7ad18f549a5a150f?

kainino0x: Oh yes, sorry I forgot to reply to that.
There's a strong reason to have False=0, True=1 so yes, I think we need to do that.

beaufortfrancois: @kainino0x I've updated this PR with your suggested changes.

kainino0x (approved):
LGTM!

Issue #212, opened 2023-08-10
Naming conventions for extension structs and implementation-specific extensions


kainino0x:
We should define some naming conventions for both standardized extension structs and implementation-specific extensions. Right now we have the following (inconsistent) names:

`WGPUPrimitiveState` extensions:
`WGPUPrimitiveDepthClipControl`

`WGPURenderPassDescriptor` extensions:
`WGPURenderPassDescriptorMaxDrawCount`

`WGPUShaderModuleDescriptor` extensions:
`WGPUShaderModuleSPIRVDescriptor`
`WGPUShaderModuleWGSLDescriptor`

`WGPUSurfaceDescriptor` extensions:
`WGPUSurfaceDescriptorFromAndroidNativeWindow`
`WGPUSurfaceDescriptorFromCanvasHTMLSelector`
`WGPUSurfaceDescriptorFromMetalLayer`
`WGPUSurfaceDescriptorFromWaylandSurface`
`WGPUSurfaceDescriptorFromWindowsHWND`
`WGPUSurfaceDescriptorFromXcbWindow`
`WGPUSurfaceDescriptorFromXlibWindow`

Additionally for Dawn we might want things like `WGPUSomeDescriptorNewFeatureDAWN`? (though idk what the naming would be for wgpu-native)

Kangz:
For ergonomics it seems that typing `Descriptor` all the time is annoying. So I'd favor a rule like:

 - Take the name of the chain root, remove `Descriptor`, `State` or other common words
 - Optionally add a name to scope to a project / vendor `Dawn`, `Wgpu`, ...
 - Add the extra words.

So the list would become:

`WGPUPrimitiveState` extensions:
`WGPUPrimitiveDepthClipControl`

`WGPURenderPassDescriptor` extensions:
`WGPURenderPassMaxDrawCount`

`WGPUShaderModuleDescriptor` extensions:
`WGPUShaderModuleSPIRV`
`WGPUShaderModuleWGSL`

`WGPUSurfaceDescriptor` extensions:
`WGPUSurfaceFromAndroidNativeWindow`
`WGPUSurfaceFromCanvasHTMLSelector`
`WGPUSurfaceFromMetalLayer`
`WGPUSurfaceFromWaylandSurface`
`WGPUSurfaceFromWindowsHWND`
`WGPUSurfaceFromXcbWindow`
`WGPUSurfaceFromXlibWindow`

WDYT?

kainino0x:
[...]

PR #309, opened 2024-07-09
Drop Descriptor from chained struct names

beaufortfrancois:
Issue: https://github.com/webgpu-native/webgpu-headers/issues/212#issuecomment-1682801259


kainino0x (approved):
LGTM but this doesn't address the DepthClipControl change, so I'm removing "fix" from the PR summary.

PR #317, opened 2024-07-30
Add `BC Sliced 3D` feature

mehmetoguzderin:
Please let me know if I missed anything or made a misconvention while trying to add this as it seems like putting after BC just changed bitflags.

PTAL @kainino0x (I'm mentioning because I couldn't request review in this repo). Thank you very much!


mehmetoguzderin:
Seems like workflows also need to be approved. Sorry for the burden

kainino0x (approved):
LGTM!

PR #320, opened 2024-08-26
Remove emscripten-only extension; slightly reorder SType

kainino0x:
Emscripten will keep this (under the new name) but it's not "standardized" in this header yet.

> All of the SurfaceSource extensions are going to get moved into the `0x0001_????` block (they're already moved in Dawn but we haven't moved them in here yet).
> 
> So reordering these puts the other 3 STypes in the exact positions I think they'll be when the header becomes stable.

Issue: https://github.com/webgpu-native/webgpu-headers/issues/214#issuecomment-2237722493

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,8 @@
-.PHONY: gen gen-check doc
+.PHONY: help gen gen-check doc
+
+# default target if you just type `make`
+help:
+	@echo 'Targets are: help, fix, gen, gen-check, doc'
 
 fix: webgpu.yml
 	go run ./fix -yaml webgpu.yml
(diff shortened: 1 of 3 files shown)


beaufortfrancois:
Out of curiosity, what was the logic for reordering SType?

kainino0x:
All of the SurfaceSource extensions are going to get moved into a the `0x0001????` block (they're already moved in Dawn but we haven't moved them in here yet).

So reordering these puts the other 3 STypes in the exact positions I think they'll be when the header becomes stable.

lokokung (approved):

PR #321, opened 2024-08-26
Add rgb10a2 vertex format

munrocket:
According to: https://github.com/gpuweb/gpuweb/issues/4275

Value from dawn: https://github.com/google/dawn/blob/efbc7fb59702b1abfdd9b95e1b03cb84678747ca/src/dawn/dawn.json#L4699

For some reason all this values differs in wgpu: 
https://github.com/gfx-rs/wgpu/blob/a93dcb646a6ea040e97950d74e10c757051b35bb/wgpu-types/src/lib.rs#L5202C5-L5202C20


kainino0x (commented):
`webgpu.yml` is the source of truth so you'll also need to edit that (then regenerate to make sure that webgpu.h is still synced - the github action will verify it).

> For some reason all this values differs in wgpu:

We haven't brought them all in sync yet, but we'll get there once the header is stabilized.

PR #327, opened 2024-09-12
Add usage field to WGPUTextureViewDescriptor

vonture:
Adds the usage field to WGPUTextureViewDescriptor for the WebGPU spec change  made in https://github.com/gpuweb/gpuweb/pull/4746


kainino0x (approved):
Thanks!

PR #336, opened 2024-09-17
Replace bitmask enums with uint64_t + constants, and remove "Flags" suffix

### LINKED ISSUE #247: Expand some bitmask enums to 64-bits?
Vulkan has had several bitmask enums that they had to expand from 32 bits to 64 bits (by adding version "2" of a bunch of things throughout the API like [VkPipelineStageFlags2](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPipelineStageFlags2.html), vkCmdWriteTimestamp2). We should probably do this early to avoid that kind of thing.

We have 5 of these (in order of likelihood to need more than ~32~ 31? bits ever):
WGPUBufferUsage
WGPUTextureUsage
WGPUMapMode
WGPUShaderStage (almost certainly not)
WGPUColorWriteMask (definitely not)

Note that with #159 we're planning to remove the Flags typedefs, so all we would do here is change these enums' Force32 value to Force64.

---

- [ ] Probably document that native-only bits should be allocated in the 10-11 MSBs (because JS can only use 53 bits anyway). (Note: maybe don't use `2**63` if we use C enums if they're not guaranteed to be unsigned?)

### LINKED ISSUE #273: Defining enums and bitmasks using `enum` vs. as constants
_Originally posted by @kainino0x in https://github.com/webgpu-native/webgpu-headers/issues/270#issuecomment-1947635248_

> Currently extended bitflag entries are similar to enums, but I think we should change generation of all bitflag entries to be `static const T N = V` instead of enum, because we need to make them 64bit anyway. [Vulkan does the same for 64bit bitflags](https://github.com/KhronosGroup/Vulkan-Headers/blob/5ac36269f50381bdd92a5e1973d8eb041771e59e/include/vulkan/vulkan_core.h#L6538)

Ah... unfortunate. I dug a little to figure out why this was the case in Vulkan. Clang makes it clear - and this is also the reason Force32 is 0x7fffffff instead of 0xffffffff.

`warning: ISO C restricts enumerator values to range of 'int' (4294967295 is too large) [-Wpedantic]`

C++ seems to have no such complaints for `enum class`, so Dawn should still be able to do the thing we do in C++ to make strongly-typed bitflags from `enum class`.
C++ _also_ seems to have no complains for `enum`, which means we _could_ use `enum` behind `#ifdef __cplusplus`, for added type safety. Doing so would be consistent with #159, where we were going to do the `operator|` thing to make C++ happy with our usage of `enum`s as bitflags. But having separate C and C++ definitions of entire enums is annoying, and also sketchy - it almost certainly breaks ABI compatibility between the C and C++ versions.
So I think we are going to have to scrap the `operator|` thing and just use 64-bit constants. Too bad.

> the downside being we loose exhaustive switches in C++ over new entries

Actually this raises another interesting point - if we extend enums outside the enum definition, then it becomes easy to write a switch that passes the exhaustiveness check, but is not actually exhaustive. Of course, this was always true of bitflags, and generally true if you cast numbers to enums, but those are corner cases.

This is probably fine, it just adds a little bit of a false sense of security. Almost certainly better to get some warnings than no warnings.
Most enums won't be commonly in application code anyway. The enums we do pass out to the application may get extended, but for *most* enums, we should never start returning new values to existing applications unless we enable some feature, so exhaustivity over the core enum is all that really matters.
For reference, I think the only enums where we may suddenly start passing out new values to the application are `FeatureName` (in AdapterEnumerateFeatures, but not DeviceEnumerateFeatures) and `WGSLFeatureName` (in GetInstanceFeatures or whatever we end up calling it). Not `SType`, I think, if we solve #272.

Maybe... after we stabilize, we should just never change the enum definitions and *only* extend them using `static const`? (maybe even rename Force32 to indicate it's there to make the enum extensible/non-exhaustive, as anyone writing exhaustive switches will end up having to type Force32)

(Aside, apparently there is a gcc/clang flag that checks for exhaustive switches even *with* default. I am now tempted to try this in Dawn. https://brevzin.github.io/c++/2019/08/01/enums-default/#the-warning-that-exists)

kainino0x:
Fixes #247
Fixes #273
(already done in Dawn)

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -1106,7 +1106,7 @@ typedef struct WGPUBufferBindingLayout {
 typedef struct WGPUBufferDescriptor {
     WGPUChainedStruct const * nextInChain;
     WGPU_NULLABLE char const * label;
-    WGPUBufferUsageFlags usage;
+    WGPUBufferUsage usage;
     uint64_t size;
     WGPUBool mappedAtCreation;
 } WGPUBufferDescriptor WGPU_STRUCTURE_ATTRIBUTE;
@@ -1366,7 +1366,7 @@ typedef struct WGPUSurfaceCapabilities {
      * The bit set of supported @ref WGPUTextureUsage bits.
      * Guaranteed to contain @ref WGPUTextureUsage_RenderAttachment.
      */
-    WGPUTextureUsageFlags usages;
+    WGPUTextureUsage usages;
     /**
      * A list of supported @ref WGPUTextureFormat values, in order of preference.
      */
@@ -1403,7 +1403,7 @@ typedef struct WGPUSurfaceConfiguration {
     /**
      * The @ref WGPUTextureUsage of the surface's textures.
      */
-    WGPUTextureUsageFlags usage;
+    WGPUTextureUsage usage;
     /**
      * The width of the surface's textures.
      */
@@ -1566,7 +1566,7 @@ typedef struct WGPUTextureViewDescriptor {
     uint32_t baseArrayLayer;
     uint32_t arrayLayerCount;
     WGPUTextureAspect aspect;
-    WGPUTextureUsageFlags usage;
+    WGPUTextureUsage usage;
 } WGPUTextureViewDescriptor WGPU_STRUCTURE_ATTRIBUTE;
 
 typedef struct WGPUVertexAttribute {
@@ -1586,7 +1586,7 @@ typedef struct WGPUBindGroupDescriptor {
 typedef struct WGPUBindGroupLayoutEntry {
     WGPUChainedStruct const * nextInChain;
     uint32_t binding;
-    WGPUShaderStageFlags visibility;
+    WGPUShaderStage visibility;
     WGPUBufferBindingLayout buffer;
     WGPUSamplerBindingLayout sampler;
     WGPUTextureBindingLayout texture;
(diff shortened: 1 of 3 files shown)


lokokung (approved):

webgpu.h
```diff
@@ -899,60 +899,45 @@ typedef enum WGPUWaitStatus {
...
+static const WGPUBufferUsage WGPUBufferUsage_Indirect = 0x0000000000000100;
+static const WGPUBufferUsage WGPUBufferUsage_QueryResolve = 0x0000000000000200;
+
+typedef WGPUFlags WGPUColorWriteMask;
+static const WGPUColorWriteMask WGPUColorWriteMask_None = 0x0000000000000000;
+static const WGPUColorWriteMask WGPUColorWriteMask_Red = 0x0000000000000001;
+static const WGPUColorWriteMask WGPUColorWriteMask_Green = 0x0000000000000002;
+static const WGPUColorWriteMask WGPUColorWriteMask_Blue = 0x0000000000000004;
+static const WGPUColorWriteMask WGPUColorWriteMask_Alpha = 0x0000000000000008;
+static const WGPUColorWriteMask WGPUColorWriteMask_All = WGPUColorWriteMask_None | WGPUColorWriteMask_Red | WGPUColorWriteMask_Green | WGPUColorWriteMask_Blue | WGPUColorWriteMask_Alpha;
```
PJB3005, line 921: This seems to give me the following error when compiling with MSVC:

```
E:\Projects\wgpu-native\ffi\webgpu-headers\webgpu.h(931,33): error C2099: initializer is not a constant [E:\Projects\wgpu-native\examples\build\framework\framework.vcxproj]
```

Apparently const ints like this are not constants themselves in C... https://stackoverflow.com/questions/6131455/compile-error-c2099-initializer-is-not-a-constant

PR #338, opened 2024-09-17
Move .suboptimal to .status=SuccessSuboptimal

### LINKED ISSUE #257: Consider renaming SwapChainSurface.suboptimal to optimal.
A slight nit is that negative member names often end up creating double negations in code that are slightly harder to parse. Renaming `suboptimal` to `optimal` would help.

kainino0x:
Fixes #257
dawn issue https://crbug.com/367401498

NOT stable yet, this is just one change to the presentation API, more coming: https://github.com/webgpu-native/webgpu-headers/labels/presentation


Kangz:
I'm kind of lukewarm on this change because [I just fixed an issue related to this in Dawn's Vulkan backend](https://dawn-review.googlesource.com/c/dawn/+/205035). Though the names suggest that there are more options than just `SuccessOptimal`.

Kangz (approved):

kainino0x:
I was worried about that exact scenario too, it's interesting to see that it actually happened in Dawn...

I am not super attached to this change, but yeah I think that `SuccessOptimal` will prevent bugs like that one (and the fact that we use a custom result enum here).

Issue #190, opened 2023-06-03
Nullability attributes need some love


litherum:
I had to make quite a bit of changes to make the nullability attributes actually compile.

- [ ] The attribute is in the wrong place. Right: `char const * WGPU_NULLABLE label;` Wrong: `WGPU_NULLABLE char const * label;`
- [ ] You probably want to add a place to stick `_Pragma("clang assume_nonnull begin")` and `_Pragma("clang assume_nonnull end")` so you don't have to annotate literally everything
- [ ] Even when using `assume_nonnull`, there are still some places that need an explicit `WGPU_NONNULL`. I don't quite understand why. The [docs](https://developer.apple.com/swift/blog/?id=25) say "More complex pointer types ... must be explicitly annotated."
- [ ] `wgpuAdapterEnumerateFeatures()` and `wgpuDeviceEnumerateFeatures()` need to have their argument be nullable. These functions are designed to be called twice: once to let the caller know how much storage to allocate, and then a second time with the actual store. That first call needs to support nullptr.

On the upside, though, after getting these attributes to compile, they found 2 real bugs in our implementation!

kainino0x:
> * [ ] You probably want to add a place to stick `_Pragma("clang assume_nonnull begin")` and `_Pragma("clang assume_nonnull end")` so you don't have to annotate literally everything
> * [ ]  Even when using `assume_nonnull`, there are still some places that need an explicit `WGPU_NONNULL`. I don't quite understand why. The [docs](https://developer.apple.com/swift/blog/?id=25) say "More complex pointer types ... must be explicitly annotated."

Since we're generating the header anyway I kind of prefer annotating everything. It's noisy but I think it's better than having to go learn the rules about exactly what is and isn't affected by assume_nonnull...

kainino0x:
Nullability attributes are broken so they can't be used at all right now. It is quite a bit of work to get them fully working so I would like to declare this `non-blocking`, can be fixed after 1.0.

kainino0x:
Dec 12 meeting:

* Fixing `WGPU_NULLABLE`, introducing `WGPU_NON_NULL`, and also fixing some `WGPU_*_ATTRIBUTE`s. Postpone to after 1.0?  
* KN: The attributes in particular will require someone to actually write obj-c bindings to make sure that they're working correctly. But nullability also not very important for c/c++ users.  
* **OK**

kainino0x:
[...]

Issue #170, opened 2022-12-06
Handling strings containing \0


austinEng:
Currently, all `const char*` members and arguments are null-terminated strings.

This doesn't work well for strings that contain the \0 character which the WebGPU JavaScript API allows. The C API should have a way to express such strings.

One suggestion for doing this would be to add an extra member/argument for every string that is an explicit length.

It could be optional, such that zero means `strlen` and non-zero is an explicit length. However, this would make it impossible to express non-null strings that are zero-length. Perhaps that's ok? If not, the alternative can be to use SIZE_MAX to mean "use `strlen`".

Kangz:
@cwfitzgerald @teoxoy WDYT?

IMHO having zero mean strlen works, and if someone really wants to represent the empty string with an explicit length faithfully, then they can change the pointer to point to `"\0"`.

cwfitzgerald:
I'm obviously a little biased, but I don't really see the point of zero-deliminated strings when we can provide size, and as zero is a perfectly valid size (i.e. do not dereference this pointer), u32::MAX seems like the most reasonable value of "this is zero-deliminated, figure it out yourself".

Kangz:
For C, and this is a C header, 0 initialisation is very convenient whereas typing a `WGPU_STRLEN_LENGTH` all the time would be very annoying. (it would be `SIZE_T_MAX`). And strings tend to be null-terminated ones. But this is also a little bit biased :) On a scale of 1 to 10 I think I care at around 3. How about you? (basically it would be a wart but somewhat manageable for developers)

cwfitzgerald:
As an implementer, about 1 (it will go in a string parsing function which can do whatever) - as a user I have no authority to speak :)

austinEng:
I also don't care so much. I've realized though, we do have `#define WGPU_WHOLE_MAP_SIZE SIZE_MAX`.
perhaps we should also use SIZE_MAX to mean `strlen` here for consistency?
All of the "undefined" constants are also the largest representable integer:
```
#define WGPU_ARRAY_LAYER_COUNT_UNDEFINED (0xffffffffUL)
#define WGPU_COPY_STRIDE_UNDEFINED (0xffffffffUL)
#define WGPU_LIMIT_U32_UNDEFINED (0xffffffffUL)
#define WGPU_LIMIT_U64_UNDEFINED (0xffffffffffffffffULL)
#define WGPU_MIP_LEVEL_COUNT_UNDEFINED (0xffffffffUL)
[...]

PR #352, opened 2024-09-24
Add WGPU_MAKE_INIT_STRUCT and WGPUStringView

kainino0x:
This is basically copied from Dawn except:
- Use `NULL` instead of `nullptr` (`nullptr` was added in C23)
- Put `WGPU_NULLABLE` in the correct place per #190 so we don't forget about it later
- Documentation! I added a few important things we did not discuss:
  - Places that accept non-nullable stringviews simply default to empty string (e.g. labels)
  - Outputs from the API always use explicit-length stringviews which may or may not be null terminated

Issue: #170 (not fixed yet, this is just the start)
(dawn bug https://crbug.com/42241188)

### DIFF
--- a/gen/cheader.tmpl
+++ b/gen/cheader.tmpl
@@ -65,6 +65,21 @@
 #include "webgpu.h"
 {{end}}
 
+{{- if eq .Name "webgpu"}}
+#define WGPU_COMMA ,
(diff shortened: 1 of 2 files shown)


Kangz (approved):
LGTM!

PR #353, opened 2024-09-24
Move __WGPU_EXTEND_ENUM out of the compiler-reserved namespace

kainino0x:
I think libraries are supposed to avoid macro names starting with underscores - see https://gcc.gnu.org/onlinedocs/cpp/System-specific-Predefined-Macros.html

We could alternatively come up with some other valid naming convention for internal things like `_wgpu_EXTEND_ENUM`, WDYT?

### DIFF
--- a/gen/cheader.tmpl
+++ b/gen/cheader.tmpl
@@ -49,14 +49,14 @@
 #endif{{"\n" -}}
 
 {{- if ne .Name "webgpu"}}
-#if !defined(__WGPU_EXTEND_ENUM)
+#if !defined(WGPU_EXTEND_ENUM)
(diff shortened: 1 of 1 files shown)


kainino0x:
@rajveermalviya PTAL

rajveermalviya:
+1 for `_wgpu_EXTEND_ENUM` (or if possible `_WGPU_EXTEND_ENUM`), as it's only intended for internal use.

kainino0x:
Filed #355 since whatever we do we need to do for other internal macros too. Will close this for now and open a new PR that does all of them when we know what we want to do.

PR #357, opened 2024-09-26
Add missing WGPU_STRLEN and document some more string things

kainino0x:
I should note that `WGPU_STRLEN` is added directly to the template, like `WGPUStringView` and `WGPU_STRING_VIEW_INIT`, because I figure these won't be exposed directly by bindings to most other languages, since most languages have their own string/string view types.

@rajveermalviya Could you also take a moment to review the semantics I previously documented for WGPUStringView? In particular, `Those which do not accept null values "default" to the empty string when null values are passed.` (Sorry I didn't cc you when I was adding it.)

### DIFF
--- a/gen/cheader.tmpl
+++ b/gen/cheader.tmpl
@@ -252,7 +252,8 @@ typedef struct WGPUChainedStructOut {
  * - `{non_null_pointer, WGPU_STRLEN}`: a null-terminated string view.
  * - `{any, 0}`: the empty string.
  * - `{NULL, non_zero_length}`: not allowed (null dereference).
- * - `{non_null_pointer, non_zero_length}`: an explictly-sized string view.
+ * - `{non_null_pointer, non_zero_length}`: an explictly-sized string view with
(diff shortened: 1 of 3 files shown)


gen/cheader.tmpl
```diff
@@ -262,6 +263,13 @@ typedef struct WGPUStringView {
     size_t length;
 } WGPUStringView;
 
+/**
+ * Sentinel value used in @link WGPUStringView to indicate that the pointer
+ * is to a null-terminated string, rather than an explicitly-sized string.
+ */
+#define WGPU_STRLEN
```
rajveermalviya, line 270: This should be `#define WGPU_STRLEN SIZE_MAX` ?

rajveermalviya:
> the semantics I previously documented for WGPUStringView

Yeah those LGTM

gen/cheader.tmpl
kainino0x: Oops ^_^ thank you!

rajveermalviya (approved):

Issue #339, opened 2024-09-19
Latest header does not compile on MSVC due to const expressions


PJB3005:
The following line doesn't compile on MSVC anymore, it seems:

https://github.com/webgpu-native/webgpu-headers/blob/2b5974750be7b2545d842ecec02e585022081952/webgpu.h#L931

```
E:\Projects\wgpu-native\ffi\webgpu-headers\webgpu.h(931,33): error C2099: initializer is not a constant [E:\Projects\wgpu-native\examples\build\framework\framework.vcxproj]
```

According to this very authorative source (stack overflow), `const int` declarations are not const themselves in C: https://stackoverflow.com/questions/6131455/compile-error-c2099-initializer-is-not-a-constant, so that's probably why this doesn't work.

Found this while trying to update wgpu-native to the latest headers.

kainino0x:
It's surprising that even the latest MSVC (checked on godbolt.org) doesn't support this yet, clang/gcc can't catch it (https://github.com/webgpu-native/webgpu-headers/blob/main/tests/compile/Makefile).
Thanks for reporting!

PR #359, opened 2024-09-29
Replace constant defines by static const typed variables

eliemichel:
This may be a naive question, but is there any reason for these constant values to be declared as macro variables using `#define` rather than typed static consts?

We already use static const for enum values now so there should be no support issue.


eliemichel:
(Will update webgpu.yml and gen script if relevant)

kainino0x:
None of us had strong opinions on which way to go, but it's a bit more traditional. It feels too arbitrary to be worth changing, but I don't know.

- for example gl.h and vulkan_core.h do it this way
- there was some preference for macros expressed here: https://github.com/webgpu-native/webgpu-headers/pull/34
- it also turns out there are some limitations with `static const` that might push us toward macros: https://github.com/webgpu-native/webgpu-headers/issues/339 (though I _don't think_ we'll use macros to define all of the flags constants - incidentally Vulkan also uses `static const` for 64-bit flags).

kainino0x:
> * it also turns out there are some limitations with `static const` that might push us toward macros: [Latest header does not compile on MSVC due to const expressions #339](https://github.com/webgpu-native/webgpu-headers/issues/339) (though I _don't think_ we'll use macros to define all of the flags constants - incidentally Vulkan also uses `static const` for 64-bit flags).

I [checked](https://godbolt.org/z/GW5PjdTWK), and this applies to MSVC C but not C++. The pattern this will create problems for is e.g.:
```c
static const uint64_t kMyUsages = WGPUBufferUsage_CopyDst | WGPUBufferUsage_Uniform;
```
but users can deal with this, by themselves using a macro:
```c
#define MY_USAGES (WGPUBufferUsage_CopyDst | WGPUBufferUsage_Uniform)
```

kainino0x:
I found a useful series of answers about C here: https://stackoverflow.com/questions/1674032/static-const-vs-define-vs-enum
(which was harder to find than the equivalent C++ question here: https://stackoverflow.com/questions/1637332/static-const-vs-define)

Given we probably don't get any significant type safety out of these (they're mostly uint32_t and they're no worse than `UINT32_MAX` etc.) I would lean weakly toward just keeping the `#defines`.

kainino0x:
Dec 12 meeting:

* This is for the few sentinel values at the top of the file.  
* KN: Only concrete difference is that you can't make static consts from static consts. But that's not a big deal.  
* ?: Avoids possibly taking up memory on bad compilers.
* LK: Seems fine to keep \#defines  
* CF: Can just let fancier languages use fancier constants.  
* **Leave as is**

PR #362, opened 2024-10-02
Add GPUSurfaceConfiguration.colorSpace

### LINKED ISSUE #200: Color spaces for presentation
Split out from #197. See https://github.com/webgpu-native/webgpu-headers/issues/197#issuecomment-1644630551
where we decided we should land other stuff eagerly before figuring this part out for WGPUSurfaceConfiguration.

kainino0x:
bikeshedding:
- naming of WGPUColorSpace (WGPUPredefinedColorSpace? note we are likely to use it in non-surface places for copyExternalImageToTexture/importExternalTexture)
- capitalization of WGPUColorSpace_SRGB (Srgb? sRGB?)

Fixes #200

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -366,6 +366,23 @@ typedef enum WGPUCallbackMode {
     WGPUCallbackMode_Force32 = 0x7FFFFFFF
 } WGPUCallbackMode WGPU_ENUM_ATTRIBUTE;
 
+/**
+ * Predefined sets of color space parameters.
+ */
+typedef enum WGPUColorSpace {
+    /**
+     * `0x00000000`.
+     * The color space defined by [`srgb`](https://www.w3.org/TR/css-color-4/#predefined-sRGB) in CSS.
+     */
+    WGPUColorSpace_SRGB = 0x00000000,
+    /**
+     * `0x00000001`.
+     * The color space defined by [`display-p3`](https://www.w3.org/TR/css-color-4/#predefined-display-p3) in CSS.
+     */
+    WGPUColorSpace_DisplayP3 = 0x00000001,
+    WGPUColorSpace_Force32 = 0x7FFFFFFF
+} WGPUColorSpace WGPU_ENUM_ATTRIBUTE;
+
 typedef enum WGPUCompareFunction {
     WGPUCompareFunction_Undefined = 0x00000000,
     WGPUCompareFunction_Never = 0x00000001,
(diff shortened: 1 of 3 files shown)


webgpu.yml
```diff
@@ -219,6 +219,15 @@ enums:
             Implementations _should_ fire spontaneous callbacks as soon as possible.
 
           @note Because spontaneous callbacks may fire at an arbitrary time on an arbitrary thread, applications should take extra care when acquiring locks or mutating state inside the callback. It undefined behavior to re-entrantly call into the webgpu.h API if the callback fires while inside the callstack of another webgpu.h function that is not `wgpuInstanceWaitAny` or `wgpuInstanceProcessEvents`.
+  - name: color_space
+    doc: Predefined sets of color space parameters.
+    entries:
+      - name: undefined
+        doc: Use the default color space.
```
kainino0x, line 226: note to self: write an article on Undefined and link to it.

webgpu.h
```diff
@@ -1539,6 +1561,14 @@ typedef struct WGPUSurfaceConfiguration {
...
      */
     WGPUPresentMode presentMode;
+    /**
+     * Color space to display the surface with.
+     *
+     * The only value that implementations are required to support is @ref WGPUColorSpace_Undefined,
+     * which may be either "unmanaged" or "sRGB".
+     * Wasm implementations should support @ref WGPUColorSpace_SRGB and @ref WGPUColorSpace_DisplayP3.
+     */
+    WGPUColorSpace colorSpace;
```
Kangz, line 1571: Should we have the surface capabilities reflect the available color spaces?

kainino0x: We talked about this in https://github.com/webgpu-native/webgpu-headers/issues/200#issuecomment-1878100315 but it seemed complex in practice and not necessary yet. On the web we currently just support all three formats with all two color spaces, but in native I don't know if all of those combinations are possible, and certainly new color spaces might not be compatible with all formats or vice versa.

kainino0x: Approve then? :)
(docs updated)

kainino0x: filed #366 for both colorspace and tonemapping capabilities

eliemichel:
So, what happens if we specify a `format` that is not Srgb but a `colorSpace` set to Srgb or the other way around? Is this a way to have the implementation apply the colorSpace change when presenting rather than manually specifying it in the fragment shaders? That's nice if it is so, maybe the following doc (which may be wrong the way I phrase it) to clarify:

```
      - name: color_space
        doc: |
          Color space transform to apply to the presented textures before displaying them on the surface.
          A value of @ref WGPUColorSpace_SRGB means that pixel values are transformed from linear to [`srgb`](https://www.w3.org/TR/css-color-4/#predefined-sRGB).
          A value of @ref WGPUColorSpace_DisplayP3 means that pixel values are transformed from linear to [`display-p3`](https://www.w3.org/TR/css-color-4/#predefined-display-p3).

          The only value that implementations are required to support is @ref WGPUColorSpace_Undefined,
          which may be either "unmanaged" or "sRGB" in native (translates to the default `"srgb"` in Wasm).
          Wasm implementations should support @ref WGPUColorSpace_SRGB and @ref WGPUColorSpace_DisplayP3.
```

kainino0x:
@eliemichel The two concepts are orthogonal though they can be used together. Please check out this draft I opened a while ago and provide any feedback you might have on it:
https://github.com/gpuweb/gpuweb/pull/4597/files

kainino0x:
Closing this PR, I think we are moving color space and tone mapping to Emscripten extensions.

PR #363, opened 2024-10-03
Add surface tone mapping mode

kainino0x:
Proposal.

This is in the JS spec, so like #200, we should probably add it here too, though I don't know if native implementations will need more than this for HDR. We could move this to a Wasm-specific extension.

I don't know anything about what this looks like in native so hopefully others do.

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -730,6 +730,32 @@ typedef enum WGPUSurfaceGetCurrentTextureStatus {
     WGPUSurfaceGetCurrentTextureStatus_Force32 = 0x7FFFFFFF
 } WGPUSurfaceGetCurrentTextureStatus WGPU_ENUM_ATTRIBUTE;
 
+/**
+ * Specifies how color values are displayed to the screen.
+ */
+typedef enum WGPUSurfaceToneMappingMode {
+    /**
+     * `0x00000000`.
+     * Use the default tone mapping mode (@ref WGPUSurfaceToneMappingMode_Standard).
+     */
+    WGPUSurfaceToneMappingMode_Undefined = 0x00000000,
+    /**
+     * `0x00000001`.
+     * Color values within the standard dynamic range of the screen are
+     * unchanged, and all other color values are projected to the standard
+     * dynamic range of the screen.
+     */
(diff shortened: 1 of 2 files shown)


webgpu.h
```diff
@@ -730,6 +730,32 @@ typedef enum WGPUSurfaceGetCurrentTextureStatus {
     WGPUSurfaceGetCurrentTextureStatus_Force32 = 0x7FFFFFFF
 } WGPUSurfaceGetCurrentTextureStatus WGPU_ENUM_ATTRIBUTE;
 
+/**
+ * Specifies how color values are displayed to the screen.
+ */
+typedef enum WGPUSurfaceToneMappingMode {
+    /**
+     * `0x00000000`.
+     * Use the default tone mapping mode (@ref WGPUSurfaceToneMappingMode_Standard).
```
kainino0x, line 739: note to self: write an article on Undefined and link to it.

webgpu.h
```diff
@@ -1539,6 +1565,12 @@ typedef struct WGPUSurfaceConfiguration {
      * When and in which order the surface's frames will be shown on the screen. Defaults to @ref WGPUPresentMode_Fifo.
      */
     WGPUPresentMode presentMode;
+    /**
+     * Specifies how color values are displayed to the screen.
+     *
+     * Implementation support is not required for @ref WGPUSurfaceToneMappping_Extended.
+     */
+    WGPUSurfaceToneMappingMode toneMappingMode;
```
Kangz, line 1573: ditto, should we expose this in surface capabilities?

kainino0x: I think this has the same issue as #362 though the interaction is a bit less complex, it's pretty likely that, if the system can do HDR for any format, then all formats that support values>1 (so, float formats, and the hypothetical [XR formats](https://github.com/gpuweb/gpuweb/issues/4108)) would support it.

I'll open a separate issue to track it, and we can start with this.

kainino0x: #366

kainino0x:
Closing this PR, I think we are moving color space and tone mapping to Emscripten extensions.

PR #364, opened 2024-10-03
Update all sentinel values, reserve 0

### LINKED ISSUE #234: Naming of sentinel values not present in JS
In the C API we have a few sentinel enum values that don't have equivalents in JS:

- `WGPUVertexStepMode_VertexBufferNotUsed` (JS has `sequence<GPUVertexBufferLayout?>`) (see #242)
- `WGPUTextureFormat_Undefined` in `WGPUFragmentState` (#160) (JS has `sequence<GPUColorTargetState?>`)
- The binding types: BufferBindingType, SamplerBindingType, TextureSampleType, StorageTextureAccess `_Undefined` (JS has the parent dictionary optional) (see #242)
- `WGPUCompositeAlphaMode_Auto`

We also have `Undefined` variants for a bunch of enums which are used as optional-without-default in JS.

And finally, constants for sentinel values of numeric fields:

```c
#define WGPU_ARRAY_LAYER_COUNT_UNDEFINED (0xffffffffUL)
#define WGPU_COPY_STRIDE_UNDEFINED (0xffffffffUL)
#define WGPU_LIMIT_U32_UNDEFINED (0xffffffffUL)
#define WGPU_LIMIT_U64_UNDEFINED (0xffffffffffffffffULL)
#define WGPU_MIP_LEVEL_COUNT_UNDEFINED (0xffffffffUL)
#define WGPU_QUERY_SET_INDEX_UNDEFINED (0xffffffffUL)
#define WGPU_WHOLE_MAP_SIZE SIZE_MAX
#define WGPU_WHOLE_SIZE (0xffffffffffffffffULL)
```

Should we call all of them "undefined"?
Or should we be trying to use more meaningful names, like `VertexBufferNotUsed` and `WHOLE_SIZE`?
Or should we have both?

Examples of meaningful names:
- `WGPUTextureFormat_ColorTargetNotUsed`
- `WGPUBufferBindingType_NotUsed`
- `WGPU_ARRAY_LAYER_COUNT_DEFAULT` or e.g. `TO_END` or `ALL_REMAINING`
- `WGPU_QUERY_SET_INDEX_NOT_WRITTEN`

### LINKED ISSUE #242: Some enums' sentinel 0 values are indistinguishable from "undefined" defaulting
Right now in webgpu.h, `WGPUVertexStepMode_VertexBufferNotUsed` has value `0`, which we use as a sentinel value (#234) to indicate a hole in the array, where in JS the hole would be represented by a `null` in `sequence<GPUVertexBufferLayout?>`.

This means we don't have a proper `WGPUVertexStepMode_Undefined`, which according to the translation rules we use elsewhere would mean "use the default" (which for `GPUVertexBufferLayout` is `"vertex"`).

kainino0x:
**Can be reviewed as separate commits.**

- Reorder fields in VertexBufferLayout so the sentinel slot is first (even though this doesn't match the upstream IDL)
- Update `BindingNotUsed`/`VertexBufferNotUsed` enum values
- Removing some `Undefined`s that don't need to exist
- Reserved 0 in all enums without `Undefined`
- Ordered some enums to match the spec

This is a bit different from the thing we previously discussed and agreed on:

- I've swapped the values of `BindingNotUsed`=`VertexBufferNotUsed`=0 with `Undefined`=1 so that zero-init will trigger `NotUsed`. See https://github.com/webgpu-native/webgpu-headers/issues/242#issuecomment-2310857163
- I haven't introduced `WGPUTextureFormat_ColorTargetNotUsed`, opting to continue using `WGPUTextureFormat_Undefined` for this, because there's basically no chance that `WGPUColorTargetState.format` becomes optional. (Even if it did become "dynamic state" or something, I think we would use a different sentinel value than `undefined` in the JS API so that it had to be named explicitly.)

Fixes #234
Fixes #242
(done in Dawn except for the NotUsed changes; TODO file a bug for that)

### DIFF
--- a/webgpu.yml
+++ b/webgpu.yml
@@ -77,8 +77,7 @@ enums:
       TODO
     entries:
       - name: undefined
-        doc: |
-          TODO
+        doc: No value. See @ref SentinelValues.
       - name: "null"
         doc: |
           TODO
@@ -224,8 +223,7 @@ enums:
       TODO
     entries:
       - name: undefined
-        doc: |
-          TODO
+        doc: No value. See @ref SentinelValues.
       - name: never
         doc: |
           TODO
@@ -471,8 +469,7 @@ enums:
       TODO
     entries:
       - name: undefined
-        doc: |
-          TODO
+        doc: No value. See @ref SentinelValues.
       - name: uint16
         doc: |
           TODO
@@ -484,8 +481,7 @@ enums:
       TODO
     entries:
       - name: undefined
-        doc: |
-          TODO
+        doc: No value. See @ref SentinelValues.
       - name: load
         doc: |
           TODO
@@ -556,8 +552,7 @@ enums:
       TODO
     entries:
       - name: undefined
-        doc: |
-          TODO
+        doc: No preference. (See also @ref SentinelValues.)
       - name: low_power
         doc: |
           TODO
@@ -795,8 +790,7 @@ enums:
       TODO
     entries:
       - name: undefined
-        doc: |
-          TODO
+        doc: No value. See @ref SentinelValues.
       - name: store
         doc: |
           TODO
@@ -851,8 +845,7 @@ enums:
       TODO
     entries:
       - name: undefined
-        doc: |
-          TODO
+        doc: No value. See @ref SentinelValues.
       - name: R8_unorm
         doc: |
           TODO
@@ -1165,8 +1158,7 @@ enums:
       TODO
     entries:
       - name: undefined
-        doc: |
-          TODO
+        doc: No value. See @ref SentinelValues.
       - name: 1D
         doc: |
           TODO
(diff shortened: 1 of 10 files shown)


kainino0x:
@cwfitzgerald @rajveermalviya @Kangz see above about the things I did differently than what we discussed (no need to review fully but feel free). Let me know if you think any of those changes are not good.

cwfitzgerald:
No notes here

webgpu.yml
```diff
@@ -304,38 +309,34 @@ enums:
...
       - name: unknown
-        value: 0x0007
         doc: |
           TODO
   - name: cull_mode
     doc: |
       TODO
     entries:
+      - name: undefined
+        doc: No value. See @ref SentinelValues.
```
Kangz, line 339:
```suggestion
        doc: The absence of a value. See @ref SentinelValues.
```

I got mildly confused when reading the C header as I though the block comment meant there should not have been an enum value associated with `0x000000`

Kangz (approved):

Issue #326, opened 2024-09-04
out-structs containing objects and refcounting


Kangz:
We should define what happens to objects contained in out-structs that will be overwritten. I had the following code and it was leaking quite unexpectedly:

```c
WGPUSurfaceTexture s;

wgpuSurfaceGetCurrentTexture(&s);
wgpuSurfacePresent();
wgpuSurfaceGetCurrentTexture(&s);

wgpuSurfaceTextureFreeMembers(&s);
```

The issue is that the second `wgpuSurfaceGetCurrentTexture` call overwrites `s.texture` without decrementing the refcount, which means that `wgpuSurfaceTextureFreeMembers` doesn't know to unref it. (or maybe it is a C++ism of Dawn that we need to fix, but in all cases the behavior is surprising).

In all cases I'd suggest that if an object is passed in an out-struct, it must be either null or valid, and if valid it will be released once.

kainino0x:
`wgpuSurfaceTextureFreeMembers` doesn't exist (I got confused about this at some point) because there are no arrays of data to free.

Generally speaking, just because there's a pointer value for a WGPUTexture in WGPUSurfaceTexture doesn't mean that's a valid reference. It could have been Released already and the app didn't zero out the pointer because they knew they were about to overwrite it:

```c
wgpuTextureRelease(s.texture);
wgpuSurfaceGetCurrentTexture(&s); // double free
```

I think it's the reality of the C API that we don't know what the intent was so we should just overwrite the pointer without releasing it. It's unfortunate that the overwrite is hard to see (vs. if we just did `texture = wgpuSurfaceGetCurrentTexture();` which is more obviously a leak) but I think it's fine.

In our C++ API, we _do_ know whether the ref is still there because the RAII wrapper already needs to know. If we implemented the C++ API in C++ that would happen automatically, because we'd write `s->texture = nextTexture;` and that would release the existing ref. I think it's just because of our C/C++ reinterpret_casting that we don't currently do that, so we would need to do `s->texture = {};` before calling from the C++ wrapper into the C API.

kainino0x:
Sep 12 meeting:

- (brief discussion, no really strong opinions)
- **Sounds good to leak the ref in this case (rather than releasing)**

kainino0x:
Closed, still needs docs.

Issue #143, opened 2022-01-07
Ownership of returned strings


kvark:
Related to #82 

We had a small discussion about `wgpuAdapterGetProperties`. It needs to return a name and some description. When are these going to be freed?

It seems to me that we should make the storage totally owned by the user. So they'd provide a pointer and the capacity, and we'd just fill out the storage with some chars.

kainino0x:
I would probably just have them owned by the WGPUAdapter. Though, if doing that, the whole WGPUAdapterProperties could just be owned by the WGPUAdapter and you wouldn't need this writable pointer thing at all.

kvark:
Ok, right, that's an option. It's just not super convenient for us, because it's a C string, and our adapters don't like holding onto those kind.

kainino0x:
We discussed a proposal in today's meeting and I think we were pretty happy with it:

```c
typedef struct WGPUAdapterPropertiesImpl* WGPUAdapterProperties;

void wgpuAdapterPropertiesReference(WGPUAdapterProperties);
void wgpuAdapterPropertiesRelease(WGPUAdapterProperties);

WGPUAdapterProperties wgpuAdapterGetProperties(WGPUAdapter);

typedef struct WGPUAdapterPropertiesV1 {
    // (no chained struct here)
    uint32_t vendorID;
    char const * vendorName;
    char const * architecture;
    uint32_t deviceID;
    char const * name;
    char const * driverDescription;
    WGPUAdapterType adapterType;
    WGPUBackendType backendType;
} WGPUAdapterPropertiesV1;

typedef struct WGPUAdapterPropertiesDAWN {
    // example extension
} WGPUAdapterPropertiesDAWN;

// The structs are owned by AdapterProperties, returned by reference, freed when it's freed
WGPUAdapterPropertiesV1 const * wgpuAdapterPropertiesGetV1(WGPUAdapterProperties);
WGPUAdapterPropertiesDAWN const * wgpuAdapterPropertiesGetDAWN(WGPUAdapterProperties);
// Extensions can also return things other than structs
int wgpuAdapterPropertiesGetFavoriteNumber(WGPUAdapterProperties);
WGPUSurface
```

[...]

PR #371, opened 2024-10-16
Add docs on ownership

kainino0x:
Also generate docs for function parameters.

I think these are the last of the docs for #143 and #326. I probably missed a few places that need ownership docs but this should be most of them.

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -109,7 +109,7 @@
                     "type": "array",
                     "description": "Optional property, list of callback arguments",
                     "items": {
-                        "$ref": "#/definitions/ParameterType"
+                        "$ref": "#/definitions/FunctionParameterType"
                     }
                 }
             },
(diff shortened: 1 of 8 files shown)


schema.json
```diff
@@ -133,6 +133,14 @@
                     "$ref": "#/definitions/Type",
                     "description": "Parameter type"
                 },
+                "ownership": {
+                    "type": "string",
+                    "description": "Ownership of the value",
+                    "enum": [
+                        "with",
+                        "without"
```
lokokung, line 141: Some bikeshedding... `with` and `without` feel a bit implicit? What if we used something more explicit like `application` vs `api`?

lokokung (approved):
Overall seems good to me! I am wondering about whether the `with` and `without` ownership terminology is a bit confusing, but I'm not sure if I have a necessarily better proposal.

schema.json
kainino0x: Whoops I shouldn't have enabled automerge. Was thinking of that too, but used "With" and "Without" to match the ownership article. Maybe `with_ownership: boolean`?

PR #376, opened 2024-10-18
WGPUChainedStructOut -> WGPUChainedStructMut

### LINKED ISSUE #260: In/out extensibility of WGPULimits and WGPUInstanceFeatures
It should be possible to ask wgpuGetInstanceFeatures for more info by adding structs to the chain. As written out here:
https://github.com/webgpu-native/webgpu-headers/issues/199#issuecomment-1701605454
this is not possible because `nextInChain` is `const *`.

We can fix this by mirroring what we have already for WGPURequiredLimits/WGPUSupportedLimits:

```c
typedef struct WGPUInstanceFeatures {
    WGPUBool timedWaitAnyEnable;
    size_t timedWaitAnyMaxCount;
} WGPUInstanceFeatures WGPU_STRUCTURE_ATTRIBUTE;

typedef struct WGPUSupportedInstanceFeatures {
    WGPUChainedStructOut * nextInChain;
    WGPUInstanceFeatures features; // out
} WGPUSupportedInstanceFeatures WGPU_STRUCTURE_ATTRIBUTE;

typedef struct WGPUInstanceDescriptor {
    WGPUInstanceFeatures features; // in
} WGPUInstanceDescriptor WGPU_STRUCTURE_ATTRIBUTE;
```

Unfortunately for both this and WGPULimits, we need to duplicate stuff for every extension, like:

```c
typedef struct WGPUDawnExperimentalSubgroupLimits {
    uint32_t minSubgroupSize;
    uint32_t maxSubgroupSize;
} WGPUDawnExperimentalSubgroupLimits WGPU_STRUCTURE_ATTRIBUTE;

// Can be chained in WGPUSupportedLimits
typedef struct WGPUDawnExperimentalSubgroupSupportedLimits {
    WGPUChainedStructOut chain;
    WGPUDawnExperimentalSubgroupLimits limits; // out
} WGPUDawnExperimentalSubgroupSupportedLimits WGPU_STRUCTURE_ATTRIBUTE;

// Can be chained in WGPURequiredLimits
typedef struct WGPUDawnExperimentalSubgroupRequiredLimits {
    WGPUChainedStruct chain;
    WGPUDawnExperimentalSubgroupLimits limits; // in
} WGPUDawnExperimentalSubgroupRequiredLimits WGPU_STRUCTURE_ATTRIBUTE;
```

(Dawn is actually currently missing an equivalent of the latter. I suspect this is because we're abusing "limits" for what are really adapter properties?)

Note the JS API has a similar "problem" except it doesn't have to deal with extension structs.

---

A totally different option could be to use an array of structs instead of a chain of structs, kind of a key-value-pair system vaguely like what JS uses. We could use this for both. Example for limits:

```c
struct WGPUExtendedLimitsBase {
    WGPUSType sType;
};

struct WGPUDawnExperimentalSubgroupLimits {
    WGPUExtendedLimitsBase type;
    uint32_t minSubgroupSize;
    uint32_t maxSubgroupSize;
};

// Pass in an array of pointers to the limit structs you want populated
WGPUBool wgpuAdapterGetLimits(WGPUAdapter adapter, WGPULimits * limits,
    size_t extendedLimitStructCount, WGPUExtendedLimitsBase * const * extendedLimitStructs);

struct WGPUDeviceDescriptor {
    // ...
    WGPU_NULLABLE WGPURequiredLimits const * requiredLimits;
    // Pass in an array of pointers to the limit structs containing limits you require.
    size_t requiredExtendedLimitStructCount;
    WGPU_NULLABLE WGPULimitStructBase const * const * requiredExtendedLimitsStructs;
    // ...
};
```

Or we could go all the way to what JS has with a real key-value interface, which is much simpler to write out - though probably not easier to use, and unfortunately a much more breaking change.

```c
struct WGPULimit {
    WGPULimitName name;
    // Due to JS we are probably unlikely to ever have limits that can't be doubles (though possible with BigInt).
    // Note we no longer need WGPU_LIMIT_*_UNDEFINED because you can just not list the limit.
    double value;
};

WGPUBool wgpuAdapterGetLimits(WGPUAdapter adapter, size_t limitCount, WGPULimit * limits);

struct WGPUDeviceDescriptor {
    // ...
    size_t limitCount;
    WGPULimit const * limits;
    // ...
};
```

---

- [x] Side thought, maybe since instance features aren't quite analogous to device features (more like device features+limits), they should be called something else like "InstanceCapabilities". Filed as #375

Related to #216.
cc @lokokung

kainino0x:
Proposal.

See https://github.com/webgpu-native/webgpu-headers/issues/260#issuecomment-1910891166

Fixes #260

### DIFF
--- a/gen/cheader.tmpl
+++ b/gen/cheader.tmpl
@@ -261,10 +261,10 @@ typedef struct WGPUChainedStruct {
     WGPUSType sType;
 } WGPUChainedStruct WGPU_STRUCTURE_ATTRIBUTE;
 
-typedef struct WGPUChainedStructOut {
-    struct WGPUChainedStructOut * next;
+typedef struct WGPUChainedStructMut {
+    struct WGPUChainedStructMut * next;
     WGPUSType sType;
-} WGPUChainedStructOut WGPU_STRUCTURE_ATTRIBUTE;
+} WGPUChainedStructMut WGPU_STRUCTURE_ATTRIBUTE;
 
 /** @} */
 {{  end}}
@@ -305,17 +305,17 @@ typedef struct WGPU{{.Name | PascalCase}}{{$.ExtSuffix}} {
 {{-   if eq .Type "base_in" }}
     WGPUChainedStruct const * nextInChain;
 {{-   else if eq .Type "base_out" }}
-    WGPUChainedStructOut * nextInChain;
+    WGPUChainedStructMut * nextInChain;
 {{-   else if eq .Type "base_in_or_out" }}
     /** This struct chain is used as mutable in some places and immutable in others. */
-    WGPUChainedStructOut * nextInChain;
+    WGPUChainedStructMut * nextInChain;
 {{-   else if eq .Type "extension_in"}}
     WGPUChainedStruct chain;
 {{-   else if eq .Type "extension_out"}}
-    WGPUChainedStructOut chain;
+    WGPUChainedStructMut chain;
 {{-   else if eq .Type "extension_in_or_out"}}
     /** This struct chain is used as mutable in some places and immutable in others. */
-    WGPUChainedStructOut chain;
+    WGPUChainedStructMut chain;
 {{-   end}}
 {{-   range $memberIndex, $_ := .Members}}
     {{- MCommentMember . 4 }}
--- a/webgpu.h
+++ b/webgpu.h
@@ -1223,10 +1223,10 @@ typedef struct WGPUChainedStruct {
     WGPUSType sType;
 } WGPUChainedStruct WGPU_STRUCTURE_ATTRIBUTE;
 
-typedef struct WGPUChainedStructOut {
-    struct WGPUChainedStructOut * next;
+typedef struct WGPUChainedStructMut {
+    struct WGPUChainedStructMut * next;
     WGPUSType sType;
-} WGPUChainedStructOut WGPU_STRUCTURE_ATTRIBUTE;
+} WGPUChainedStructMut WGPU_STRUCTURE_ATTRIBUTE;
 
 /** @} */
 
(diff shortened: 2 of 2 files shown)


lokokung:
Hmm, what is the reasoning behind the name? I assume `Mut` is short for something like `Mutable`? Couldn't we just be explicit and call it `Mutable` then?

kainino0x:
It is short for Mutable, yes, maybe I am too used to Rust where that's a keyword (opposite of const) :)

Basically since now we use these chains as either input or output, calling them "out" chains doesn't make sense.

Kangz:
Given we are in C, these would be more like `WGPUConstChainedStruct` and `WGPUChainedStruct` because mutability is the default. As much as I think Rust got it right with default constness, it would be better to stick to the semantics of C here.

Kangz:
Also I don't think that the changes you did for in-out chains do what you intended. I might be missing something, but if the root of a chain is a `WGPUChainedStructMut`, then all members of the chain need to be `WGPUChainedStructMut`. Likewise for the non-mut ones. What we seem to want to allow is a mix of both mut and non-mut, IDK that there's a way to do this without splitting chains. Or we could make all the chains mut and just guarantee which ones will be mutated vs not.

kainino0x:
> if the root of a chain is a `WGPUChainedStructMut`, then all members of the chain need to be `WGPUChainedStructMut`.

My intent was to do this, not allow a mix of mut and non-mut structs in the chain, since that's not possible. I split the concept of `in_or_out` from `out` mainly so I could generate different docs for them. Idea is:
- base_in can only have extension_in in its chain
- base_out can only have extension_out in its chain
- base_in_or_out can only have extension_in_or_out in its chain

So it is possible to split that into three different C structs e.g. `WGPUChainedStruct{In,Out,InOrOut}` for some additional "safety". However `Out` and `InOrOut` would be otherwise identical so my simpler idea was just to give them both one name.

kainino0x:
> Or we could make all the chains mut and just guarantee which ones will be mutated vs not.

I considered this, my only concern is that it prevents applications from using const at function boundaries where they're passing struct chain links around, but I don't really know why they would do that, so maybe this is a better solution.

kainino0x:
Closing for now in favor of #395

Issue #194, opened 2023-06-14
GetMappedRange on READ-mapped buffer


kainino0x:
When a buffer is mapped for READ, GetConstMappedRange must be used to get a pointer. What happens if you use GetMappedRange? Dawn currently returns nullptr, but I think perhaps it should be invalid (e.g. crash on assertion in debug mode), for debuggability.

https://matrix.to/#/!ZSOHTEPDbwuEgSJwYw:matrix.org/$LHAITjEUC-TmB-YZ1des5TeUBO-FFRIvnZRfgfgbSZg?via=matrix.org

kainino0x:
With or without that change, these could also be renamed as explicitly Read or ReadWrite. Though I like the current names personally.

kainino0x:
In last week's meeting I think we settled on a proposal to:
- keep API surface as it is
- fail with an actual error string if you GetMappedRange on a READ mapping (see #143) or do anything else wrong (e.g. ask for an overlapping range, at least in Emscripten, see #64)

kainino0x:
> * keep API surface as it is

Sorry I think I probably forgot to finish typing this sentence? the actual proposal adds some stuff to the API surface, and is here https://github.com/webgpu-native/webgpu-headers/issues/143#issuecomment-1612208345

kainino0x:
We introduced the concept of "implementation-defined logging", which probably happened after the resolution above. I would prefer to use that here, and simply say "if there is an error, returns null and should log an error message to an implementation-defined log"

lokokung:
+1 on using the implementation-defined logging for this to avoid the additional argument change. A validation error could also be a possible solution.

eliemichel:
Couldn't the use of `wgpuBufferGetMappedRange` on a read-only mapped buffer trigger the device's uncaptured error callback? I remember doing this mistake a couple of time, and it always took me a bit more time to figure out my mistake than if the device's error callback would have been directly called to tell me "what you do makes no sense".

I believe this situation only occurs in case a programming error that should be solved by fixing the source code of the application, rather than something one wants to catch at runtime (but there can be edge cases I don't think about ofc).

eliemichel:
All right following the JS API is important enough, no need to create an overload if there is clear enough logging :)

kainino0x:
> We introduced the concept of "implementation-defined logging", which probably happened after the resolution above. I would prefer to use that here, and simply say "if there is an error, returns null and should log an error message to an implementation-defined log"

[...]

PR #377, opened 2024-10-19
Add and document many synchronous error statuses

### LINKED ISSUE #64: Synchronous validation in getMappedRange
How should synchronous validation errors be exposed in getMappedRange? It can return null, but that doesn't tell the caller what went wrong. This would probably be fine, except e.g. Chromium needs to know what string to attach to the DOMException it generates. To do that now it needs to duplicate all of the validation that's already in getMappedRange.

### LINKED ISSUE #115: Handling invalid/not-enabled chained structs on extensible Out-structs
WebGPU's [`limits`](https://gpuweb.github.io/gpuweb/#dom-gpuadapter-limits) don't seem to be optional, so it doesn't make much sense why `wgpuAdapterGetLimits()` returns a bool.

### LINKED ISSUE #225: Tracking issue: synchronous exceptions in JS API
I'll attempt to make a complete list of entrypoints that have synchronous exceptions in the JS API:

*   [x] All cases where you try to use a feature that's not enabled on the device that would throw a TypeError on a browser that hasn't implemented the feature at all ([see spec](https://gpuweb.github.io/gpuweb/#features))
    *   Using an enum value introduced by the feature (e.g. query type "timestamp")
    *   Calling a method introduced by the feature (e.g. writeTimestamp)
*   [x] getMappedRange: not mapped; not aligned; out of range; overlaps (#64)
*   [x] createBuffer with mappedAtCreation (OOM-ish conditions) (#57)
*   [ ] external image stuff (copyExternalImageToTexture/importExternalTexture) (not in C API yet, #232):
    *   "origin-clean" and "check the usability" of external image sources
    *   copyExternalImageToTexture source image size check (can be checked in the bindings beforehand so it’s possible to know what the error is, except maybe for HTMLVideoElement because it’s weird and frames advance in the background)
*   [x] writeBuffer: size alignment. Maybe this should be a validation error in the upstream spec?
*   [x] getCurrentTexture: not configured. See #197, maybe we need an extra GetCurrentTextureStatus for this?
*   [x] Things that can't happen from C
    *   Invalid extent/origin/color arrays
    *   Errors about the size of a source BufferSource in writeBuffer/writeTexture/setBindGroup
*   [x] (EDIT 2024-10-18) Invalid formats for configure() - this was recently moved from device timeline to content timeline in the JS spec.

One approach we've been taking for some entrypoints is to say "return a generic error code, with implementation-defined logging" (eventually we're likely to make it well-defined, with a log callback on the instance):

- CreateInstance/GetInstanceFeatures https://github.com/webgpu-native/webgpu-headers/issues/199#issuecomment-1701569747
- GetMappedRange https://github.com/webgpu-native/webgpu-headers/issues/64#issuecomment-1710774712.

### LINKED ISSUE #255: Error management in unconfigured Surface methods
A `Surface` object is created from an `Instance`, so it has no associated `Device` by default, until its `Configure` method gets called.

However, the typical management of errors is done through the `Device`'s `UncapturedErrorCallback`.

So how should a `Surface` report an error when a method like `Present` is called before the surface was configured? This concern holds for all methods that were inherited from the former `SwapChain` object.

### LINKED ISSUE #291: WGPUSurfaceGetCurrentTextureStatus needs a value for "surface not configured"
There's just one error case in `getCurrentTexture()` in the JS spec:
> 1. If this.[[[configuration]]](https://gpuweb.github.io/gpuweb/#dom-gpucanvascontext-configuration-slot) is null:
>     1. Throw an [InvalidStateError](https://webidl.spec.whatwg.org/#invalidstateerror) and stop.

We don't have a `WGPUSurfaceGetCurrentTextureStatus` to represent this currently.

```c
typedef enum WGPUSurfaceGetCurrentTextureStatus {
    WGPUSurfaceGetCurrentTextureStatus_Success = 0x00000000,
    WGPUSurfaceGetCurrentTextureStatus_Timeout = 0x00000001,
    WGPUSurfaceGetCurrentTextureStatus_Outdated = 0x00000002,
    WGPUSurfaceGetCurrentTextureStatus_Lost = 0x00000003,
    WGPUSurfaceGetCurrentTextureStatus_OutOfMemory = 0x00000004,
    WGPUSurfaceGetCurrentTextureStatus_DeviceLost = 0x00000005,
    WGPUSurfaceGetCurrentTextureStatus_Force32 = 0x7FFFFFFF
} WGPUSurfaceGetCurrentTextureStatus WGPU_ENUM_ATTRIBUTE;
```

Should we add a new one, or merge with one of those?

Followup issue to #203. cc @rajveermalviya @Kangz @eliemichel

kainino0x:
Sorry this is a lot of things in one PR, but I kept changing all of the docs to be consistent with each other, so this was easier.
(Can split if there are some parts that need more time to discuss.)

Fixes #255
Fixes #291 (again?)
~Issue~Fixes #225
Fixes #64 (EDIT: and its duplicate #194)
EDIT: also fixes #115

### DIFF
--- a/doc/articles/Errors.md
+++ b/doc/articles/Errors.md
@@ -1,0 +1,42 @@
+# Errors {#Errors}
+
+Errors are surfaced in several ways.
+
+Most errors only result from incorrect use of the API and should not need to be handled at runtime.
+However, a few (@ref WGPUErrorType_OutOfMemory, @ref WGPUErrorType_Internal) are potentially useful to handle.
+
+## Device Error {#DeviceError}
+
+These behave the same way as [in the WebGPU JavaScript API specification](https://www.w3.org/TR/webgpu/#errors-and-debugging).
+They are receivable via @ref wgpuDevicePopErrorScope() and @ref WGPUDeviceDescriptor::uncapturedErrorCallbackInfo.
+
+These errors include:
+
+- All device-timeline errors in the WebGPU specification.
+- Enum values which are numerically invalid (this is not possible in JavaScript).
+- Enum values which are require features not enabled on the device (a [content-timeline](https://www.w3.org/TR/webgpu/#content-timeline) error in JavaScript), for example compressed texture formats.
+- Other content-timeline errors where specified.
+
+## Callback Errors {#CallbackErrors}
+
+These behave similarly to the Promise-returning JavaScript APIs. Instead of there being two callbacks like in JavaScript (one for resolve and one for reject), there is a single callback which receives a status code, and depending on the status, _either_ a valid result with an empty message string (`{NULL, 0}`), _or_ an invalid result with a non-empty message string.
+
+## Synchronous Errors {#SynchronousErrors}
+
+These errors include:
+
+- @ref StructChainingError cases.
+- [Content-timeline](https://www.w3.org/TR/webgpu/#content-timeline) errors other than those which are surfaced as @ref DeviceError in `webgpu.h`. See specific documentation to determine how each error is exposed.
+
+Generally these will return some kind of failure status (like \ref WGPUStatus_Error), or `NULL`.
+
+### Struct-Chaining Error {#StructChainingError}
+
+A struct-chaining error happens when the `SType` of a struct in a struct chain is not valid for that chain.
+The API returns a failure status or `NULL` and produces an @ref ImplementationDefinedLogging message.
+
+### Implementation-Defined Logging {#ImplementationDefinedLogging}
+
+Entry points may also specify that they produce "implementation-defined logging".
+These messages are logged in an implementation defined way (e.g. to an implementation-specific callback, or to a logging runtime).
+They are intended to be intended to be read by humans, useful primarily for development and crash reporting.
--- a/doc/articles/Errors.md
+++ b/doc/articles/Errors.md
@@ -17,26 +17,34 @@ These errors include:
 - Enum values which are require features not enabled on the device (a [content-timeline](https://www.w3.org/TR/webgpu/#content-timeline) error in JavaScript), for example compressed texture formats.
 - Other content-timeline errors where specified.
 
-## Callback Errors {#CallbackErrors}
+## Callback Error {#CallbackError}
 
 These behave similarly to the Promise-returning JavaScript APIs. Instead of there being two callbacks like in JavaScript (one for resolve and one for reject), there is a single callback which receives a status code, and depending on the status, _either_ a valid result with an empty message string (`{NULL, 0}`), _or_ an invalid result with a non-empty message string.
 
-## Synchronous Errors {#SynchronousErrors}
+## Synchronous Error {#SynchronousError}
 
 These errors include:
 
-- @ref StructChainingError cases.
+- @ref SynchronousStructChainingError cases.
 - [Content-timeline](https://www.w3.org/TR/webgpu/#content-timeline) errors other than those which are surfaced as @ref DeviceError in `webgpu.h`. See specific documentation to determine how each error is exposed.
 
-Generally these will return some kind of failure status (like \ref WGPUStatus_Error), or `NULL`.
-
-### Struct-Chaining Error {#StructChainingError}
-
-A struct-chaining error happens when the `SType` of a struct in a struct chain is not valid for that chain.
-The API returns a failure status or `NULL` and produces an @ref ImplementationDefinedLogging message.
+Generally these will return some kind of failure status (like \ref WGPUStatus_Error) or `NULL`, and produce an @ref ImplementationDefinedLogging message.
 
 ### Implementation-Defined Logging {#ImplementationDefinedLogging}
 
 Entry points may also specify that they produce "implementation-defined logging".
 These messages are logged in an implementation defined way (e.g. to an implementation-specific callback, or to a logging runtime).
 They are intended to be intended to be read by humans, useful primarily for development and crash reporting.
+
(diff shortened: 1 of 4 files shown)


doc/articles/Errors.md
```diff
@@ -0,0 +1,42 @@
...
+## Synchronous Errors {#SynchronousErrors}
+
+These errors include:
+
+- @ref StructChainingError cases.
+- [Content-timeline](https://www.w3.org/TR/webgpu/#content-timeline) errors other than those which are surfaced as @ref DeviceError in `webgpu.h`. See specific documentation to determine how each error is exposed.
+
+Generally these will return some kind of failure status (like \ref WGPUStatus_Error), or `NULL`.
+
+### Struct-Chaining Error {#StructChainingError}
```
Kangz, line 33: Why can't these piggy-back on device errors when possible? There's a lot of call that take extensible structs and that return void.

Kangz: Basically they are validation errors.

webgpu.yml
```diff
@@ -840,6 +840,8 @@ enums:
         doc: The system ran out of memory.
       - name: device_lost
         doc: The @ref WGPUDevice configured on the @ref WGPUSurface was lost.
+      - name: error
+        doc: The surface is not configured, or there was a @ref StructChainingError.
```
Kangz, line 844: Maybe just point at the Surface.md article for a description of what might return an error?

webgpu.yml
```diff
@@ -3184,8 +3190,8 @@ objects:
           TODO
         returns:
           doc: |
-            TODO
-          type: bool
+            Returns @ref WGPUStatus_Error (and leaves `limits` unchanged) if there was a @ref StructChainingError.
```
Kangz, line 3193: What about if the device/adapter itself is an error? IMHO the error cases should be defined in the algorithm steps of the articles, and not call out a particular kind of error.

webgpu.yml
```diff
@@ -3291,34 +3299,43 @@ objects:
           TODO
         returns:
           doc: |
-            TODO
+            Returns a mutable pointer to beginning of the mapped range.
+            It may be written, but reading is undefined behavior unless the buffer was also mapped with @ref WGPUMapMode_Read.
```
Kangz, line 3303: Reading should be a valid for mapped write buffers, because any mapped range should get flushed to the GPU and not be modified if the implementation didn't end up writing to it. That's what happens in JS.

webgpu.yml
```diff
@@ -4081,7 +4096,8 @@ objects:
         callback: callback.queue_work_done
       - name: write_buffer
         doc: |
-          TODO
+          Produces a @ref DeviceError if `size` is not a multiple of 4
```
Kangz, line 4099: This error is already handled in the JS spec.

webgpu.yml
kainino0x: I couldn't remember if this was a requirement for native mapping to avoid reading uninitialized memory. I know reading is valid in JS, but that was also because it's currently impossible to modify a MAP_WRITE buffer _except_ from the CPU via mapping, so there can be a CPU-side shadow copy. That shadow copy prevents JS from actually reading from the real map anyway.

doc/articles/Errors.md
kainino0x: I am pretty sure you're right that's what we were supposed to do. I kind of lost the thread of the discussion over the many issues we've discussed this stuff, but we were mostly talking about the ones with out-structs.

webgpu.yml
kainino0x: I kind of think struct chaining isn't really relevant to pile into the Surface.md article and prefer to just document it here.

webgpu.yml
kainino0x: This is meant to mirror `adapter.limits`/`device.limits` in JS, where there is no possibility of the JS-wrapper object being an error.

I'm OK with moving this stuff into articles if we want to write those, but it's easier to start documenting things here and move them to articles if they get larger.

webgpu.yml
kainino0x: In the JS spec it's a content-timeline error; here I'm noting that it's different because it's a DeviceError. I could say that all content-timeline errors in the JS spec are DeviceErrors, but the other two don't make sense in C.

kainino0x:
Revised finally.

kainino0x:
@Kangz re-review bump!

webgpu.h
```diff
@@ -3233,8 +3248,36 @@ WGPU_EXPORT void wgpuBindGroupLayoutRelease(WGPUBindGroupLayout bindGroupLayout)
...
+ *
+ * @param size
+ * Byte size of the range to get. The returned pointer is valid for exactly this many bytes.
+ *
+ * @returns
+ * Returns a const pointer to beginning of the mapped range.
+ * It must not be written; writing to this range causes undefined behavior.
+ * Returns `NULL` with @ref ImplementationDefinedLogging if:
+ *
+ * - There is any content-timeline error as defined in the WebGPU specification for `getMappedRange()` (alignments, overlaps, etc.)
```
Kangz, line 3263: Note that we don't check for overlapped ranges in Dawn. But maybe that's ok to keep as a caveat.

Kangz (approved):

webgpu.h
kainino0x: Oh, good catch, we did actually say we would allow this (I think). Will fix.
https://github.com/webgpu-native/webgpu-headers/issues/64#issuecomment-1613603534

PR #384, opened 2024-10-31
Add wgpuDeviceGetAdapterInfo

beaufortfrancois:
Spec issue: https://github.com/gpuweb/gpuweb/issues/4810

Spec PR: https://github.com/gpuweb/gpuweb/pull/4809

Chromium bug: https://issues.chromium.org/issues/376600838

### DIFF
--- a/webgpu.yml
+++ b/webgpu.yml
@@ -3926,6 +3926,14 @@ objects:
             type: struct.supported_features
             pointer: mutable
             passed_with_ownership: true
+      - name: get_adapter_info
+        doc: |
+          TODO
+        returns:
+          doc: |
+            TODO
+          type: struct.adapter_info
+          passed_with_ownership: true
       - name: get_queue
         doc: |
           TODO
--- a/webgpu.h
+++ b/webgpu.h
@@ -2534,6 +2534,11 @@ typedef WGPUTexture (*WGPUProcDeviceCreateTexture)(WGPUDevice device, WGPUTextur
  */
 typedef void (*WGPUProcDeviceDestroy)(WGPUDevice device) WGPU_FUNCTION_ATTRIBUTE;
 /**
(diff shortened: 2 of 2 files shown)


webgpu.yml
```diff
@@ -3926,6 +3926,14 @@ objects:
...
             pointer: mutable
             passed_with_ownership: true
+      - name: get_adapter_info
+        doc: |
+          TODO
+        returns:
+          doc: |
+            TODO
+          type: struct.adapter_info
+          passed_with_ownership: true
```
kainino0x, line 3936: `passed_with_ownership` is only for args, it gets ignored return values. (maybe it should be supported for return values, but it isn't currently; not sure if there are even any return values that are _not_ passed with ownership.)

kainino0x (approved):

webgpu.yml
beaufortfrancois: I initally followed same pattern as `get_queue`. 
Looking at it, here is a non-exhaustive list of return values usages of `passed_with_ownership`:
`create_surface`, `finish`, `get_bind_group_layout`, `create_view`, `create_instance`, `begin_compute_pass`,  `begin_render_pass`, `get_bind_group_layout`, `create_bind_group`, `create_bind_group_layout`, `create_buffer`, `create_command_encoder`, `create_compute_pipeline`, etc.

Shall we update there as well?

kainino0x: Ah sorry, I didn't realize, I must have put passed_with_ownership on a bunch of return values without thinking about the fact that it wouldn't generate docs.

That's fine then, let's keep it as is, and I'll try to remember to make it generate docs later!

beaufortfrancois: Okay!
I think we're good now then

PR #390, opened 2024-11-07
Remove UTF-16 from WGPUCompilationMessage; add docs

### LINKED ISSUE #246: Character counting in WGPUCompilationMessage
webgpu.h takes shaders in UTF-8.
Right now WGPUCompilationMessage is supposed to return the linePos/offset/length in units of both UTF-8 and UTF-16 code units:
```
    ...
    uint64_t lineNum;
    uint64_t linePos;
    uint64_t offset;
    uint64_t length;
    uint64_t utf16LinePos;
    uint64_t utf16Offset;
    uint64_t utf16Length;
```

These are the same for ASCII, but otherwise they differ:
- `c` = `U+0063` = utf-8 `63` (1) = utf-16 `0063` (1)
- `ç` = `U+00E7` = utf-8 `C3 A7` (2) = utf-16 `\u00E7` (1)
- `😀` = `U+1F600` = utf-8 `F0 97 98 80` (4) = utf-16 `\uD83D\uDE00` (2)

Problems:

- Computing these is wasteful because any given caller only actually wants one of them (JavaScript, JVM, and some Win32-centric C++ code want UTF-16; almost everybody else wants UTF-8). We can optimize this computation away only if we know the WGSL source is pure ASCII.
- For WASM, it's nontrivial to convert the UTF-16 offsets we get from JS back to UTF-8 offsets for the WASM code. Unfortunately there's no avoiding this, unless the JS API changes to provide both or let you choose which one you want.
- I'm not completely confident it's actually possible to unambiguously compute UTF-16 code unit offsets given a UTF-8 string. I think it is possible though - see [Dawn's implementation](https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/src/dawn/native/CompilationMessages.cpp;l=107;drc=cc2516ac15ebeb6c3c07a427a76d6b57832d3154) (also note the WebIDL doesn't allow unpaired surrogates which simplifies things).

I think we should probably add some request somewhere (in WGPUShaderModuleDescriptor?) of what units the caller wants. Then we can remove the `utf16` fields and just return values in the units that were requested.

kainino0x:
Fixes #246 (dawn bug https://crbug.com/377940841)

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -1410,17 +1410,31 @@ typedef struct WGPUCommandEncoderDescriptor {
 typedef struct WGPUCompilationMessage {
     WGPUChainedStruct const * nextInChain;
     /**
+     * Human-readable message text (non-computer-readable; may be locale dependent!). UTF-8 encoded.
+     *
      * This is an \ref OutputString.
      */
     WGPUStringView message;
+    /**
+     * Severity level of the message.
+     */
     WGPUCompilationMessageType type;
+    /**
+     * Line number where the message is attached, starting at 1.
+     */
     uint64_t lineNum;
+    /**
+     * Offset in UTF-8 code units (bytes) from the beginning of the line, starting at 1.
+     */
     uint64_t linePos;
+    /**
+     * Offset in UTF-8 code units (bytes) from the beginning of the shader code, starting at 0.
+     */
     uint64_t offset;
+    /**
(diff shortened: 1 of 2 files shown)


lokokung:
> LGTM, but please file Dawn side bug to track this update there (I think it should be a pretty small fix)

Nvm, I found https://issues.chromium.org/issues/377940841 for this one.

PR #394, opened 2024-11-09
[proposal] Change SetBlendConstant to take 4 floats

### LINKED ISSUE #378: Should `SetBlendConstant` take float,float,float,float instead of WGPUColor?
`WGPUColor` is set up to use `double` because it can represent all of the possible render pass clear values - it's used as a union of `f32`, `i32`, and `u32`.

However, AFAIK, blending, and therefore `wgpuRenderPassEncoderSetBlendConstant`, never applies to sint/uint type textures, so should we just take four 32-bit `float`s instead? Slightly more efficient.

kainino0x:
Fixes #378 (dawn bug: TODO)

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -172,7 +172,7 @@
                         "doc": {
                             "type": "string",
                             "$comment": "Doxygen doesn't support multi-paragraph param docs (containing \\n\\n)",
-                            "pattern": "^\n?.+(\n.+)*\n?$"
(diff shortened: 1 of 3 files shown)


lokokung (approved):
LGTM, but can we file a Dawn side bug to update this as well? (I don't think it exists atm?)

kainino0x:
I'll be sure to file it if we decide to make this change.

Issue #209, opened 2023-08-03
Reporting DeviceLost in various functions and callbacks


kainino0x:
Currently we report DeviceLost in the following places aside from just the DeviceLost callback - places where the JS API does _not_ report it:
- MapAsyncStatus (see #183)
- CompilationInfoRequestStatus
- CreatePipelineAsyncStatus
- WorkDoneStatus
- PopErrorScope
- GetCurrentTextureStatus (after #203)

The JS API specifies generally that async operations pretend to work successfully after the device is lost (except for MapAsync): https://github.com/gpuweb/gpuweb/issues/1629#issuecomment-1664642751

Should we keep these DeviceLost enum values or should we do more like what the JS API does and pretend everything worked? Notably WASM can only do what JS does, so this makes it a native-only return value - but so are many of the others.

Kangz:
We should probably match the JS API and pretend things worked.

kainino0x:
*   (discussion)
*   **Remove the DeviceLost option everywhere**
*   including for MapAsync, return Unknown?
*   **Yes, remove it**
*   See also #183

kainino0x:
> * including for MapAsync, return Unknown?

This changes to `Aborted` because that's what the upstream JS API does now.

kainino0x:
This was done in #396 and #397.

PR #397, opened 2024-11-09
Remove DeviceLost status from WGPUSurfaceGetCurrentTextureStatus

kainino0x:
~We didn't discuss which status we would use for this, but "Lost" seems to makes sense.~
This should behave like JS where we pretend it worked by returning a texture, but you can't use it because the device is lost.

Issue #209

### DIFF
--- a/doc/articles/Surfaces.md
+++ b/doc/articles/Surfaces.md
@@ -234,23 +234,24 @@ wgpuTextureRelease(surfaceTexture.texture);
 
 The behavior of `::wgpuSurfaceGetCurrentTexture``(surface, surfaceTexture)` is:
 
- - Set `surfaceTexture->texture` to `NULL`.
+1. Set `surfaceTexture->texture` to `NULL`.
+1. If any of these validation steps fails, set `surfaceTexture->status` to `WGPUSurfaceGetCurrentTextureStatus_Error` and return (TODO send error to device?).
+    1. Validate that `surface` is not an error.
+    1. Validate that `surface.config` is not `None`.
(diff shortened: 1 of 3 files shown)


kainino0x:
I just realized... in JS if you getCurrentTexture on a lost device, you just get an invalid texture (well, a texture that you can't observe is valid, because the device is lost). That is probably needs to be done here, to match - it's not exactly possible in JS or on remoting implementations to do what I wrote here, because you don't know synchronously when the device is lost. Hopefully this isn't a problem for non-remoting native....

Revised.

doc/articles/Surfaces.md
```diff
@@ -242,8 +242,8 @@ The behavior of `::wgpuSurfaceGetCurrentTexture``(surface, surfaceTexture)` is:
    - Validate that `surface.config` is not `None`.
    - Validate that `surface.currentFrame` is `None`.
 
- - If `surface.config.device` is not alive, set `surfaceTexture->status` to `WGPUSurfaceGetCurrentTextureStatus_DeviceLost` and return.
- - If the implementation detects any other problem preventing use of the surface, set `surfaceTexture->status` to an appropriate status (something other than `SuccessOptimal`, `SuccessSuboptimal`, `Error`, or `DeviceLost`) and return.
+ - If `surface.config.device` is not alive, set `surfaceTexture->status` to `WGPUSurfaceGetCurrentTextureStatus_Success`, set `surfaceTexture->texture` to a new invalid `WGPUTexture`, and return.
```
Kangz, line 245: We should still have the correct reflection on that texture.

Kangz (approved):

Issue #293, opened 2024-05-21
Hardware feature levels in WGPURequestAdapterOptions


kainino0x:
I have posted a proposal for the JS API here:
https://github.com/gpuweb/gpuweb/issues/4656

I would like to do the same thing in the C API, but propose one major difference: we *require* a feature level in WGPURequestAdapterOptions instead of defaulting to "FL1"/"Core".

kainino0x:
May 9 meeting:

*   Kai’s proposal for C API: same but **require** minFeatureLevel (it has no default, for now)
*   …
*   CF: Feels like “compat” is now the actual baseline…
*   KN: Compat by default?
*   CF: As long as we have features to reach back up to core
*   …
*   AE: Branching feature levels? “modern mobile”/”modern desktop”. Should be OK, you can be upgraded along the same path
    *   Features that will be in mobile but not desktop? ASTC, tile local storage?
*   CF: Between defaulting to compat and requiring a feature level request, don’t think I care.
*   AE: D3D allows you to request a _list_ of feature levels.
    *   CF: Weird D3D11ism, like Win7 doesn’t know what FL11\_1 is, you have to try twice.
*   **KN: will bring this proposal to the WG and we can talk about it again after it’s had some discussion there**

kainino0x:
One point someone on our team raised about having webgpu.h default to compat, is that compat does enable a bunch of extra validation some of which is not trivial (for example the validation that one texture is only used with one sampler).

PR #399, opened 2024-11-11
Add WGPUFeatureLevel

kainino0x:
I also changed the struct order to go: required -> optional -> native-only optional.

Issue #293 - does not officially resolve it yet because we need to decide on defaulting.
(dawn bug https://crbug.com/378514251)

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -1642,13 +1642,19 @@ typedef struct WGPURequestAdapterOptions {
      */
     WGPUFeatureLevel featureLevel;
     WGPUPowerPreference powerPreference;
+    /**
+     * If true, requires the adapter to be a "fallback" adapter as defined by the JS spec.
+     * If this is not possible, the request returns null.
+     */
     WGPUBool forceFallbackAdapter;
     /**
-     * If set, requires the adapter to have a particular backend type. If this is not possible, the request returns null.
+     * If set, requires the adapter to have a particular backend type.
+     * If this is not possible, the request returns null.
      */
     WGPUBackendType backendType;
     /**
-     * If set, requires the adapter to be able to output to a particular surface. If this is not possible, the request returns null.
+     * If set, requires the adapter to be able to output to a particular surface.
+     * If this is not possible, the request returns null.
      */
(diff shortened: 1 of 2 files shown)


kainino0x:
Oh also I changed the struct order to go require -> optional -> native-only optional.

lokokung (approved):

webgpu.h
```diff
@@ -1637,10 +1654,28 @@ typedef struct WGPURenderPassTimestampWrites {
 
 typedef struct WGPURequestAdapterOptions {
     WGPUChainedStruct const * nextInChain;
-    WGPU_NULLABLE WGPUSurface compatibleSurface;
+    /**
+     * "Feature level" for the adapter request. If an adapter is returned, it must support the features and limits in the requested feature level.
+     *
+     * Implementations may ignore @ref WGPUFeatureLevel_Compatibility and provide @ref WGPUFeatureLevel_Core instead. @ref WGPUFeatureLevel_Core is the default in the JS API, but in C, this field is **required** (must not be undefined).
```
Kangz, line 1660: Prior to this it would have been allowed to call `wgpuInstanceRequestAdapter(instance, nullptr)` to get any adapter that the instance chooses. Can't we say that it defaults to core like the JS API?

webgpu.h
```diff
@@ -507,6 +507,23 @@ typedef enum WGPUErrorType {
     WGPUErrorType_Force32 = 0x7FFFFFFF
 } WGPUErrorType WGPU_ENUM_ATTRIBUTE;
 
+/**
+ * See @ref WGPURequestAdapterOptions::featureLevel.
+ */
+typedef enum WGPUFeatureLevel {
+    /**
+     * `0x00000001`.
+     * "Compatibility" profile which can be supported on OpenGL ES 3.1.
```
Kangz, line 516: nit: and D3D11

kainino0x: I'm unclear on this, I thought we came to the conclusion D3D11 could actually support core? Or did that turn out not to be true?

Kangz: We guessed that maybe it is possible but we never tried. So for now it is safe to assume that D3D11 only provides compat.

PR #404, opened 2024-11-12
Proposal: Make TimestampWrites structs extensible, move nullability inside

kainino0x:
Issue: None, I think @Kangz found this while we were reviewing the header.


lokokung (approved):

kainino0x:
Nov 21 meeting:

- KN: Avoids a pointer. Only reason not to is that we have to migrate people. Worth effort? (Not really, to us.)
- **Leave it behind a pointer, but rename/merge the two identical structs into one (any chains can be validated differently depending on whether it's render or compute)**
- CF: yaml should express this as "optional" if it doesn't

PR #406, opened 2024-11-12
Make WGPURenderPassDepthStencilAttachment extensible

kainino0x:
WGPURenderPassColorAttachment is already extensible, this should be too.

Issue: None, I think @Kangz found this while we were reviewing the header.
(dawn bug https://crbug.com/378514256)


kainino0x:
Actually it's not a gimme that this should be extensible. WGPURenderPassColorAttachment is in an array so it's really annoying to extend otherwise. WGPURenderPassDepthStencilAttachment is optional (behind a pointer) but not in an array.

I still think it makes sense to make it extensible though.

lokokung (approved):

kainino0x:
Merging tentatively but keeping `!discuss` label.

kainino0x:
Nov 21 meeting:

- **This change is OK**

PR #411, opened 2024-11-12
Proposal: Move nullability inside WGPURenderPassDepthStencilAttachment

### LINKED ISSUE #408: Should nullability move inside WGPURenderPassDepthStencilAttachment?
`WGPURenderPassDescriptor.depthStencilAttachment` is a nullable pointer. Should we instead make it an inline struct and move the nullability to `.depthStencilAttachment.view`?

I think the only reason not to do this would be if we expect you would ever want to set a `WGPURenderPassDepthStencilAttachment` without a texture view (and without any new extended fields that take its place).

kainino0x:
Fixes #408

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -1630,6 +1630,9 @@ typedef struct WGPURenderBundleEncoderDescriptor {
 } WGPURenderBundleEncoderDescriptor WGPU_STRUCTURE_ATTRIBUTE;
 
 typedef struct WGPURenderPassDepthStencilAttachment {
+    /**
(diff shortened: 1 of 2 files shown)


Kangz:
I'm not looking forward to doing all the changes for this, but either decision here seems fine. (IDK that it is strictly better than what we have now though).

kainino0x:
It's more consistent with other structs, but given `WGPURenderPassColorAttachment` have to be a pointer (because they're an array), I don't feel that the consistency is very important, so I'm OK not doing this. Will close for now (issue is still on the agenda for discussion).

Issue #403, opened 2024-11-12
Representation of GPUBindGroupEntry


kainino0x:
In `WGPUBindGroupEntry` the `offset` and `size` fields are supposed to apply only if the binding is a `buffer`, but that's not really clear. We could:

- Leave as is
- Rename to `bufferBindingOffset` and `bufferBindingSize`
- Collect the three fields together into a `WGPUBufferBinding` [like in JS](https://gpuweb.github.io/gpuweb/#dictdef-gpubufferbinding)

Not sure it's actually worth the effort to change.

eliemichel:
My 2 cents: It would make sense to match the JS API (using `entry.bufferBinding.buffer` for nullability), leaving as-is is fine, ad-hoc renaming with "bufferBinding" prefix sounds less ideal to me.

kainino0x:
Nothing _really_ matches JS perfectly since JS has a real union and we have three values in a trenchcoat. Unlike `GPUBindGroupLayoutEntry` which uses this trenchcoat pattern in JS and therefore needs to carry over to C, we could technically use a real tagged union here. It would honestly be less ergonomic though, and it doesn't _really_ map any better - either Wasm has to inject an error if more than one of the three types is used, or it has to inject an error if the tag is invalid. (Or it can just debug-assert, which is probably how I'd implement it.)

All that said, I would probably pick `entry.buffer.buffer` if doing this from scratch, but it doesn't really seem worth changing. We can write docs that explain what `offset` and `size` are.

kainino0x:
Tentatively closing as I'd like to go with no change, but leaving the label to discuss if we want to reopen.

kainino0x:
Nov 21 meeting:

- KN: Probably not really worth changing. I already added docs to clarify.
- CF: Should really be a tagged union.
- KN: Concerned about extensibility (pattern for extending is messy). Maybe use SType?
- **Investigate more**

kainino0x:
Decided it's easier to separate this out from #438. Mirroring the options there:

0 - Today:

```c
struct WGPUBindGroupEntry {
    WGPUChainedStruct const * nextInChain;
    uint32_t binding;
    WGPU_NULLABLE WGPUBuffer buffer;
    uint64_t offset;
    uint64_t size;
    WGPU_NULLABLE WGPUSampler sampler;
    WGPU_NULLABLE WGPUTextureView textureView;
};
[...]

PR #414, opened 2024-11-14
Add docs to WGPUBindGroupEntry fields

kainino0x:
Issue #403

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -1398,11 +1398,35 @@ typedef struct WGPUAdapterInfo {
 
 typedef struct WGPUBindGroupEntry {
     WGPUChainedStruct const * nextInChain;
+    /**
+     * Binding index in the bind group.
+     */
     uint32_t binding;
+    /**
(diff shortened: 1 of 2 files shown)


webgpu.h
```diff
@@ -1398,11 +1398,35 @@ typedef struct WGPUAdapterInfo {
...
     uint32_t binding;
+    /**
+     * Set this if the binding is a buffer object.
+     * Otherwise must be null.
+     */
     WGPU_NULLABLE WGPUBuffer buffer;
+    /**
+     * If the binding is a buffer, this is the byte offset of the binding range.
+     * Otherwise ignored.
+     */
```
lokokung, line 1413: We still might move these into either a substruct or rename them right?

lokokung (approved):

webgpu.h
lokokung: Nvm, I just got to the update in the other issue where you mentioned we probably won't change it. The documentation should be fine then!

Issue #401, opened 2024-11-12
Status enums still have too many values


kainino0x:
Following on #396 we should check if there are any other extraneous status values. Checked box = definitely good, or we decided what to do about it.

These look good:

- WGPUPopErrorScopeStatus
  - [x] WGPUPopErrorScopeStatus_Success
  - [x] WGPUPopErrorScopeStatus_InstanceDropped

- WGPUErrorType (not really a status but we'll include it)
  - [x] WGPUErrorType_NoError
  - [x] WGPUErrorType_Validation
  - [x] WGPUErrorType_OutOfMemory
  - [x] WGPUErrorType_Internal
  - [x] WGPUErrorType_Unknown - this is a bit niche but it's used if the browser returns some new status code and Wasm bindings don't understand it and need to translate it into a WGPUErrorType. It could be renamed/renumbered to `Undefined=0` but I don't think there's any real reason to do that.

- WGPUAdapterType (this really isn't a status but it also has an "unknown")
  - [x] WGPUAdapterType_Unknown - this could also be renamed/renumbered to `Undefined=0`

- WGPUDeviceLostReason (also not technically a status) (see also https://github.com/webgpu-native/webgpu-headers/issues/299#issuecomment-2469381621)
  - [x] WGPUDeviceLostReason_Unknown - this is from the JS spec
  - [x] WGPUDeviceLostReason_Destroyed
  - [x] WGPUDeviceLostReason_InstanceDropped
  - [x] WGPUDeviceLostReason_FailedCreation

- WGPUWaitStatus
  - [x] WGPUWaitStatus_Success
  - [x] WGPUWaitStatus_TimedOut
  - [x] WGPUWaitStatus_UnsupportedTimeout
  - [x] WGPUWaitStatus_UnsupportedCount
  - [x] WGPUWaitStatus_UnsupportedMixedSources

These I'm not sure about:

- WGPURequestAdapterStatus
  - [x] WGPURequestAdapterStatus_Success
  - [x] WGPURequestAdapterStatus_InstanceDropped
  - [x] WGPURequestAdapterStatus_Unavailable - presumably this is used when no adapter is available, but JS treats this as "success" and returns a null adapter. Should we really have a separate status for this? (Probably yes.)
    **Keep**
  - [x] WGPURequestAdapterStatus_Error (validation error)
[...]

PR #415, opened 2024-11-14
Remove all "Unknown" async statuses

kainino0x:
These were used by Dawn to indicate wire connection errors, but those should either be Success/Unavailable or InstanceDropped (TBD which).

Issue #401


lokokung (approved):

kainino0x:
@rajveermalviya is this OK for wgpu-future? (see issue for context)

Issue #158, opened 2022-06-02
Default descriptor initialization (and discussion of C vs C++ callbacks)


mrshannon:
Many of the descriptors in WebGPU have default values, and more importantly default values which are not 0 and so a simple `memset` or zero initialization is not appropriate.

An example of what does not work
```c
WGPUTextureDataLayout TextureDataLayout = {0};
TextureDataLayout.bytesPerRow = 512;
```
In this case `rowsPerImage = 0` instead of `WGPU_COPY_STRIDE_UNDEFINED`. 

Therefore, I propose adding default initialization defines to `webgpu.h`. While it is possible for each user to create their own, there are some benefits to implementing this in `webgpu.h`:

- Adding fields later with default values to descriptors does not introduce undefined behavior in existing code.
- It is possible to change default values later without breaking existing code (while `WGPU_COPY_STRIDE_UNDEFINED` fixes this, many defaults do not have a define as they are simply 0).
- One source of truth

One way to implement this is to add defines for each descriptor:
```c
#define WGPUTextureDataLayout_DEFAULT { NULL, 0, WGPU_COPY_STRIDE_UNDEFINED, WGPU_COPY_STRIDE_UNDEFINED }
```

This would allow users to write:
```c
WGPUTextureDataLayout TextureDataLayout = WGPUTextureDataLayout_DEFAULT;
TextureDataLayout.bytesPerRow = 512;
```

---

TODO list:

- [ ] Don't forget to write compile tests for these! #240

grovesNL:
This sounds reasonable to me. We use defaults for a lot of types in `wgpu` and find it really useful, e.g. `PrimitiveState` https://github.com/gfx-rs/wgpu/blob/77b2c25684f971daa977596403f01c71d32ea685/wgpu/examples/msaa-line/main.rs#L65-L69

mrshannon:
> Also @mrshannon, you could use the C++ headers generated by Dawn, they have default initialization for the various structs.

Sort of off topic, but we were initially using the C++ headers but dropped them for the C headers as we kept having to cast everywhere because the callbacks always use the C types and not the C++ types. But that came with it's own issues (`WGPUBufferUsage` vs `WGPUBufferUsageFlags`), maybe it's time to consider switching back.

Kangz:
Any idea on how to improve WGPUBufferUsage vs. WGPUBufferUsageFlags?

[...]

PR #422, opened 2024-11-16
Add INIT macros with default values

eliemichel:
Motivated by #158, and inspired by how Dawn does it, this PR introduces a `WGPU_FOO_INIT` macro to initialize a struct `WGPUFoo` with default values.

- [x] Generate INIT macro for all structs
- [x] Collect default values for all struct from WebGPU JS spec
- [x] Enable the use of constants as default values
- [x] State default values in generated docstring (point to INIT macro in struct docstring, give default value in member docstring)
- [x] Initialize SType for `WGPUChainedStruct` and `WGPUChainedStructOut`


eliemichel:
Another limitation spotted: `.chain` fields are not automatically initialized with their proper `SType`. This is related to the previous issue in that we should initialize the `WGPUChainedStruct` type differently depending on its parent struct.

kainino0x:
Wow, thank you so much!! I have been working through all of the [resolved changes](https://github.com/webgpu-native/webgpu-headers/issues?q=is%3Aissue+is%3Aopen+-label%3Anon-breaking+label%3A%22has+resolution%22) and many tentative changes but I had been procrastinating on this one because it would be a lot of work :) Will start reviewing soon.

kainino0x:
> * State default values in generated docstring: **Shall we?**

I think this would be great.

> I see 2 options:
> 
> * **Option A** We create two different `WGPU_BUFFER_BINDING_LAYOUT_INIT` macros, one with this name and that sets `type` to `Uniform` and one internal that is used to initialize `WGPUBindGroupLayoutEntry`. Unclear how to unambiguously define this behavior.
> * **Option B** We set the default value for `type` to `BindingNotUsed`, which diverges slightly from the JS spec but there is probably only little case where one initializes a `WGPUBufferBindingLayout` object without it being part of a `WGPUBindGroupLayoutEntry`.

I have been planning to use Option B. We actually started doing this already in Dawn: https://dawn-review.googlesource.com/c/dawn/+/214854, here for `VertexBufferStepMode` which has the same problem.
We still apply the defaulting from `Undefined` -> `(default)` but that only happens if the user explicitly writes `Undefined`.

aside: In C++ it occurs to me we _could_ have the default values in `wgpu::BindGroupLayoutEntry` be `BindingNotUsed` but have the default values in e.g. `wgpu::BufferBindingLayout` be `Vertex` so that if you do `entry.buffer = {};` it does what it does in JS. But I think I prefer to simply have it required to be explicit in C/C++, unlike in JS. (Similarly I've determined I want the default in `wgpu::VertexBufferLayout` to be `VertexBufferNotUsed` so users can default-initialize or zero-initialize an array of `wgpu::VertexBufferLayout` and have it be "empty".)

> Another limitation spotted: `.chain` fields are not automatically initialized with their proper `SType`. This is related to the previous issue in that we should initialize the `WGPUChainedStruct` type differently depending on its parent struct.

Hmm, I hope we don't run into problems with the INIT design.

When would we need to initialize `WGPUChainedStruct` differently? I believe `sType` values should always be 1:1 with the struct types.

webgpu.h
```diff
@@ -2086,6 +2598,20 @@ typedef struct WGPUDepthStencilState {
     float depthBiasClamp;
 } WGPUDepthStencilState WGPU_STRUCTURE_ATTRIBUTE;
 
+#define WGPU_DEPTH_STENCIL_STATE_INIT _wgpu_MAKE_INIT_STRUCT(WGPUDepthStencilState, { \
+    /*.nextInChain=*/NULL _wgpu_COMMA \
+    /*.format=*/WGPUTextureFormat_Undefined _wgpu_COMMA \
+    /*.depthWriteEnabled=*/{} _wgpu_COMMA \
```
kainino0x, line 2604: This needs to default to `WGPUOptionalBool_Undefined` because that isn't the zero-value. (We do this weird thing so that C++ and stdbool.h `bool`s will cast correctly to `WGPUOptionalBool`.)

webgpu.h
```diff
@@ -67,6 +67,9 @@
 #else
 #  define _wgpu_MAKE_INIT_STRUCT(type, value) value
 #endif
+#define _wgpu_COMMA ,
```
kainino0x, line 70: Duplicated definition

webgpu.h
```diff
@@ -67,6 +67,9 @@
...
+#define _wgpu_TRUE 1
+#define _wgpu_FALSE 0
```
kainino0x, line 72: nit: I think I prefer to just inline the `0` so we don't have as many "pinky promise private" macros.

(same for `_wgpu_TRUE` but it's unused)

gen/cheader.tmpl
```diff
@@ -294,6 +297,16 @@ typedef struct WGPU{{.Name | PascalCase}}CallbackInfo{{$.ExtSuffix}} {
     WGPU_NULLABLE void* userdata1;
     WGPU_NULLABLE void* userdata2;
 } WGPU{{.Name | PascalCase}}CallbackInfo{{$.ExtSuffix}} WGPU_STRUCTURE_ATTRIBUTE;
+
+#define WGPU_{{.Name | ConstantCase}}_CALLBACK_INFO{{$.ExtSuffix}}_INIT _wgpu_MAKE_INIT_STRUCT(WGPU{{.Name | PascalCase}}CallbackInfo{{$.ExtSuffix}}, { \
+    /*.nextInChain=*/NULL _wgpu_COMMA \
+{{-   if eq .Style "callback_mode" }}
+    /*.mode=*/WGPUCallbackMode_AllowProcessEvents _wgpu_COMMA \
```
kainino0x, line 304: Default should be `WaitAnyOnly` (that's the most "basic" - allows the fewest behaviors)...
or it should be `Undefined` so the caller has to set it explicitly.

webgpu.h
```diff
@@ -2173,6 +2755,14 @@ typedef struct WGPUVertexBufferLayout {
     WGPUVertexAttribute const * attributes;
 } WGPUVertexBufferLayout WGPU_STRUCTURE_ATTRIBUTE;
 
+#define WGPU_VERTEX_BUFFER_LAYOUT_INIT _wgpu_MAKE_INIT_STRUCT(WGPUVertexBufferLayout, { \
+    /*.nextInChain=*/NULL _wgpu_COMMA \
+    /*.stepMode=*/WGPUVertexStepMode_Vertex _wgpu_COMMA \
```
kainino0x, line 2760: as noted in the other comment, I think the default here should be `VertexBufferNotUsed`

kainino0x (commented):
I did not review thoroughly yet but it is looking quite good! Just a few small notes for things I saw so far.

BTW I filed #423 as a possible (noncritical) followup.

gen/cheader.tmpl
kainino0x: Dawn currently defaults to `WaitAnyOnly`.

eliemichel: I agree with `WaitAnyOnly`, at least for didactic purpose it is good to have a non-undefined mode by default.

webgpu.h
eliemichel: Ok! I added this directly in the generator, rather than manually specifying a default value to "undefined" for all use cases of optional bool, so that this rule is strongly enforced

eliemichel:
> I have been planning to use Option B.

Ok! I've applied it for all occurences of "NotUsed" in enum values, assuming it encompasses all such cases (namely vertex step mode and the 3 binding layouts).

> Hmm, I hope we don't run into problems with the INIT design.

I think that's fine actually, I could use `_wgpu_MAKE_INIT_STRUCT` inline

>> State default values in generated docstring: Shall we?
>
> I think this would be great.

Added them!

webgpu.h
kainino0x, line 1: I totally forgot to mention this before - we will need to add compile tests to `main.inl`. Like so (we can fix the TODO later):

```c
// TODO: Would be nice to autogenerate this so we don't forget to test any new structs.
{ WGPUBufferMapCallbackInfo x = WGPU_BUFFER_MAP_CALLBACK_INFO_INIT; }
{ WGPUCompilationInfoCallbackInfo x = WGPU_COMPILATION_INFO_CALLBACK_INFO_INIT; }
{ WGPUCreateComputePipelineAsyncCallbackInfo x = WGPU_CREATE_COMPUTE_PIPELINE_ASYNC_CALLBACK_INFO_INIT; }
{ WGPUCreateRenderPipelineAsyncCallbackInfo x = WGPU_CREATE_RENDER_PIPELINE_ASYNC_CALLBACK_INFO_INIT; }
{ WGPUDeviceLostCallbackInfo x = WGPU_DEVICE_LOST_CALLBACK_INFO_INIT; }
{ WGPUPopErrorScopeCallbackInfo x = WGPU_POP_ERROR_SCOPE_CALLBACK_INFO_INIT; }
{ WGPUQueueWorkDoneCallbackInfo x = WGPU_QUEUE_WORK_DONE_CALLBACK_INFO_INIT; }
{ WGPURequestAdapterCallbackInfo x = WGPU_REQUEST_ADAPTER_CALLBACK_INFO_INIT; }
{ WGPURequestDeviceCallbackInfo x = WGPU_REQUEST_DEVICE_CALLBACK_INFO_INIT; }
{ WGPUUncapturedErrorCallbackInfo x = WGPU_UNCAPTURED_ERROR_CALLBACK_INFO_INIT; }
{ WGPUAdapterInfo x = WGPU_ADAPTER_INFO_INIT; }
{ WGPUBindGroupEntry x = WGPU_BIND_GROUP_ENTRY_INIT; }
{ WGPUBlendComponent x = WGPU_BLEND_COMPONENT_INIT; }
{ WGPUBufferBindingLayout x = WGPU_BUFFER_BINDING_LAYOUT_INIT; }
{ WGPUBufferDescriptor x = WGPU_BUFFER_DESCRIPTOR_INIT; }
{ WGPUColor x = WGPU_COLOR_INIT; }
{ WGPUCommandBufferDescriptor x = WGPU_COMMAND_BUFFER_DESCRIPTOR_INIT; }
{ WGPUCommandEncoderDescriptor x = WGPU_COMMAND_ENCODER_DESCRIPTOR_INIT; }
{ WGPUCompilationMessage x = WGPU_COMPILATION_MESSAGE_INIT; }
{ WGPUComputePassTimestampWrites x = WGPU_COMPUTE_PASS_TIMESTAMP_WRITES_INIT; }
{ WGPUConstantEntry x = WGPU_CONSTANT_ENTRY_INIT; }
{ WGPUFuture x = WGPU_FUTURE_INIT; }
{ WGPUInstanceCapabilities x = WGPU_INSTANCE_CAPABILITIES_INIT; }
{ WGPULimits x = WGPU_LIMITS_INIT; }
{ WGPUMultisampleState x = WGPU_MULTISAMPLE_STATE_INIT; }
{ WGPUPipelineLayoutDescriptor x = WGPU_PIPELINE_LAYOUT_DESCRIPTOR_INIT; }
{ WGPUPrimitiveState x = WGPU_PRIMITIVE_STATE_INIT; }
{ WGPUQuerySetDescriptor x = WGPU_QUERY_SET_DESCRIPTOR_INIT; }
{ WGPUQueueDescriptor x = WGPU_QUEUE_DESCRIPTOR_INIT; }
{ WGPURenderBundleDescriptor x = WGPU_RENDER_BUNDLE_DESCRIPTOR_INIT; }
{ WGPURenderBundleEncoderDescriptor x = WGPU_RENDER_BUNDLE_ENCODER_DESCRIPTOR_INIT; }
{ WGPURenderPassDepthStencilAttachment x = WGPU_RENDER_PASS_DEPTH_STENCIL_ATTACHMENT_INIT; }
{ WGPURenderPassMaxDrawCount x = WGPU_RENDER_PASS_MAX_DRAW_COUNT_INIT; }
{ WGPURenderPassTimestampWrites x = WGPU_RENDER_PASS_TIMESTAMP_WRITES_INIT; }
{ WGPURequestAdapterOptions x = WGPU_REQUEST_ADAPTER_OPTIONS_INIT; }
{ WGPUSamplerBindingLayout x = WGPU_SAMPLER_BINDING_LAYOUT_INIT; }
{ WGPUSamplerDescriptor x = WGPU_SAMPLER_DESCRIPTOR_INIT; }
{ WGPUShaderModuleDescriptor x = WGPU_SHADER_MODULE_DESCRIPTOR_INIT; }
{ WGPUShaderSourceSPIRV x = WGPU_SHADER_SOURCE_SPIRV_INIT; }
{ WGPUShaderSourceWGSL x = WGPU_SHADER_SOURCE_WGSL_INIT; }
{ WGPUStencilFaceState x = WGPU_STENCIL_FACE_STATE_INIT; }
{ WGPUStorageTextureBindingLayout x = WGPU_STORAGE_TEXTURE_BINDING_LAYOUT_INIT; }
{ WGPUSupportedFeatures x = WGPU_SUPPORTED_FEATURES_INIT; }
{ WGPUSupportedWGSLLanguageFeatures x = WGPU_SUPPORTED_WGSL_LANGUAGE_FEATURES_INIT; }
{ WGPUSurfaceCapabilities x = WGPU_SURFACE_CAPABILITIES_INIT; }
{ WGPUSurfaceConfiguration x = WGPU_SURFACE_CONFIGURATION_INIT; }
{ WGPUSurfaceDescriptor x = WGPU_SURFACE_DESCRIPTOR_INIT; }
{ WGPUSurfaceSourceAndroidNativeWindow x = WGPU_SURFACE_SOURCE_ANDROID_NATIVE_WINDOW_INIT; }
{ WGPUSurfaceSourceMetalLayer x = WGPU_SURFACE_SOURCE_METAL_LAYER_INIT; }
{ WGPUSurfaceSourceWaylandSurface x = WGPU_SURFACE_SOURCE_WAYLAND_SURFACE_INIT; }
{ WGPUSurfaceSourceWindowsHWND x = WGPU_SURFACE_SOURCE_WINDOWS_HWND_INIT; }
{ WGPUSurfaceSourceXCBWindow x = WGPU_SURFACE_SOURCE_XCB_WINDOW_INIT; }
{ WGPUSurfaceSourceXlibWindow x = WGPU_SURFACE_SOURCE_XLIB_WINDOW_INIT; }
{ WGPUSurfaceTexture x = WGPU_SURFACE_TEXTURE_INIT; }
{ WGPUTexelCopyBufferLayout x = WGPU_TEXEL_COPY_BUFFER_LAYOUT_INIT; }
{ WGPUTextureBindingLayout x = WGPU_TEXTURE_BINDING_LAYOUT_INIT; }
{ WGPUTextureViewDescriptor x = WGPU_TEXTURE_VIEW_DESCRIPTOR_INIT; }
{ WGPUVertexAttribute x = WGPU_VERTEX_ATTRIBUTE_INIT; }
{ WGPUBindGroupDescriptor x = WGPU_BIND_GROUP_DESCRIPTOR_INIT; }
{ WGPUBindGroupLayoutEntry x = WGPU_BIND_GROUP_LAYOUT_ENTRY_INIT; }
{ WGPUBlendState x = WGPU_BLEND_STATE_INIT; }
{ WGPUCompilationInfo x = WGPU_COMPILATION_INFO_INIT; }
{ WGPUComputePassDescriptor x = WGPU_COMPUTE_PASS_DESCRIPTOR_INIT; }
{ WGPUComputeStage x = WGPU_COMPUTE_STAGE_INIT; }
{ WGPUDepthStencilState x = WGPU_DEPTH_STENCIL_STATE_INIT; }
{ WGPUDeviceDescriptor x = WGPU_DEVICE_DESCRIPTOR_INIT; }
{ WGPUFutureWaitInfo x = WGPU_FUTURE_WAIT_INFO_INIT; }
{ WGPUInstanceDescriptor x = WGPU_INSTANCE_DESCRIPTOR_INIT; }
{ WGPURenderPassColorAttachment x = WGPU_RENDER_PASS_COLOR_ATTACHMENT_INIT; }
{ WGPUTexelCopyBufferInfo x = WGPU_TEXEL_COPY_BUFFER_INFO_INIT; }
{ WGPUTexelCopyTextureInfo x = WGPU_TEXEL_COPY_TEXTURE_INFO_INIT; }
{ WGPUTextureDescriptor x = WGPU_TEXTURE_DESCRIPTOR_INIT; }
{ WGPUVertexBufferLayout x = WGPU_VERTEX_BUFFER_LAYOUT_INIT; }
{ WGPUBindGroupLayoutDescriptor x = WGPU_BIND_GROUP_LAYOUT_DESCRIPTOR_INIT; }
{ WGPUColorTargetState x = WGPU_COLOR_TARGET_STATE_INIT; }
{ WGPUComputePipelineDescriptor x = WGPU_COMPUTE_PIPELINE_DESCRIPTOR_INIT; }
{ WGPURenderPassDescriptor x = WGPU_RENDER_PASS_DESCRIPTOR_INIT; }
{ WGPUVertexState x = WGPU_VERTEX_STATE_INIT; }
{ WGPUFragmentState x = WGPU_FRAGMENT_STATE_INIT; }
{ WGPURenderPipelineDescriptor x = WGPU_RENDER_PIPELINE_DESCRIPTOR_INIT; }
```

kainino0x (commented):
> > Hmm, I hope we don't run into problems with the INIT design.
> 
> I think that's fine actually, I could use `_wgpu_MAKE_INIT_STRUCT` inline

Ah, that's true!

> > > State default values in generated docstring: Shall we?
> > 
> > 
> > I think this would be great.
> 
> Added them! **NB:** The default value is given for all member, including when it is not explicitly specified in `webgpu.yml` due to the zero-value being the expected default. I think it is good to make things explicit but we could also leave them unspecified and globally say that unspecified means 0.

Will review but that seems good to me.

webgpu.h
kainino0x: Oh, this will only work with a `git merge main` because I just landed #425

webgpu.yml
```diff
@@ -2971,6 +3092,7 @@ structs:
           indicates a "hole" in the parent @ref WGPUVertexState `buffers` array:
           the pipeline does not use a vertex buffer at this `location`.
         type: enum.vertex_step_mode
+        default: vertex_buffer_bot_used
```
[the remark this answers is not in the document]
eliemichel: Oops! Fixed

eliemichel:
Adding test revealed a lot of small things (which I should have tested indeed).

I notably took the questionable liberty of adding (back?) an Undefined entry to the `VertexFormat` enum.

Another note: I realize the generation of default values in `g.DefaultValue` is not robost to non-empty value of `$.ExtSuffix`: is there a way to access this value on the gen.go side?

kainino0x:
> I notably took the questionable liberty of adding (back?) an Undefined entry to the `VertexFormat` enum.

Hm, I don't feel strongly either way. Alternative would be to zero-init those fields with `{}` which is what Dawn does, but that's also less self-documenting. There has been some back and forth on whether we should name the mostly-unused 0 values.

> Another note: I realize the generation of default values in `g.DefaultValue` is not robost to non-empty value of `$.ExtSuffix`: is there a way to access this value on the gen.go side?

Honestly I have no idea. If it doesn't cause problems for generation right now, it's fine and we'll just deal with it later.

eliemichel:
> Hm, I don't feel strongly either way. Alternative would be to zero-init those fields with {} which is what Dawn does, but that's also less self-documenting. There has been some back and forth on whether we should name the mostly-unused 0 values.

The issue occurs in some other cases as well actually, because we generate INIT macros for structs that are only used to return info (e.g., about the adapter) for which it really doesn't make sense to add an `Undefined`.

However, as you can see above CI complains about using `{}` to zero-initialize things in C (it is considered a GNU extension). IIRC it is an annoying case where what C++ writes `{}` is actually what C write `{0}` (and notably `{0}` in C initializes all fields to 0 while in C++ it only initializes the first member to 0...). Anyways, I thus generate initializers of the form  `(WGPUFoo)0` when a member has type `enum WGPUFoo` and no default value is specified in `webgpu.yml`.

Since I had to add this, I'll roll back the (re)introduction of `Undefined` in `VertexFormat`.

> Honestly I have no idea. If it doesn't cause problems for generation right now, it's fine and we'll just deal with it later.
It does not cause any issue for now

eliemichel:
Mmh now C++ complains about C-style cast... I'll try with a `_wgpu_ZERO_INIT` macro then

eliemichel:
Now It is smart enough to realize that 0 is not a valid value, better to default to the first non-null entry of the enum in the end.

webgpu.h
```diff
@@ -1373,891 +1442,2452 @@ typedef struct WGPURequestDeviceCallbackInfo {
...
 } WGPUAdapterInfo WGPU_STRUCTURE_ATTRIBUTE;
 
+#define WGPU_ADAPTER_INFO_INIT _wgpu_MAKE_INIT_STRUCT(WGPUAdapterInfo, { \
+    /*.nextInChain=*/NULL _wgpu_COMMA \
+    /*.vendor=*/WGPU_STRING_VIEW_INIT _wgpu_COMMA \
+    /*.architecture=*/WGPU_STRING_VIEW_INIT _wgpu_COMMA \
+    /*.device=*/WGPU_STRING_VIEW_INIT _wgpu_COMMA \
+    /*.description=*/WGPU_STRING_VIEW_INIT _wgpu_COMMA \
+    /*.backendType=*/WGPUBackendType_Undefined _wgpu_COMMA \
+    /*.adapterType=*/WGPUAdapterType_Unknown _wgpu_COMMA \
```
kainino0x, line 1521: I think this should default to a zero value. `Unknown` is a value that comes back from `GetInfo` and tells you the adapter type is unknown. (Though it could reasonably be argued that zero and `Unknown` should be the same value.)

That said I'm also not really sure `WGPUAdapterInfo` should even _have_ an `INIT` macro (since it's output-only). Dawn has it but maybe we should remove it? (I will deal with this in a followup.)

webgpu.h
```diff
@@ -1373,891 +1442,2452 @@ typedef struct WGPURequestDeviceCallbackInfo {
...
      * If this is not possible, the request returns null.
+     *
+     * Defaults to NULL
      */
     WGPU_NULLABLE WGPUSurface compatibleSurface;
 } WGPURequestAdapterOptions WGPU_STRUCTURE_ATTRIBUTE;
 
+#define WGPU_REQUEST_ADAPTER_OPTIONS_INIT _wgpu_MAKE_INIT_STRUCT(WGPURequestAdapterOptions, { \
+    /*.nextInChain=*/NULL _wgpu_COMMA \
+    /*.featureLevel=*/WGPUFeatureLevel_Core _wgpu_COMMA \
```
kainino0x, line 2441: This default is still TBD, for now it should be 0.

webgpu.h
```diff
@@ -1373,891 +1442,2452 @@ typedef struct WGPURequestDeviceCallbackInfo {
...
+#define WGPU_SURFACE_CONFIGURATION_INIT _wgpu_MAKE_INIT_STRUCT(WGPUSurfaceConfiguration, { \
+    /*.nextInChain=*/NULL _wgpu_COMMA \
+    /*.device=*/NULL _wgpu_COMMA \
+    /*.format=*/WGPUTextureFormat_Undefined _wgpu_COMMA \
+    /*.usage=*/WGPUTextureUsage_RenderAttachment _wgpu_COMMA \
+    /*.width=*/0 _wgpu_COMMA \
+    /*.height=*/0 _wgpu_COMMA \
+    /*.viewFormatCount=*/0 _wgpu_COMMA \
+    /*.viewFormats=*/NULL _wgpu_COMMA \
+    /*.alphaMode=*/WGPUCompositeAlphaMode_Opaque _wgpu_COMMA \
```
kainino0x, line 2800: This should default to "Auto" (it's slightly different from the JS API)

webgpu.h
```diff
@@ -1373,891 +1442,2452 @@ typedef struct WGPURequestDeviceCallbackInfo {
...
     uint32_t bytesPerRow;
+    /**
+     * Defaults to 0
+     */
     uint32_t rowsPerImage;
 } WGPUTexelCopyBufferLayout WGPU_STRUCTURE_ATTRIBUTE;
 
+#define WGPU_TEXEL_COPY_BUFFER_LAYOUT_INIT _wgpu_MAKE_INIT_STRUCT(WGPUTexelCopyBufferLayout, { \
+    /*.offset=*/0 _wgpu_COMMA \
+    /*.bytesPerRow=*/0 _wgpu_COMMA \
```
kainino0x, line 3044: Should be `WGPU_COPY_STRIDE_UNDEFINED`

webgpu.h
```diff
@@ -1373,891 +1442,2452 @@ typedef struct WGPURequestDeviceCallbackInfo {
...
+    /*.rowsPerImage=*/0 _wgpu_COMMA \
```
kainino0x, line 3045: Should be `WGPU_COPY_STRIDE_UNDEFINED`

webgpu.h
```diff
@@ -1373,891 +1442,2452 @@ typedef struct WGPURequestDeviceCallbackInfo {
...
     WGPUTextureUsage usage;
 } WGPUTextureViewDescriptor WGPU_STRUCTURE_ATTRIBUTE;
 
+#define WGPU_TEXTURE_VIEW_DESCRIPTOR_INIT _wgpu_MAKE_INIT_STRUCT(WGPUTextureViewDescriptor, { \
+    /*.nextInChain=*/NULL _wgpu_COMMA \
+    /*.label=*/WGPU_STRING_VIEW_INIT _wgpu_COMMA \
+    /*.format=*/WGPUTextureFormat_Undefined _wgpu_COMMA \
+    /*.dimension=*/WGPUTextureViewDimension_Undefined _wgpu_COMMA \
+    /*.baseMipLevel=*/0 _wgpu_COMMA \
+    /*.mipLevelCount=*/0 _wgpu_COMMA \
```
kainino0x, line 3125: Should be `WGPU_MIP_LEVEL_COUNT_UNDEFINED`

webgpu.h
```diff
@@ -1373,891 +1442,2452 @@ typedef struct WGPURequestDeviceCallbackInfo {
...
+    /*.baseArrayLayer=*/0 _wgpu_COMMA \
+    /*.arrayLayerCount=*/0 _wgpu_COMMA \
```
kainino0x, line 3127: Should be `WGPU_ARRAY_LAYER_COUNT_UNDEFINED`

webgpu.h
```diff
@@ -1373,891 +1442,2452 @@ typedef struct WGPURequestDeviceCallbackInfo {
...
+     * Defaults to @ref WGPUColorWriteMask_None
+     */
     WGPUColorWriteMask writeMask;
 } WGPUColorTargetState WGPU_STRUCTURE_ATTRIBUTE;
 
+#define WGPU_COLOR_TARGET_STATE_INIT _wgpu_MAKE_INIT_STRUCT(WGPUColorTargetState, { \
+    /*.nextInChain=*/NULL _wgpu_COMMA \
+    /*.format=*/WGPUTextureFormat_Undefined _wgpu_COMMA \
+    /*.blend=*/NULL _wgpu_COMMA \
+    /*.writeMask=*/WGPUColorWriteMask_None _wgpu_COMMA \
```
kainino0x, line 3696: Should be `WGPUColorWriteMask_All`

webgpu.h
kainino0x, line 1: I went through all the differences between this and Dawn and added comments for things that should change in this PR. Here are all the other differences that don't need to change in this PR:

- Places where Dawn is behind or the definition order is different
- Spelling of default values differs (but value is the same so it doesn't matter except for the C vs C++ differences you were already investigating):
  - `0.0f` vs `0.f`
  - In many (but not all) places it uses `{}` instead of a named value
  - In some places it uses a named value instead of `{}`
  - It doesn't use `_wgpu_MAKE_INIT_STRUCT` when initializing `.chain`, it just uses a bare `{NULL _wgpu_COMMA WGPUSType_RenderPassMaxDrawCount}`
- (I was wrong about this before) the new `CallbackInfo.mode` fields init to `{}` instead of `WaitAnyOnly` (some outdated duplicates are `WaitAnyOnly`)
- It incorrectly comments `/*.next=*/` as `/*.nextInChain*/`
- It incorrectly uses `false` instead of `0` (since it doesn't include `stdbool.h`)
- It incorrectly uses `nullptr` instead of `NULL` (in some places where it doesn't use `{}`)

webgpu.h
eliemichel: Thing is, there is no zero-value for type `WGPUAdapterType`, and forcing a value of 0 has not been a success in my attempts above

kainino0x (commented):
Somehow I submitted some of the comments before others. Here's the rest

webgpu.h
kainino0x: Oh and `0xFFFFFFFF` vs `4294967295`

eliemichel: Thanks for going through all this!

I think the key remaining difference now is that we cannot use `{}`, as far as I can tell (as mentionned above, it is not standard C) so the generator now just goes and figure the very first enum entry and use it as default. This also means we need to decide on a default value for `CallbackInfo.mode`: `WaitAnyOnly` was fine with me but if we rather want to force the user to choose one, we need to reintroduce an `Undefined` value to initialize to (otherwise C++ complains)

kainino0x: I pulled your latest code and I'm experimenting. I see that `0` or `{0}` hits `-Wassign-enum` in clang (from `-Weverything`). Trying to determine if that is just a "maybe you made a typo" warning or a "you're doing UB" warning (I've researched this before but didn't remember). This answer https://stackoverflow.com/a/33814842 says basically that the underlying type will be a signed or unsigned integer type of arbitrary bit length large enough to hold all of the values, which means this is _safe_ because 0 is always representable.

`(WGPUCompilationMessageType)0` works in C but of course hits `-Wold-style-cast` in C++. We _could_ conditionally choose between that for C and `WGPUCompilationMessageType(0)` for C++ (assuming MSVC is OK with it).

Or we could just give names to all the enums' zero values. IMO this seems more reasonable. :)

kainino0x: Tried it out, that seems to work fine: https://github.com/kainino0x/webgpu-headers/actions/runs/11945415786
So perhaps we can land the PR with that and we'll follow up on it in #427.

webgpu.h
kainino0x: Changed this to `_wgpu_ENUM_ZERO_INIT(WGPUAdapterType)`

webgpu.h
kainino0x: Changed to `_wgpu_ENUM_ZERO_INIT(WGPUFeatureLevel)`

kainino0x (approved):
I addressed things using `_wgpu_ENUM_ZERO_INIT`, finally reviewed the Go code, and did some minor nits to the documentation. Everything looks great now, if there was anything else you wanted to change feel free to open more PR(s) - I'm going to land this now!

Issue #428, opened 2024-11-22
supports 'subgroups' feature


dneto0:
This issue tracks adding support for the 'subgroups' optional feature in WebGPU.

It has GPUFeatureName "subgroups"
- Original issue:  https://github.com/gpuweb/gpuweb/issues/4306
- PR https://github.com/gpuweb/gpuweb/pull/4963

dneto0:
The subgroups feature has landed in the JS API, via https://github.com/gpuweb/gpuweb/pull/4963 

The native API needs a way to get the two new properties on AdapterInfo:

* `subgroupMinSize`, see https://gpuweb.github.io/gpuweb/#dom-gpuadapterinfo-subgroupminsize
* `subgroupMaxSize`, see https://gpuweb.github.io/gpuweb/#dom-gpuadapterinfo-subgroupminsize


Currently this is modeled in Dawn as an `WGPUAdapterPropertiesSubgroups` that extends WGPUAdapterInfo.
(See [dawn.json](https://dawn.googlesource.com/dawn/+/refs/heads/chromium/6989/src/dawn/dawn.json#1545))

```c
// Can be chained in WGPUAdapterInfo 
typedef struct WGPUAdapterPropertiesSubgroups {
    WGPUChainedStruct chain;
    uint32_t subgroupMinSize;
    uint32_t subgroupMaxSize;
} WGPUAdapterPropertiesSubgroups WGPU_STRUCTURE_ATTRIBUTE;
```

If it goes into v1, then we can simplify all this an put it directly into `WGPUAdapterInfo`

dneto0:
Comment https://github.com/webgpu-native/webgpu-headers/pull/429#discussion_r1854872246 suggested renaming the struct to WGPUAdapterInfoSubgroups.  But the new struct won't be needed if we pull it into v1 and put the properties into WGPUAdapterInfo.

PR #429, opened 2024-11-22
Support "subgroups" optional feature

dneto0:
struct adapter_properties_subgroups is chained out from adapter_info, and reports minimum and maximum supported subgroup sizes for the adapter.

Issue: #428

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -185,6 +185,7 @@ typedef struct WGPUTextureViewImpl* WGPUTextureView WGPU_OBJECT_ATTRIBUTE;
 /** @} */
 // Structure forward declarations
 struct WGPUAdapterInfo;
+struct WGPUAdapterPropertiesSubgroups;
 struct WGPUBindGroupEntry;
 struct WGPUBlendComponent;
 struct WGPUBufferBindingLayout;
@@ -742,6 +743,7 @@ typedef enum WGPUSType {
     WGPUSType_SurfaceSourceWaylandSurface = 0x00000007,
     WGPUSType_SurfaceSourceAndroidNativeWindow = 0x00000008,
     WGPUSType_SurfaceSourceXCBWindow = 0x00000009,
+    WGPUSType_AdapterPropertiesSubgroups = 0x0000000A,
     WGPUSType_Force32 = 0x7FFFFFFF
 } WGPUSType WGPU_ENUM_ATTRIBUTE;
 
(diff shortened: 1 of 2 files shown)


dneto0:
cc: @kainino0x 

I took a shot at this.    
The corresponding Dawn change landed via https://dawn-review.googlesource.com/c/dawn/+/216116

webgpu.h
```diff
@@ -1566,6 +1568,37 @@ typedef struct WGPUAdapterInfo {
...
+ * Provides bounds on the sizes of subgroups supported by the adapter.
+ *
+ * Default values can be set using @ref WGPU_ADAPTER_PROPERTIES_SUBGROUPS_INIT as initializer.
+ */
+typedef struct WGPUAdapterPropertiesSubgroups {
+    WGPUChainedStructOut * nextInChain;
+    /**
+     * If the "subgroups" feature is supported, the minimum supported subgroup size for the adapter.
+     *
+     * Defaults to `0`.
```
dneto0, line 1582: Not sure if this should be UNDEFINED

webgpu.h
```diff
@@ -1566,6 +1568,37 @@ typedef struct WGPUAdapterInfo {
     /*.deviceID=*/0 _wgpu_COMMA \
 })
 
+/**
+ * Enabled by the "subgroups" feature.
+ * Provides bounds on the sizes of subgroups supported by the adapter.
```
eliemichel, line 1573: Would be worth precising what base structure this can be chained to (is it `WGPUAdapterInfo`?).

webgpu.h
```diff
@@ -1566,6 +1568,37 @@ typedef struct WGPUAdapterInfo {
...
+ *
+ * Default values can be set using @ref WGPU_ADAPTER_PROPERTIES_SUBGROUPS_INIT as initializer.
+ */
+typedef struct WGPUAdapterPropertiesSubgroups {
```
eliemichel, line 1577: If this extends `WGPUAdapterInfo`, it may better be named `WGPUAdapterInfoSubgroups`?

webgpu.h
eliemichel: Maybe `WGPU_LIMIT_U32_UNDEFINED` (i.e., `default: constant.limit_u32_undefined` in `webgpu.yml`?

kainino0x: This is only an output (not an input) so there's no special handling of UNDEFINED, and it's not really a limit anyway. Since there's no real meaning to either value I think the INIT struct should just do zero-init.

(BTW I am going to change the wording of "Defaults to" because I was talking to someone yesterday and we realized that that sounds like the defaulting occurs _inside_ the API, but what we mean is the INIT value.)

kainino0x (commented):
No further comments, but I don't think we should add this until it's stable in the JS API; otherwise we could need to change it after we stabilize `webgpu.h` 1.0.

webgpu.h
alan-baker, line 1582: The spec PR makes the default values (even for unsupported) for min and max 4 and 128.

dneto0:
Converted to Draft. Per Kai's suggestion, this shouldn't proceed until the feature lands in the JS API.

kainino0x:
Closing in favor of #509 which adds the new info fields directly instead of adding an extension struct.

Issue #427, opened 2024-11-21
INIT macro follow-up tasks


kainino0x:
Follow-ups collected from PR #422 (issue #158).

Checkbox = fixed or at least has a PR or issue about it.

- [x] #445
- [x] (EDIT: big one so I added it at the top) #446
  Maybe INIT macros should prefer `Undefined` over trivial defaults (e.g. `/*.dimension=*/WGPUTextureDimension_2D`), in case the JS API changes them from trivial (WebIDL) defaults to non-trivial defaults (non-breakingly for JS, in such a way that they depend on the values of other new members or something, but that would make the INIT macros differ slightly).
  (I am not sure if we ever discussed this before. I know we said that we should implement all of WebIDL's trivial defaulting in webgpu.h implementations though.)
  - Affects #424
- [x] Are we happy with `#define WGPU_DEPTH_CLEAR_VALUE_UNDEFINED (NAN)` (requires including `math.h`). Note the JS API disallows NaN (and Infinity) at the WebIDL level
  - [x] We should document the sentinel (any `isnan()` value I think, but maybe it should be `!isfinite()` or maybe we should disallow infinities as a validation rule, or clamp them to large finite numbers).
    #467
    #466
- [x] Make sure we have the default we want for `WGPUCallbackMode` (I can't remember if we wanted to default it or not. Does it even matter since people are probably usually going to use these inline anyway? Maybe they don't need INIT macros?)
    #471
- [x] Should we have INIT macros for output-only structs? (like `WGPUAdapterInfo`)
  - Maybe we can guarantee these are all zero-init?
- [x] Should we have INIT macros when they are equivalent to zero-init, and therefore not really necessary? (see also #423)
  - Could be auto-detected or just explicit in the YAML
- [x] Should we name all of the enums' zero values so that the INIT values can be more self-documenting (instead of using `_wgpu_ENUM_ZERO_INIT()` or whatever)?
  - (we also initialize booleans to `0` but there isn't really a better option for those)
- [x] "Defaults to" sounds kind of like the defaulting occurs inside the API. This is not always true (though it is true for some fields). What we are actually documenting is what the value of the field is when you use the INIT macro.
  #436
- [x] Make sure INIT macros are properly postfixed when defining extension (see https://github.com/webgpu-native/webgpu-headers/pull/453)

eliemichel:
> Should we have INIT macros for output-only structs? (like WGPUAdapterInfo)

[...]

PR #436, opened 2024-11-23
Clarify "Defaults to" generated doc

kainino0x:
I was too lazy to have this `@ref` the actual INIT macro, but I could if you think it's better. (It would be even more verbose though!)

Issue: #427


eliemichel (approved):
Nah this is great, the INIT macro is already @ref'd in the struct's docstring so it should be visually far from this in any generated doc!

PR #439, opened 2024-11-26
Proposal: remove WGPUVertexStepMode_VertexBufferNotUsed, with 1:1 mapping to JS

### LINKED ISSUE #432: Use attributeCount=0 as sentinel instead of VertexBufferNotUsed?
See discussion here: https://github.com/webgpu-native/webgpu-headers/pull/424#issuecomment-2495245707

> * to answer question: **No, don't need `Undefined`**
> * **KN to investigate using attributeCount (what does the JS API do - and could it change to treat `[]` as a hole?)**

But if we do remove VertexBufferNotUsed=0 then there will still be a reserved 0 for Undefined and then we should use it for defaulting since that's what we do everywhere else

kainino0x:
I realized that there was actually no good way to paper over the difference between the C API we had proposed, and the current JS API.

This draft finds a compromise that does work: it distinguishes between `WGPUVertexStepMode_Undefined` and `WGPUVertexStepMode_Vertex` _if_ `.attributeCount == 0`.

In other words, the sentinel is
`{.stepMode=Undefined, .attributeCount=0}` instead of just
`{.stepMode=VertexBufferNotUsed}`, or just
`{.attributeCount=0}`.

Notably, this is flexible to whatever JS does in https://github.com/gpuweb/gpuweb/issues/4999
since it can still represent both `null` and `{ stepMode: whatever, attributes: [] }`.

Fixes #432

### DIFF
--- a/doc/articles/SentinelValues.md
+++ b/doc/articles/SentinelValues.md
@@ -36,7 +36,7 @@ WebIDL's more flexible typing:
 - \ref WGPUFuture has a special null value
 - Special cases to indicate the parent struct is null, avoiding extra layers of
   pointers just for nullability:
-    - \ref WGPUVertexBufferLayout::stepMode = \ref WGPUVertexStepMode_VertexBufferNotUsed
+    - \ref WGPUVertexBufferLayout::stepMode = \ref WGPUVertexStepMode_Undefined with \ref WGPUVertexBufferLayout::attributeCount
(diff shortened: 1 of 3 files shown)


kainino0x:
I think this is sound and reasonable and would like to merge it **tentatively** (with #432 still on the agenda to discuss). PTAL

lokokung (approved):

PR #441, opened 2024-11-26
Make Doxygen warnings strict

kainino0x:
- I finally figured out how to fail on warnings, but not warn on undocumented things.
- Fixed all the errors
- Found out that newer versions of Doxygen don't check for typos in `` `::symbol` `` style refs so removed all of them
- And figured out workarounds for problems where `@ref symbol:` and similar things with punctuation would not work


kainino0x:
Feel free to review but it's just documentation fixes so I went ahead and landed it (also if any problems come up they can be reverted easily)

Issue #430, opened 2024-11-23
Deduplicate TimestampWrites structs


kainino0x:
And put it back behind a pointer like it was before.

Per https://github.com/webgpu-native/webgpu-headers/pull/404#issuecomment-2495228026

PR #442, opened 2024-11-26
Move nullability back outside of TimestampWrites

kainino0x:
Issue: #430, partially reverts #404


kainino0x:
Pretty trivial revert, landing now so I can open another PR on top of it.

PR #445, opened 2024-11-27
Fix default for buffer binding type

kainino0x:
- Fix default for buffer binding type
- Remove defaults that weren't doing anything
- Validate out defaults that don't do anything
  (I wanted to do this with the schema but it was breaking the unevaluatedProperties check so I just gave up and did this)

Issue #427

### DIFF
--- a/gen/gen.go
+++ b/gen/gen.go
@@ -492,11 +492,17 @@ func (g *Generator) DefaultValue(member ParameterType, isDocString bool) string 
 			return s
 		}
 	}
+
 	switch {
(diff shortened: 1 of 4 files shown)


eliemichel (approved):
Looks goot to me!

(Also, IDK where to thank you for that but thanks a lot for taking to the time to report WG meeting notes everywhere :) )

kainino0x:
Thanks, and no problem! I just take notes in a random Google doc so I have to post them publicly somewhere :)

PR #446, opened 2024-11-27
Proposal: enums in INIT macros always use Undefined

kainino0x:
The actual default may be dependent on where the struct is used, or could be changed from a trivial default to a non-trivial default in the future.

The `Undefined` enum values (or `WGPUCompositeAlphaMode_Auto`) are defined to always do the defaulting inside the API regardless of whether it's a trivial default or not.

All enum fields (searched for `type: enum\..*\n\s*default: `) now use `Undefined`(/`Auto`) or `*NotUsed`. (except for #445)

I also added documentation to all the ones I changed, explaining the default. All of the others are either optional or required in the JS spec, so should be fairly clear.

I might turn the defaulting/optional/required docs into something more generated later, but for now this does what we need.

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -1639,15 +1639,15 @@ typedef struct WGPUBindGroupEntry {
  */
 typedef struct WGPUBlendComponent {
     /**
-     * The `INIT` macro sets this to @ref WGPUBlendOperation_Add.
+     * The `INIT` macro sets this to @ref WGPUBlendOperation_Undefined.
      */
     WGPUBlendOperation operation;
     /**
-     * The `INIT` macro sets this to @ref WGPUBlendFactor_One.
+     * The `INIT` macro sets this to @ref WGPUBlendFactor_Undefined.
      */
     WGPUBlendFactor srcFactor;
     /**
-     * The `INIT` macro sets this to @ref WGPUBlendFactor_Zero.
+     * The `INIT` macro sets this to @ref WGPUBlendFactor_Undefined.
      */
     WGPUBlendFactor dstFactor;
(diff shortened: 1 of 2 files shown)


eliemichel (commented):
If I understand correctly, setting e.g. `WGPUBlendComponent` to `Undefined` will have the same effect as setting it to `Add` (in the current stats of the API), so it is preferred to init it to `Undefined`.

I see the point of automatically handling future updates of the API, but on the other hand it makes the header much **less self-contained** (and knowing the default values is one of the main reasons why I always have to look up the docs).

I'm tempted to suggest we adopt a soft trade-off where the default value is in effect `Undefined` but the docstring explicitly tells what non-undefined value this corresponds to. We're back with the risk of doc being out of sync with the actual behavior but at least it solves the issue of expressing non-trivial defaulting scheme.

We could phrase the doc such that we invite one to check out a reliable source of info about where to find up-to-date information about the defaulting mechanism.

kainino0x:
@eliemichel Sorry - I _completely_ forgot to push my second commit before opening this PR (which I was referring to in my PR summary). It adds documentation for all of the defaulting. WDYT?

eliemichel (approved):
Solves my concern indeed!

kainino0x:
Landing tentatively - pretty sure this is what we'll settle on, but we'll `!discuss` to verify.

kainino0x:
Dec 9 meeting:

* KN: (summary)  
* **CF: Don't love it but seems fine**. Prefer not to lose documentation
* KN: Agree, I added documentation for all of these in the PR.

PR #449, opened 2024-11-28
Fixes for return type docs

### LINKED ISSUE #388: Generate docs for passed_with_ownership on return types
See https://github.com/webgpu-native/webgpu-headers/pull/384/files#r1826317247

kainino0x:
- Require return docs to be single-paragraph

  Verified that the bullets still work even though Markdown sorta treats
them as a separate paragraph. In the generated docs the bullets now end
up inside the `<dd>` for the return value, instead of outside.
- Generate docs for `passed_with_ownership` on return types

Fixes #388

### DIFF
--- a/gen/gen.go
+++ b/gen/gen.go
@@ -47,10 +47,10 @@ func (g *Generator) Gen(dst io.Writer) error {
 					}
 				}
 				for _, arg := range fn.Args {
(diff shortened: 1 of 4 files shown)


eliemichel (approved):
👌

PR #450, opened 2024-11-28
Proposal: Allow use of WGPU_BUFFER_BINDING_LAYOUT_INIT/etc for defaulting

kainino0x:
An alternative to PR #424. (This proposal prevents us from removing `Undefined` from the binding type enums, which that PR did.)

This moves the defaulting to `BindingNotUsed` out to `WGPU_BIND_GROUP_LAYOUT_ENTRY_INIT` so that it still generates a `WGPUBindGroupLayoutEntry` with no bindings set, but changes
`WGPU_BUFFER_BINDING_LAYOUT_INIT`
`WGPU_SAMPLER_BINDING_LAYOUT_INIT`
`WGPU_TEXTURE_BINDING_LAYOUT_INIT`
`WGPU_STORAGE_TEXTURE_BINDING_LAYOUT_INIT`
so that _they_ set their respective types to `Undefined` instead of `BindingNotUsed`.

There honestly isn't _that much_ value in this in C, but it is more consistent with JS and should work better for higher-level language bindings. These examples all resolve to
`{ binding, visibility, buffer: { type: "uniform", hasDynamicOffset: false, minBindingSize: 0 } }`:

In JS:
```js
const entry = { binding, visibility, buffer = {} };
```

In C:
```c
WGPUBindGroupLayoutEntry entry = WGPU_BIND_GROUP_LAYOUT_ENTRY_INIT;
entry.binding = binding;
entry.visibility = visibility;
entry.buffer = WGPU_BUFFER_BINDING_LAYOUT_INIT;
// or
WGPUBindGroupLayoutEntry entry = {
  .binding = 0,
  .visibility = visibility,
  .buffer = WGPU_BUFFER_BINDING_LAYOUT_INIT,
  // other fields should get zero-initialized I think, which if so is what we want
};
```

In C++, something like:
```cpp
wgpu::BindGroupLayoutEntry entry{
  .binding = 0,
  .visibility = visibility,
  .buffer = {},
  // note that unlike in C, gcc/clang with -Wextra do warn on missing fields in designated initializers
};
```

### DIFF
--- a/doc/articles/SentinelValues.md
+++ b/doc/articles/SentinelValues.md
@@ -16,7 +16,7 @@ can be called equivalently as `b.getMappedRange()`, `b.getMappedRange(0)`,
 
 To represent `undefined` in C, `webgpu.h` uses `NULL` where possible (anything
 behind a pointer, including objects), `*_UNDEFINED` sentinel numeric values
-(generally `UINT32_MAX`, etc.) and `*_Undefined` enum values (generally `0`).
+(usually `UINT32_MAX` or similar) and `*_Undefined` enum values (usually `0`).
 
 The place that uses the type will define what to do with an undefined value.
 It may be:
--- a/gen/cheader.tmpl
+++ b/gen/cheader.tmpl
@@ -70,6 +70,7 @@
 #define _wgpu_COMMA ,
 #if defined(__cplusplus)
 #  define _wgpu_ENUM_ZERO_INIT(type) type(0)
+#  define _wgpu_STRUCT_ZERO_INIT {}
 #  if __cplusplus >= 201103L
 #    define _wgpu_MAKE_INIT_STRUCT(type, value) (type value)
 #  else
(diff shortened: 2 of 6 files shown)


kainino0x:
An alternative to PR #424. (This proposal prevents us from removing `Undefined` from the binding type enums, which that PR did.)

eliemichel:
Now that the semantics of `Undefined` is to mean "use the default value" rather than "this field must be manually defined" (due to #446), this makes a lot of sense. We should note to ourselves somewhere that this relies on `NotUsed` being 0, as `NotUsed` is not explicitly stated in `WGPU_BIND_GROUP_LAYOUT_ENTRY_INIT`.

kainino0x:
Dec 9 meeting:
* (Discuss only after \#446 is resolved\!)  
* This proposal is competing with [\#424](https://github.com/webgpu-native/webgpu-headers/pull/424)  
* **CF: sounds good**  
* **And close \#424**

lokokung (approved):

PR #451, opened 2024-11-28
Add missing defaulting docs to binding layout types

kainino0x:
Missed these in #446 because their defaults are not `Undefined`, but they still _have_ `Undefined` entries which do trigger defaulting.

Docs are extra explicit because `Undefined` is not the zero value and won't be the initial value in `WGPU_BIND_GROUP_LAYOUT_ENTRY_INIT` (regardless of #450)


eliemichel (approved):
👍 I'm thinking we should maybe use this phrasing (with the introductory "If set to @ref WGPUBufferBindingType_Undefined,") in all other member docs as well, because reading them back they feel ambiguous when they say "Default to Foo\n\nINIT sets this to Undefined" (even if technically the link to sentinel value doc disambiguates).

PR #453, opened 2024-11-28
Fix inconsistent management of ExtSuffix in INIT macros

eliemichel:
Proper handling of ExtSuffix was missing from in #422 (could be listed in the followup issue #427).

Note that this does not account for potential `namespace`, because I am not sure how we should handle it: if a member from a struct `foo` has type to `struct.bar`, can we be sure that there is a single `bar` entry and use its namespace, or should we look for `bar` first in the same namespace as `foo`, then fallback to default namespace?

SIde note: neither ExtSuffix not namespaces are tested, we could have a minimal extension example in `test`.

### DIFF
--- a/gen/cheader.tmpl
+++ b/gen/cheader.tmpl
@@ -306,7 +306,7 @@ typedef struct WGPU{{.Name | PascalCase}}CallbackInfo{{$.ExtSuffix}} {
 /**
  * Initializer for @ref WGPU{{.Name | PascalCase}}CallbackInfo{{$.ExtSuffix}}.
  */
-#define WGPU_{{.Name | ConstantCase}}_CALLBACK_INFO{{$.ExtSuffix}}_INIT _wgpu_MAKE_INIT_STRUCT(WGPU{{.Name | PascalCase}}CallbackInfo{{$.ExtSuffix}}, { \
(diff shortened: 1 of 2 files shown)


kainino0x (approved):
LGTM except of course we'll need to implement #212 (I forgot that actually needed code changes, since it only affected the C header).

I agree it would be really nice to test this. I'll bet _something_ I added in the past few months is handling it wrong. (It would also be a really useful reference for how to add use extension files.)

We can reuse the `gen-check` pattern but for a sample extension header that we check in under the tests directory. Are you interested in adding it?

> Note that this does not account for potential `namespace`, because I am not sure how we should handle it: if a member from a struct `foo` has type to `struct.bar`, can we be sure that there is a single `bar` entry and use its namespace, or should we look for `bar` first in the same namespace as `foo`, then fallback to default namespace?

I'm not sure either. Off the top of my head, either or both:
- The file format needs to change a bit, because it currently can't distinguish, for example, "add new value to existing enum" vs "add new enum with new value".
- We should always assume that if something is named in an extension header that exists in the core header, it's referring to the same thing, and error if there are conflicting definitions, e.g. two structs with the same name

eliemichel:
> We can reuse the gen-check pattern but for a sample extension header that we check in under the tests directory. Are you interested in adding it?

FTR, started in #462!

Issue #211, opened 2023-08-09
[Push|Pop]ErrorScope thread local error stacks


lokokung:
Came up in a discussion with Kai today, but main question is how we intend to deal with error scopes in a multithread environment. The current model could race when an unrelated error on thread 1 is caught in a `popErrorScope` on thread 2.

The simplest solution for now is a thread-local error stack, however, there are still some open questions / discussion points especially for other language bindings...

- Users must understand that they cannot rely on error scopes set on the passing thread of a callback. Instead, users that are interested in error scopes in callbacks need to explicitly call the pop/push in the callback. (This is probably fine?)
- Does this work for cases where tasks can yield and be rescheduled on a different OS level thread? One thing we talked about is potentially having to add an API that can be called in a context switch hook (does this exist?)? Do we have any other alternatives for this?
- Given that a thread-local solution should be fine for JS, is it possible to make the native bindings the same? One thought was that since the JS API isn't standardized for multithread yet, we **could** make sure that the multithread version doesn't allow calls to `[Push|Pop]ErrorScope`, and instead expose something different so that we could be aligned with how the native version would work? Kai also mentioned that while the thread-local solution **should** work for JS, that it may be possible for some Rust runtimes on top of WASM to do something similar to the point above and run into the same issue.

kainino0x:
(Mostly my own thoughts but some from @lokokung @cwfitzgerald @rajveermalviya and old ideas from the JS API standardization)

The current proposal for JS multithreading would have error scope stacks be thread-local. It's unlikely we'd want to make them racy, as this introduces major threading hazards to the JS API, and JS usually doesn't like those (I think the web platform currently has none - they're in JavaScript itself via SharedArrayBuffer - though the proposal does add another hazard to GPUBuffer mapping).

If we went forward with that, then the model for WASM would have to be the same. Then in order to match that in native, we'd need to make the error scope stack thread local.

Some contexts will not like this, e.g. bindings from languages that have green threads. It could still be the default, if we add a way for language bindings to cover it up. For example:
- we could make the thread-to-error-scope-stack mapping mutable, so the bindings can maintain one per green thread (instead of one per OS thread) and set it current whenever the green thread's OS thread changes. May require JS API support.

[...]

PR #455, opened 2024-12-04
Some additional flavor about thread local error scopes

kainino0x:
Issue: #211


cwfitzgerald (approved):
Now that's done tasty flavor.

PR #456, opened 2024-12-04
Add error sinks

kainino0x:
**WIP draft**

I've loosely sketched out what function-local error sinks would look like. However after sketching them out I'm not totally sure what the point was. Is it just ergonomics? I don't think it makes much sense to introduce an ergonomics feature at this point, and besides using error scopes is really not that bad - and wrt threading, it's easy to see that a straight pushErrorScope/someCall/popErrorScope sequence is all on the same thread.

cc @cwfitzgerald 

Issue: #211


cwfitzgerald:
Hum - I had the thought that maybe this should actually be a pnext thing?

Adding an argument feels a little clunky.

lokokung:
Thoughts on moving this past the v1? If it's not too urgent, maybe we can iterate on it as necessary in the future?

kainino0x:
Closing per https://github.com/webgpu-native/webgpu-headers/issues/211#issuecomment-2529665017

PR #457, opened 2024-12-04
Add PostTask hook

### LINKED ISSUE #187: Worker threads for background pipeline compilation
Without worker threads, "async" pipeline compilation can't really be async. It's also been discussed that shader compilation and non-async pipeline compilation could take advantage of worker threads as well.

If implementations want worker threads, should they start them internally, or should there be something in the API where the user launches their own thread and then starts the webgpu worker in it?

Having the user launch threads presents a few issues:
- In remoting implementations, the threads need to be on the service end, not the client end. So that means you need some library-specific insight into the remoting to do it.
- Similarly in WASM, it won't make any sense.

So I think it should be a per-implementation decision (for example Dawn might launch its own workers, with some platform hooks that control how it does that, whereas wgpu might make the user launch them, and Emscripten won't have any such option).

kainino0x:
Very rough first draft

Fixes #187

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -274,6 +274,7 @@ struct WGPURequestAdapterCallbackInfo;
 struct WGPURequestDeviceCallbackInfo;
 struct WGPUUncapturedErrorCallbackInfo;
 
+struct WGPUPostTaskCallbackInfo;
 
 /**
  * \defgroup Enumerations Enumerations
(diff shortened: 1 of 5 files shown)


webgpu.h
```diff
@@ -1508,6 +1517,15 @@ typedef struct WGPUUncapturedErrorCallbackInfo {
     /*.userdata2=*/NULL _wgpu_COMMA \
 })
 
+typedef struct WGPUPostTaskHookInfo {
+    WGPUPostTaskHook hook;
```
eliemichel, line 1521: Have I missed it or this `WGPUPostTaskHook` symbol is never defined?

kainino0x: It's hard to see because it's a function pointer type... and out of order because I wasn't thinking that hard about it:

`typedef (*WGPUPostTaskHook)(...);`

PR #458, opened 2024-12-05
Move mainpage documentation to yml file

eliemichel:
As I was starting to play around with the generation of header files for custom extensions, I felt the need to have this mainpage comment be driven by the yaml spec file rather than hardcoded in the template!

### DIFF
--- a/webgpu.yml
+++ b/webgpu.yml
@@ -6,10 +6,10 @@ name: webgpu
 enum_prefix: '0x0000'
 doc: |
   **Important:** *This documentation is a Work In Progress.*
-  
+
   This is the home of WebGPU C API specification. We define here the standard
   `webgpu.h` header that all implementations should provide.
-  
+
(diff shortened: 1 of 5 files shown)


eliemichel:
Erf we have a `\n` vs `\r\n` issue here, will investigate.

webgpu.yml
```diff
@@ -697,8 +708,9 @@ enums:
       - name: render_pass_max_draw_count
         doc: |
           TODO
-      # TODO(#214): Move all of the surface sources into block 0x0001
       - name: surface_source_metal_layer
+        # TODO(#214): Move all of the surface sources into block 0x0001
+
```
kainino0x, line 713: This doesn't seem like it should have changed, but otherwise LGTM!

kainino0x (approved):

webgpu.yml
eliemichel: Oh, dunno how this came up here

PR #462, opened 2024-12-09
Test extension header generation

eliemichel:
This is a WIP of a test dedicated to the generation of extension header. The sample yaml file is mainly based on wgpu-native's `wgpu.h` file ([from this branch](https://github.com/eliemichel/wgpu-native/blob/eliemichel/2024-11-12-upgrade-webgpu-headers/ffi/wgpu.h)) as I was thinking of handling two birds with one PR.

Hick-ups I could notice:
- [x] One could not customize the header doc intro (fixed by #458)
- [x] Hyphens in header name (i.e. `my-extension.yml`) would lead to hyphens in include guards (fixed in this PR)
- [x] I believe there are many missing underscores in generated names
- [ ] I did not test much the "namespace" mechanism as I am not sure how to
- [ ] I did not check the whole generated file yet
- [x] The generator was adding suffix rather than prefix as settled by #212


kainino0x:
> The sample yaml file is mainly based on wgpu-native's `wgpu.h` file ([from this branch](https://github.com/eliemichel/wgpu-native/blob/eliemichel/2024-11-12-upgrade-webgpu-headers/ffi/wgpu.h)) as I was thinking of handling two birds with one PR.

It's nice to test a real case but it would be very useful to have a test file that has just one or two of every possible type of extension in it, as that would make it easy to review that the generator is producing the right thing.
Probably not all of these are possible now, but whichever of these _are_ possible should be tested (I'm sure you already hit many or all of these but I haven't checked yet, sorry):
- New constant
- New typedef
- New enum with new value
- New value on old enum
- New bitflags type with new flag
- New flag on old bitflags type
- New struct
- New extension struct to new base struct
- New extension struct to old base struct
- New callback type (and a method that uses it)
- New function
- New object with new method
- New method on old object

eliemichel:
> It's nice to test a real case but it would be very useful to have a test file that has just one or two of every possible type of extension in it, as that would make it easy to review

Sure! Sorry for throwing all that here, I'll strip it down to a more strategic test prior to marking as non-draft!

kainino0x:
No problem, just pre-review thoughts (about how I'm going to review it) :)

PR #463, opened 2024-12-09
Make all struct chains mutable

### LINKED ISSUE #395: Proposal: Make all struct chains mutable, instead of having separate in/out/inout chain types
It doesn't seem like we get much value from `const` in these struct chains, and it introduces some complexity - every chained struct needs to match the mutability of its root struct, which means they can't be mixed. Right now we have "in", "out", and "inout" (#260) struct chains; we could imagine a situation where we need to attach a new "inout" non-root struct to an "in" struct but can't, and it becomes kludgy.

Doing this would obsolete #376, and allow us to defer the decision from #261 about whether to mix inputs and outputs in the same struct chain (also potentially reducing the number of version `2`s of APIs we need if we don't do #216 right now). Loosely affects #264 as well, see there.

kainino0x:
Fixes #395

### DIFF
--- a/gen/README.md
+++ b/gen/README.md
@@ -32,15 +32,11 @@ Since the generator does some duplication validation, the order of the files mat
 | `float64`       |
 | `c_void`        |
 
-| Structure types       |                                                         |
-| --------------------- | ------------------------------------------------------- |
-| `base_in`             | in-structs that can be extended                         |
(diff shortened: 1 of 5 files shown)


lokokung (approved):
Nice! The rename of the structure type is also really nice since I feel like its a lot more clear the difference now!

PR #465, opened 2024-12-09
Remove GetCurrentTextureStatus_OutOfMemory

kainino0x:
This doesn't seem super likely to be handleable distinctly from Lost (surface loss) or losing the device, but if it is, the implementation (wgpu-native) can expose it as an extension. It isn't used in JS or Dawn.

> - LK: Could maybe return a message string? (Though requires a FreeMembers, or a static string.)
> - (... we didn't really discuss this much but I think we thought it wasn't worth it)

@lokokung to confirm because I forgot to take notes

Issue: #401

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -863,14 +863,9 @@ typedef enum WGPUSurfaceGetCurrentTextureStatus {
     WGPUSurfaceGetCurrentTextureStatus_Lost = 0x00000005,
     /**
      * `0x00000006`.
-     * The system ran out of memory.
-     */
-    WGPUSurfaceGetCurrentTextureStatus_OutOfMemory = 0x00000006,
-    /**
-     * `0x00000007`.
      * The surface is not configured, or there was an @ref OutStructChainError.
      */
-    WGPUSurfaceGetCurrentTextureStatus_Error = 0x00000007,
+    WGPUSurfaceGetCurrentTextureStatus_Error = 0x00000006,
     WGPUSurfaceGetCurrentTextureStatus_Force32 = 0x7FFFFFFF
 } WGPUSurfaceGetCurrentTextureStatus WGPU_ENUM_ATTRIBUTE;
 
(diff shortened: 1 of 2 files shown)


lokokung (approved):

kainino0x:
I can't find any places where wgpu could return `SurfaceError::OutOfMemory`:
https://github.com/search?q=org%3Agfx-rs%20SurfaceError%3A%3AOutOfMemory&type=code
I could be looking in the wrong repos though.
It doesn't exist in [`wgpu_hal::SurfaceError`](https://wgpu.rs/doc/wgpu_hal/enum.SurfaceError.html) or [`wgpu::SurfaceStatus`](https://wgpu.rs/doc/wgpu/enum.SurfaceStatus.html) either.
@cwfitzgerald You mentioned it could be returned, is that still true?

Tried looking for historical references but it's difficult because there have been so many different repositories over time.

kainino0x:
> * LK: Could maybe return a message string? (Though requires a FreeMembers, or a static string.)

Loko and I discussed this a bit more, I think it would be pretty annoying to have a string that needs FreeMembers called on it _but only in the error case_ which would make it really difficult to notice if you had a leak.
Making it a static string could work but then there's not that much value in the string, especially if the error cases are pretty bounded.

It does seem like the "Error" case ("The surface is not configured, or there was an @ref OutStructChainError") should (1) be open-ended and (2) come with implementation-defined logging.

kainino0x:
> but if it is, the implementation (wgpu-native) can expose it as an extension.

Oh I forgot to note, I realized this is not directly possible. There would need to be a way for the application to signal that it _wants_ to know about OutOfMemory; an implementation can't just return it without being asked.
Could be done with an extension to WGPUSurfaceTexture, maybe an extra field telling you the reason for Lost, or something.

kainino0x:
Dec 12 meeting:

* CF: `wgpu::SurfaceError` can come from any arbitrary device error. Something weird happened that we think shouldn't happened, only reasonable conclusion is that there's no memory left to create a simple object. Have been trying to move away from that anyway because we have cases like "create semaphone failed with out of memory" but actually it was device lost. Think any problems we have can be bucketed into surface loss.  
* **Still OK to remove.**

Issue #466, opened 2024-12-09
Behavior of NaNs and Infinities passed as float/double


kainino0x:
The JS API throws an exception (or rejects a promise with an exception) if NaN or Infinity is passed into any float/double (this rule comes from WebIDL; we don't use `unrestricted` which would lift it).

What should webgpu.h do when NaN or Infinity is passed to a required float/double, or Infinity is passed to an optional float/double (`depthClearValue`)?

The obvious answer is to produce a regular validation error, but Wasm implementations would _technically_ need to detect these and inject them into the error scope. In practice, this is really unnecessary and impractical, so Wasm implementations would probably just crash by hitting the aforementioned JS exception. We could maybe say "may crash, and if not produces a validation error" I guess?

Forked from isnan question on #427.

eliemichel:
Could be deemed as "undefined behavior", which in practice throws in JS and does whatever in native? Or do we want to avoid UB as much as possible?

kainino0x:
We have been mostly avoiding UB as much as possible, however in this case I think it's unreasonable to fully define the behavior. Even trying to specify it as a crash gets into too many details (what if you catch the Wasm trap and handle it somehow, what if you re-enter Wasm from a new entrypoint; in native, what if you have signal handlers, etc.)

kainino0x:
Dec 12 meeting:

* What happens if you pass infinity/nan to a required float, or infinity to an optional float?  
* KN: Only problem is for wasm, otherwise validation error makes most sense.
* CF: Can we just say exactly that it will cause an exception in wasm? Not make it more undefined than necessary.
* **Do that.**

kainino0x:
I should make a list of what this applies to exactly (`depthClearValue`, `WGPUColor`, ..?).

kainino0x:
Here's the list of all floats and doubles:

- nullable floats
  - `depthClearValue`
- non-nullable floats
  - `lodMinClamp`, `lodMaxClamp`, `depthBiasSlopeScale`, `depthBiasClamp`, `setViewport()` args, 
- non-nullable doubles used as numeric "unions" (see also https://github.com/gpuweb/gpuweb/issues/5020):
  - `WGPUColor.*`, `WGPUConstantEntry.value`

> * Can we just say exactly that it will cause an exception in wasm?

To be more precise we should say it _may_ cause an exception in wasm. On the off chance that either the JS spec changes or some wasm implementation wants to avoid crashing on non-finite inputs.

[...]

PR #467, opened 2024-12-09
depthClearValue is undefined iff NaN

kainino0x:
API documentation.

Issue: #427 
See also #466 for the behavior of infinities


eliemichel:
Maybe we should recall here that `WGPU_DEPTH_CLEAR_VALUE_UNDEFINED` is NaN and so that one SHALL use `isnan()` rather than `depthClearValue == WGPU_DEPTH_CLEAR_VALUE_UNDEFINED`?

eliemichel (approved):

Issue #264, opened 2024-01-11
FreeMembers with struct chains


kainino0x:
Jan 11 meeting:

*   Should FreeMembers iterate down the struct chain for you, or should we have FreeMembers individually on each item in the chain?
*   discussion of how this works in a C++ API for example
    *   If C++ explicitly represents the chain, then is it OK if dropping the root of the chain frees the members of the rest of the chain?
    *   If C++ flattens all chained types into one struct, then it doesn’t matter except that the C++ bindings may have to make multiple FreeMembers calls in a row
*   ~**Tentatively it should iterate, but we don’t have any cases of this yet so TBD**~

kainino0x:
A [question on Matrix chat](https://matrix.to/#/!MFogdGJfnZLrDmgkBN:matrix.org/$5JImNgY-cVTpvfWlBaadE2hgAvgC1UpsxMUMF6Jo_jI?via=matrix.org&via=mozilla.org&via=kit.edu) about how to deal with this in C++ bindings (RAII) raised some issues:

- FreeMembers frees a bunch of arrays that are passed by pointer. Is it guaranteed to handle some/all of them being nullptr? I say "yes" because (1) length=0,ptr=null represents an empty array which could be a valid output of some function and (2) it shouldn't just work on those particular arrays being null, it should also work on "empty" (zeroed) structs.
- If no, wrappers _must_ avoid calling FreeMembers on zeroed structs. If yes, _should_ wrappers still try to avoid it? There is a tradeoff between:
  - complexity/time in the bindings, checking all of the arrays for null (recursively through the struct chain? [1])
    - The performance cost should be optimizable with inlining, but if the binding function is complex enough then it is less likely to get inlined. And bindings code complexity is annoying in its own right.
  - unnecessary calls into the C API
    - If the C API is dynamically linked or called by function pointer, then that is an unoptimizable cost.

  [1]: Recursing through the struct chain isn't required for anything right now but will probably eventually happen. OTOH if we reverse the tentative decision above and have FreeMembers for each struct in the chain (instead of one recursive one for the head struct), then each struct can be individually RAII'd, avoiding recursion. Then perhaps we can rely on compilers to inline+dead-code-eliminate the FreeMembers on default-initialized/moved-from structs.

kainino0x:
cc @Kangz

Kangz:
[...]

PR #469, opened 2024-12-12
Add struct chaining docs

kainino0x:
Also add compile tests for the things I wrote in the documentation sample code.

Issue: none, #264, should have been listed in #25, related to #463

### DIFF
--- a/doc/articles/StructChaining.md
+++ b/doc/articles/StructChaining.md
@@ -62,8 +62,10 @@ The pointer links in a struct chain are all mutable pointers. Whether the struct
 A struct-chaining error occurs if a struct chain is incorrectly constructed (in a detectable way).
 They occur if and only if:
 
-- The `sType` of a struct in the chain is not valid for the type of the chain root.
+- The `sType` of a struct in the chain is not valid _in the context of the chain root's static type_.
+    - If this happens, the implementation must not downcast the pointer to access the rest of the struct (even if it would know how to downcast it in other contexts).
 - Multiple instances of the same `sType` value are seen in the same struct chain. (Note this also detects and disallows cycles.)
+    - Implementation-specific extensions also _should_ avoid designs that use unbounded recursion (such as linked lists) in favor of iterative designs (arrays or arrays-of-pointers). This is to avoid stack overflows in struct handling and serialization/deserialization.
 
 Struct chains which are used in device-timeline validation/operations (e.g. @ref WGPUBufferDescriptor in @ref wgpuDeviceCreateBuffer) have their chain errors surfaced asynchronously, like any other validation error.
 
--- a/doc/articles/StructChaining.md
+++ b/doc/articles/StructChaining.md
@@ -1,0 +1,81 @@
+# Struct-Chaining {#StructChaining}
+
+Struct-chaining is a C API pattern using linked lists and dynamic typing to extend existing structs with new members, while maintaining API and ABI compatibility. For example:
+
+An extensible struct is statically typed. It is the root of a linked list, containing a pointer to the next struct in the chain:
+
+```c
+typedef struct WGPUMyStructBase {
+    WGPUChainedStruct * nextInChain;
+    uint32_t x;
+} WGPUMyBaseStruct;
+```
+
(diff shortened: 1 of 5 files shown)


doc/articles/StructChaining.md
```diff
@@ -0,0 +1,81 @@
...
+
+The pointer links in a struct chain are all mutable pointers. Whether the structs in the chain are actually mutated depends on the function they are passed to (whether the struct chain is passed as an input or an output). Some structs (e.g. @ref WGPULimits) may be either an input or an output depending on the function being called.
+
+## Struct-Chaining Error {#StructChainingError}
+
+A struct-chaining error occurs if a struct chain is incorrectly constructed (in a detectable way).
+They occur if and only if:
+
+- The `sType` of a struct in the chain is not valid for the type of the chain root.
+- Multiple instances of the same `sType` value are seen in the same struct chain. (Note this also detects and disallows cycles.)
```
lokokung, line 66: nit. To clarify this a bit more (at least w.r.t the fuzzer case), it's not just that the same `sType` can't be in the same chain, but that the same `sType` should never appear in a chain or member chain on a root struct. Not sure how to word it clearly here...

doc/articles/StructChaining.md
```diff
@@ -0,0 +1,81 @@
...
+
+Struct chains which are used in device-timeline validation/operations (e.g. @ref WGPUBufferDescriptor in @ref wgpuDeviceCreateBuffer) have their chain errors surfaced asynchronously, like any other validation error.
+
+Struct chains which are used in content-timeline operations (e.g. @ref OutStructChainError) have their chain errors surfaced synchronously, like other content-timeline validation errors.
+
+### Out-Struct-Chain Error {#OutStructChainError}
+
+Operations which take out-struct-chains (e.g. @ref WGPULimits, in @ref wgpuAdapterGetLimits and @ref wgpuDeviceGetLimits, but not in @ref WGPUDeviceDescriptor) handle struct-chaining errors as follows:
+
+- The output struct and struct chain is not modified.
```
lokokung, line 76: Is this what we decided? I thought that we decided that this might be hard for implementation in cases of multiple output struct chains. i.e. It might be harder for implementations to handle resetting everything in the chain if a later chain errors?

My initial thought was that if error, users should still call `FreeMember` on all output structs.  For any struct that we successfully wrote to, that will free the resources, and on the ones we error-ed on, we will not write anything, and hence if users used INIT macros to initialize the out structs, `FreeMember` should work and not do anything.

lokokung (changes requested):

doc/articles/StructChaining.md
kainino0x: (discussed offline)

The first requirement (sType is valid for the type of the chain root) should be the one that validates that. The only gap is if some implementation adds an extension that can _actually_ recurse, which we probably should not allow them to do, so I'll note that.

We should also guarantee that if the implementation doesn't allow the sType in that context, it will never try to downcast the struct. It's the same as if the sType were totally bogus.

In Dawn's wire we came up with a way to guarantee that invalid chains can be serialized and deserialized without hitting serialization errors: if the client doesn't allow an sType in context, convert it to something like `WGPUSType_DawnInvalidChainedStruct` which indicates `struct WGPUDawnInvalidChainedStruct { WGPUChainedStruct chain; WGPUSType sTypeForErrorMessage; }`. As long as the client and server are in sync the server can always handle this without a deserialization error, and forward it into dawn-native which will produce a validation error. If the client is out of sync or is malicious (or a fuzzer) then if the server sees an sType that's not valid in context it immediately gives up and fails deserialization.

kainino0x: > The only gap is if some implementation adds an extension that can _actually_ recurse, which we probably should not allow them to do, so I'll note that.

Actually I started trying to write this down and realized there _could_ be legitimate use cases that pass variable-sized data as a linked list (rather than an array or array of pointers), regardless of whether they're using struct chaining. So I'm going to make this a "should", not a "must".

doc/articles/StructChaining.md
kainino0x: I added this in #377 (it just moved in this PR), but that was when we thought `FreeMembers` would be called on the whole chain, not on individual structs in the chain. Then it wouldn't make any sense to tell people to call `FreeMembers` because `FreeMembers` wouldn't be able to traverse the chain either. (Which now that I think about it is another reason `FreeMembers` perhaps shouldn't traverse the chain; it would need to return a `wgpu::Status`.)

That said I think the only thing implementations need to do to implement this correctly is unpack the struct chain before writing anything to it. This seems kinda necessary anyway because if the struct hasn't been unpacked we can't even really write anything out except for the base struct.

lokokung: Ack!

lokokung (approved):

Issue #470, opened 2024-12-14
[try for V1] add new limits added for Compatibility Mode


kainino0x:
I don't expect these will be removed when we drop compatibility mode, so they should be in core (not a Compat extension struct). Currently:
- maxStorageTexturesInFragmentStage
- maxStorageBuffersInFragmentStage
- maxStorageTexturesInVertexStage
- maxStorageBuffersInVertexStage

kainino0x:
I went to implement this and then I made this note:

> Note these limits will be part of the core JS spec, because they can't be safely removed from JS (even when compat doesn't exist and they won't do anything).

The reason for this is that if we remove them from `GPUSupportedLimits` then applications accessing `GPUSupportedLimits.maxStorageBuffersInVertexStage` will get `undefined`.

In C, it would be somewhat safer to remove them, but not entirely. The equivalent of `adapter.limits` is `wgpuAdapterGetLimits` which takes `WGPULimits` as an out-struct. If we chained `WGPULimits` -> `WGPUCompatibilityModeLimits` which has an `SType` in the Compat block, then non-compat-supporting webgpu.h implementations would skip over the extension struct and leave it uninitialized, instead of initializing it to the correct values, which could impact application behavior.

HOWEVER! An application that doesn't target Compat also won't care about these limits. I was thinking it was good to get these into the main struct just to avoid more extensions. But this is kinda similar to `MaxDrawCount`, where it's core, but just something most people won't care about, so I think these should be in a block-0x0000 extension instead. Which makes them non-blocking for 1.0.

PR #472, opened 2024-12-19
Add storage limits for Compat

kainino0x:
I think we're reasonably confident that these will be the correct spellings. Hopefully this is all of the limits we add for compat but even if not, better to have them in now than later.

Note these limits will be part of the core JS spec, because they can't be safely removed from JS (even when compat doesn't exist and they won't do anything).

Issue: #470


kainino0x:
I realized this probably is not necessary. Will post on the issue.

Issue #272, opened 2024-02-15
Passing extensible structures to application callbacks, backward-compatibly


kainino0x:
Raised in #266: we want a backward-compatible way of passing extended data to application-defined callbacks. This comes up in two places:
- GetCompilationInfo: the callback receives WGPUCompilationInfo which is extensible, and also contains WGPUCompilationMessage objects which are also extensible
- RequestAdapterInfo: the callback receives WGPUAdapterInfo which is not yet extensible, but needs to be

If a webgpu.h implementation suddenly starts passing new extension structs to the application, it runs the risk of the application mishandling them.

For RequestAdapterInfo we decided that the WGPUAdapterInfo should be passed as an argument to RequestAdapterInfo so the application can set up the extension chain. Then we need a wgpuAdapterInfoFreeMembers (which replaces wgpuAdapterPropertiesFreeMembers proposed in #143 but which is becoming obsolete with #266).

kainino0x:
Feb 29 meeting:

*   (Offshoot of [#266](https://github.com/webgpu-native/webgpu-headers/issues/266) but more generally - includes GetCompilationInfo too)
*   KN: (explanation of problem) CompilationInfo is extensible but also holds a
*   AE: RequestAdapterInfo/GetAdapterInfo are now more like MapAsync where it
*   …
*   RM: We need a pattern to request additional info anyway
*   KN: Think this is the only one that currently needs it
*   LK: A parallel array with the same indices in the CompilationInfo chain?
*   (take the chain out of the CompilationMessage)
*   LK: What if not every message has the same chain?
    *   KN: Can make it nullable somehow
*   How would this look? RequestCompilationInfo + GetCompilationInfo?
    *   Compilation info already lives forever (with the shader module) but in the wire we would be holding a _second_ copy alive (on the client side)
*   (discussion)
*   **OK with removing extensibility from CompilationMessage**
*   (discussion of doing the previous proposal for AdapterInfo)
*   KN: Not the worst if we say applications need to handle chain structs they don’t know about. (If we do that then CompilationMessage can be kept as extensible.)
*   LK: Would need a way to request which AdapterInfo extensions you want anyway
    *   Bitmask? Enum array? Dummy “prototype” struct chain?
    *   AE: Prefer array
        *   RM: agree, array of sTypes rather than creating new bitmasks
*   — out of time —
[...]

PR #476, opened 2024-12-21
Docs: implementation-allocated struct chains, and FreeMembers traversal

kainino0x:
Issue: #272, #264

### DIFF
--- a/webgpu.yml
+++ b/webgpu.yml
@@ -1815,10 +1815,7 @@ structs:
   - name: compilation_info
     doc: |
       TODO
-
-      This is an @ref ImplementationAllocatedStructChain root.
-      Arbitrary chains must be handled gracefully by the application!
-    type: extensible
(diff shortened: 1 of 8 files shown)


webgpu.yml
```diff
@@ -1815,6 +1815,9 @@ structs:
   - name: compilation_info
     doc: |
       TODO
+
+      This is an @ref ImplementationAllocatedStructChain root.
+      Arbitrary chains must be handled gracefully by the application!
```
lokokung, line 1820: nit. Would it be possible to add this note to all `type: extensible` instead?

lokokung (approved):

webgpu.yml
kainino0x: It doesn't apply to all `type: extensible` but I can add a new `type: extensible_callback_arg` or something

Issue #57, opened 2020-06-24
Signaling allocation failure for buffers with `mapAtCreation==true`


austinEng:
Following [gpuweb#872](https://github.com/gpuweb/gpuweb/issues/872), we need a way for buffer creation with `mapAtCreation==true` to fail.

I can think of a couple options, none of which I love, so I would be open to other ideas.

## Option 1
`wgpuDeviceCreateBuffer` returns `nullptr`. This would match the JS API best (throws an exception, and the `GPUBuffer` would be `undefined`), but has the unfortunate consequence that it is the only part of the C API that returns something null.

## Option 2
`wgpuDeviceCreateBuffer` works, but `wgpuBufferGetMappedRange` returns `nullptr`. This would be a little more natural as `wgpuBufferGetMappedRange` is what gives you the pointer, and developers would be more used to checking if a `void*` is non-null. The drawbacks are as follows:
 - It would not match the JS API. You would still get back a `WGPUBuffer` when using the C API.
 - Implementing it would involve some magic in the WASM to JS layer (ex. Emscripten) to `try {} catch {}` around the call to `wgpuDeviceCreateBuffer` and then give back some proxy object which does nothing on the other methods and return `nullptr` from `getMappedRange`

Thinking about this after I wrote it, Option 2 may have way too much magic involved with other places a `WGPUBuffer` is used like making bind groups and stuff. *Maybe* it is workable if the implementation reserves a handle for the "not actually allocated" `WGPUBuffer` object but it doesn't seem like a good idea. Only Option 1 looks feasible.

kvark:
Agreed, Option 1 is fine. Thank you for raising this!

austinEng:
@Kangz any thoughts?

Kangz:
Both have large drawbacks. Option 1 is unfortunate because it means that applications have to check all buffer creation if they want to do perfect error handling. It is relatively simple to implement in dawn_wire by forcing the creation of an error buffer if CreateBuffer return nullptr. Actually this gives a good implementation of Option 2 in WASM -> JS: if JS return an exception, then the shim can create an error buffer, and use it instead.

I have a slight preference for 2, so that people don't have to check all buffer creations, and only GetMappedRange. But no super strong opinion.

kvark:
The contract is that only buffers with `mapAtCreation == true` could return null, so the users are safe to omit that check and not expect a NPE for other buffers.

austinEng:
[...]

PR #477, opened 2024-12-21
Docs: wgpuDeviceCreateBuffer returns null if mapAtCreation fails

kainino0x:
Issue: #57


lokokung (approved):

kainino0x:
We should also specify either that there's an `out-of-memory` error scope error, or implementation-defined logging. Adding `needs docs` to do this.

kainino0x:
I don't think it should be an `out-of-memory` error because the JS spec doesn't produce one in this case. It just throws a `RangeError` from `createBuffer()`. https://gpuweb.github.io/gpuweb/#dom-gpudevice-createbuffer

Issue #360, opened 2024-10-01
Clarify the behavior of `wgpuInstanceWaitAny` called from a callback


eliemichel:
Question about `WGPUFuture`s: Is it legal to call `wgpuInstanceWaitAny` from within a callback, and if so, what completion status is reported for the `WGPUFuture` of the callback we are currently in?

(EDIT: parent issue #199)

---

FTR, replies got on matrix:

> *From @Kangz*
> IMHO it should work, but it might be difficult to implement

> *From @kainino0x*
> I think it should work, it's implementable in theory but YMMV with actual implementations
> The (now outdated, but mostly correct) design doc says that the completed field on your first WaitAny should have already been set before the callback is called. However at least at that time, I left the answer to your question undefined: https://docs.google.com/document/d/1qJRTJRY318ZEqhK6ryw4P-91FumYQfOnNP6LpANYy6A/edit#bookmark=id.8zqwnqgij106
> Until recently Dawn would actually assert-crash in debug mode if you did that, but we took it out

kainino0x:
Thinking on this further, I think it's necessary to make this safe. Sure, we could say it's unsafe to `WaitAny(f1)` during `f1`'s own callback inside another `WaitAny(f1)`.

But what if you `WaitAny(f1)` while `f1`'s callback is being called by `ProcessEvents`? If that's unsafe too, it becomes hazardous to use `WaitAny` on any `AllowProcessEvents`-mode future in a multithreaded program.

What if you `WaitAny(f1)` while `f1`'s callback is being called _spontaneously_? You effectively can't `WaitAny` on any `Spontaneous`-mode future at all, even in a single-threaded application. (The application can't just add its own locking in the callback - the critical section is larger than the whole callback.)

If those things have to be safe then nested WaitAny may as well have to be safe too.

kainino0x:
> If those things have to be safe

OTOH they don't necessarily have to be safe. There just would be no way to do them safely.

kainino0x:
I formalized the less safe version in #416. This might be too pessimistic though.

cc @lokokung

lokokung:
FWIW, I think `WaitAny` inside a callback _should_ work in our current implementation... (I haven't explicitly tried it, but I do remember ensuring that we release any `Instance` related locks before calling callbacks which should in turn make it fine.)

[...]

PR #481, opened 2024-12-21
Docs: callback reentrancy

kainino0x:
Issue: #360

### DIFF
--- a/doc/articles/Asynchronous Operations.md
+++ b/doc/articles/Asynchronous Operations.md
@@ -28,6 +28,7 @@ The `callback` function pointer is called when the application _observes complet
 > @copydoc ::WGPUCallbackMode_AllowSpontaneous
 
 ## wgpuInstanceWaitAny {#Wait-Any}
+
 `WGPUWaitStatus wgpuInstanceWaitAny(WGPUInstance, size_t futureCount, WGPUFutureWaitInfo * futures, uint64_t timeoutNS)`
 
 Waits on any WGPUFuture in the list of `futures` to complete for `timeoutNS` nanoseconds. Returns when at least one `WGPUFuture` is completed or `timeoutNS` elapses, whichever is first. If `timeoutNS` is zero, all `futures` are polled once, without blocking.
@@ -35,6 +36,7 @@ Waits on any WGPUFuture in the list of `futures` to complete for `timeoutNS` nan
 Returns @ref WGPUWaitStatus_Success if at least one `WGPUFuture` completes. WGPUFutureWaitInfo::completed is set to true for all completed futures. See @ref WGPUWaitStatus for other status codes.
 
 Within this call, for any `WGPUFuture`s that completed, their respective callbacks will fire.
+Note that the timeout applies only to the "wait" phase, and does not affect the callback firing phase.
 
 ### Timed Wait {#Timed-Wait}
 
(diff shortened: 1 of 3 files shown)


doc/articles/Asynchronous Operations.md
```diff
@@ -69,3 +71,16 @@ Device events are slightly different in that their callback info (`WGPUDeviceLos
...
+    - This does not introduce any unsafety. Implementations are required to exit critical sections before invoking callbacks.
+    - This includes making nested calls to @ref wgpuInstanceWaitAny() and @ref wgpuInstanceProcessEvents(). (However, beware of overflowing the call stack!)
+- Calls to the API are nested inside @ref WGPUCallbackMode_AllowSpontaneous callbacks, which may happen at any time (during a `webgpu.h` function call or on an arbitrary thread).
+    - It is therefore always unsafe to make calls to the API in a @ref WGPUCallbackMode_AllowSpontaneous callback. Additionally, applications should take extra care even when accessing their *own* state (and locks) inside the callback.
+    - This includes @ref WGPUDeviceDescriptor::uncapturedErrorCallbackInfo, which is always allowed to fire spontaneously.
+- Two threads are doing things in parallel and so calls to the API are made while callbacks (or API functions in general) are running on another thread.
```
lokokung, line 85: nit in parallel and so calls -> in parallel so calls

doc/articles/Asynchronous Operations.md
```diff
@@ -69,3 +71,16 @@ Device events are slightly different in that their callback info (`WGPUDeviceLos
...
+
+## Callback Reentrancy {#CallbackReentrancy}
+
+There are three cases of async/callback re-entrancy:
+
+- Calls to the API are nested inside non-spontaneous callbacks (those with @ref WGPUCallbackMode_WaitAnyOnly or @ref WGPUCallbackMode_AllowProcessEvents). Such callbacks always happen in @ref wgpuInstanceWaitAny() or @ref wgpuInstanceProcessEvents().
+    - This does not introduce any unsafety. Implementations are required to exit critical sections before invoking callbacks.
+    - This includes making nested calls to @ref wgpuInstanceWaitAny() and @ref wgpuInstanceProcessEvents(). (However, beware of overflowing the call stack!)
+- Calls to the API are nested inside @ref WGPUCallbackMode_AllowSpontaneous callbacks, which may happen at any time (during a `webgpu.h` function call or on an arbitrary thread).
+    - It is therefore always unsafe to make calls to the API in a @ref WGPUCallbackMode_AllowSpontaneous callback. Additionally, applications should take extra care even when accessing their *own* state (and locks) inside the callback.
```
lokokung, line 83: Do you think "always unsafe" would be too pessimistic? I'm wondering if we could say "potentially unsafe"? (Mainly I'm thinking about `getMappedRange` calls in `MapAsync` callback which would be nice to use in spontaneous mode)

kainino0x: I prefer "always unsafe" to avoid tempting too much questioning of _when_ it's safe. That said, am I remembering correctly that it's never really _unsafe_, it's just that it can cause a self-deadlock if it tries to take a lock that's already held on the same thread?

Also, for GetMappedRange maybe we could specifically document that it's safe to call in any callback (or at least in `BufferMap` and `OnSubmittedWorkDone`). That would require the implementation to guarantee either of the following:
- said callbacks won't called spontaneously while holding any lock that GetMappedRange needs
- GetMappedRange doesn't take any locks (it probably doesn't need them, it just needs an atomic)

```suggestion
    - Therefore, calls to the API should never be made in @ref WGPUCallbackMode_AllowSpontaneous callbacks, except where otherwise noted, due to the possibility of self-deadlock.
      Additionally, applications should take extra care even when accessing their *own* state (and locks) inside the callback.
```

lokokung: The suggestion LGTM!

lokokung (approved):

PR #482, opened 2024-12-21
Docs: Extensions

kainino0x:
Issue: #214

### DIFF
--- a/doc/articles/Extensions.md
+++ b/doc/articles/Extensions.md
@@ -31,8 +31,8 @@ When an application is running against an unknown `webgpu.h` implementation, ext
 
 - New functions/methods may be runtime-detected by loading them dynamically, and checking whether loading succeeds. (@ref wgpuGetProcAddress() returns `NULL` for unknown function names.)
     - New objects may be detected by the presence of the methods that create them.
-    - New (root) structs and callback types are always supported if the methods that accept them exist.
-- New enums, bit flags, [chained structs](@ref StructChaining) are available iff enabled. Like other features, these must be both detected and enabled:
+    - New (root) structs, enum/bitflag types, and callback types are always supported if the methods that accept them exist.
+- New enum/bitflag values and [chained structs](@ref StructChaining) are available iff the corresponding "feature" was already explicitly enabled for the context where they're used:
     - Device features are detected via @ref wgpuAdapterGetFeatures() and enabled via @ref WGPUDeviceDescriptor::requiredFeatures.
     - Instance features are detected via @ref wgpuGetInstanceCapabilities() and enabled via @ref WGPUInstanceDescriptor::capabilities.
 
--- a/doc/articles/Extensions.md
+++ b/doc/articles/Extensions.md
@@ -54,8 +54,8 @@ Implementation-specific and multi-implementation extensions **must** not use blo
 
 |                    | Prefix       | Enum Block    | Description
 |--------------------|--------------|---------------|------------
-| Core               | (none)       | `0x0000_????` | Extensions that are required to implement (e.g. added after 1.0).
-| Multi-Vendor       | (none)       | `0x0001_????` | Extensions that are optional to implement (e.g. platform-specific extensions). *TBD if enum values in this block may become required, or if they would be aliased into block `0x0000_????` first.*
+| Core               | (none)       | `0x0000_????` | Extensions that are required to implement, e.g. added after 1.0.
+| Multi-Vendor       | (none)       | `0x0001_????` | Extensions that are optional to implement (unless otherwise specified as required), e.g. platform-specific extensions.
 | Compatibility Mode | *TBD*        | `0x0002_????` | **Special:** implementations that don't support Compatibility Mode must ignore any chained structs with @ref WGPUSType values in this block, instead of erroring. This block must only be used for Compat additions that can be ignored without affecting the semantics of a non-erroring program.
 | wgpu-native        | `Wgpu`       | `0x0003_????` |
 | Emscripten         | `Emscripten` | `0x0004_????` |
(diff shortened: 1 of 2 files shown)


doc/articles/Extensions.md
```diff
@@ -0,0 +1,71 @@
...
+## Registry of prefixes and enum blocks
+
+Implementation extensions **should**, and multi-implementation extensions **must**, use the naming conventions listed above with the prefixes listed here.
+
+Implementation-specific and multi-implementation extensions **must** not use block `0x0000_????` when adding values to existing enums.
+
+|                    | Prefix       | Enum Block    | Description
+|--------------------|--------------|---------------|------------
+| Core               | (none)       | `0x0000_????` | Extensions that are required to implement (e.g. added after 1.0).
+| Multi-Vendor       | (none)       | `0x0001_????` | Extensions that are optional to implement (e.g. platform-specific extensions). *TBD if enum values in this block may become required, or if they would be aliased into block `0x0000_????` first.*
```
kainino0x, line 58: There is a question buried here which I noticed in #417 and then completely forgot about.

> Should the new enums really use block 0x0001? Seems like if we ever put this stuff in "core" they could use the same enums, would we then duplicate them into block 0x0000 or just use the block 0x0001 values? Or should we just put them in block 0x0000 to start?

Ugh. Can't keep track of all this minutiae.

kainino0x: @cwfitzgerald any thoughts? (no rush, I'm ~out until Jan)

lokokung: Maybe stupid question... but would it be possible to just renumber an enum if it gets promoted? Would that be a breaking change? (I guess it COULD be if someone is using the actual value of the enum or if we have different header versions in a wire situation?) I'm just wondering how much breakage it could cause and whether we could just promote them and say here that we may do that so don't rely on the actual value of the enums.

kainino0x: We could say that the value is not API stable, but for ABI stability, we can't just renumber. We can do it if we reserve and implement both the old and new numbers (even though only one is in any given version of the header) but that's kinda obtuse.

lokokung: Ok makes sense! Though honestly, the more than one number thing doesn't sound TOO bad IMO.

kainino0x: I guess the problem here is lack of prefixing. If it were `WGPUSType_DawnNewThing = 0x0005_0004` we could easily duplicate it as `WGPUSType_NewThing = 0x0000_00A3`. But since these are unprefixed we can't duplicate, we can only renumber. But the implementation still needs to handle old numbers that now have no name. We could _give_ them a name with something like:
`WGPUSType_NewThing = 0x0001_0004`
->
`WGPUSType_NewThing = 0x0000_00A3`
`WGPUSType_OldNameForNewThing = 0x0001_0004`

... but I think that's worse than the other two options:
- Give optional extensions a prefix (like `Ext`) and remove it when making core (which has a questionable benefit of being able to change the behavior when making it core, though I would avoid doing that)
- Merge blocks 0 and 1 and just use documentation to indicate which things are required and which are not

doc/articles/Extensions.md
```diff
@@ -0,0 +1,71 @@
...
+WGPUPrefixNewCallback
+
+WGPU_PREFIX_NEW_CONSTANT
+```
+
+("Prefix" is the name of the implementation that owns the extension, if any; see below.)
+
+When an application is running against an unknown `webgpu.h` implementation, extension support may be detected at runtime as follows:
+
+- New functions/methods may be detected via @ref wgpuGetProcAddress().
```
kainino0x, line 32: This is only true with dynamic linking. Filed #488

doc/articles/Extensions.md
```diff
@@ -0,0 +1,71 @@
...
+    - New objects may be detected by the presence of the methods that create them.
+    - New (root) structs and callback types are only relevant if the methods that accept them exist.
+- New enums, bit flags, [chained structs](@ref StructChaining) are available iff enabled. Like other features, these must be both detected and enabled:
```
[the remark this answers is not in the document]
kainino0x: Revised, PTAL

doc/articles/Extensions.md
kainino0x: ^ ... or what would happen by default if we didn't change anything:
- Keep blocks 0 and 1 but with the possibility that some things in block 1 end up being required. This is honestly fine and I think we can just go with it.

---

Actually thinking about the scenario in which this actually happens, it would have to be a multi-vendor extension (meaning we standardized it) which we _initially_ thought would be optional but later _became_ required. I think the most likely case of this is something that's multi-vendor, but we're not sure it's going to show up in the JS API, and then it does. (Like, if the JS API got SPIR-V support somehow, or we decided it was critical to polyfill, then WGPUShaderSourceSPIRV could become required.)

lokokung (approved):

doc/articles/Extensions.md
```diff
@@ -0,0 +1,71 @@
...
+## Registry of prefixes and enum blocks
+
+Implementation extensions **should**, and multi-implementation extensions **must**, use the naming conventions listed above with the prefixes listed here.
+
+Implementation-specific and multi-implementation extensions **must** not use block `0x0000_????` when adding values to existing enums.
+
+|                    | Prefix       | Enum Block    | Description
+|--------------------|--------------|---------------|------------
+| Core               | (none)       | `0x0000_????` | Extensions that are required to implement, e.g. added after 1.0.
+| Multi-Vendor       | (none)       | `0x0001_????` | Extensions that are optional to implement (unless otherwise specified as required), e.g. platform-specific extensions.
```
kainino0x, line 58: Oops this should say Multi-Implementation, will fix

doc/articles/Extensions.md
kainino0x: Filed #490

Issue #490, opened 2025-01-08
Extension enum numbering: block 0 core vs block 1 multi-vendor


kainino0x:
Yet another question on the minutiae of these blocks :)
Over in https://github.com/webgpu-native/webgpu-headers/pull/482#discussion_r1894576999 we've been discussing the relation between enum value block `0x0000_????` "Core"/"required to implement" and block `0x0001_????` "Multi-Implementation Extension"/"optional to implement".

The particular issue arises because (1) both of them are unprefixed and (2) "multi-implementation" extensions could be promoted to "required". We thought of 4 strategies for handling this:

1. Allow things in block 1 to become required in the future.
  - Potentially we could do this in the future with implementation-specific extensions too (`WGPUSType_DawnNewThing = 0x0005_000a`), at which point it would make sense to add an unprefixed name, but we could still use a prefixed enum value (`WGPUSType_NewThing = 0x0005_000a`.
2. Just merge blocks 0 and 1 into block 0, and have some extensions be required and some be optional.
3. Give block 1 a prefix so we can have duplicate things (`WGPUSType_ExtNewThing = 0x0001_000a`) into block 0 (`WGPUSType_NewThing = `0x0000_0015`).
  - These can mean the same thing, or be two incompatible versions of the extension, but that's kind of hazardous; various problems have arisen from doing so in GL, though sometimes it's not really avoidable when taking an implementation-specific extension and making it multi-implementation or core.
4. When making something optional (`WGPUSType_NewThing = 0x0001_000a`) required, have the header renumber it into core (`WGPUSType_NewThing = 0x0000_0015`) while having implementations (that accepted both values) keep accepting the old value (`0x0001_000a`, which could be, but does not necessarily have to be, named in the header)

TBH it doesn't really matter because the point of numbering blocks is really just so that implementations have namespaces to work with, so with that in mind I think option 2 makes the most sense, but option 1 is also totally fine.

cc @lokokung @cwfitzgerald

kainino0x:
I forgot again that I should look at Vulkan to see what it does. It was a little tricky to find the documentation, but it's here:
https://registry.khronos.org/vulkan/specs/latest/styleguide.html#extensions-assigning-token-values

In short, IIUC:
- Everything in block 0 is from vulkan 1.0 and they're not adding new things? Kinda unclear about this
[...]

PR #493, opened 2025-01-11
Require use of the bitflag registry; improve bitflag docs

kainino0x:
Issue: #490
(dawn bug https://crbug.com/389099883)

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -1155,44 +1155,149 @@ typedef enum WGPUWaitStatus {
  *
  * @{
  */
+/**
+ * For reserved non-standard bitflag values, see @ref BitflagRegistry.
+ */
 typedef WGPUFlags WGPUBufferUsage;
+/**
+ * `0x0000000000000000`.
+ */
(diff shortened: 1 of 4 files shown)


webgpu.h
```diff
@@ -1155,44 +1155,149 @@ typedef enum WGPUWaitStatus {
  *
  * @{
  */
+/**
+ * For reserved non-standard bitflag values, see @ref BitflagRegistry.
+ */
 typedef WGPUFlags WGPUBufferUsage;
+/**
```
lokokung, line 1162: nit. These all seem really redundant, given that its documented in the code right below. Is it for doc generation that this is necessary?

lokokung (approved):

webgpu.h
kainino0x: You're right it's pretty redundant. Turns out the docs do show the definition:
image
unlike for enums:
image

Of course this html view is not the only possible way to look at things. But maybe the only useful docs are special cases like 0 and Red|Green|Blue|Alpha. Will follow up.

Issue #66, opened 2020-07-31
memset/memcpy for mapped buffers in WASM


kainino0x:
@juj on the implementation of these two functions, which involve a `malloc`, `Uint8Array.set`, and `free`.

> Oh, these two functions `wgpuBufferGetConstMappedRange` and `wgpuBufferGetMappedRange` are horrible 🤮 😞
> 
> I suppose mapping a GPU buffer directly to wasm heap address space will be a hard no-go? (has that been discussed with Wasm folks? CC @lukewagner @dschuff )
> 
> Assuming it is not possible, I wonder if we should consider changing up the API completely. Given that there is only asynchronous mapping support, we should make this a callback-based API where a cb function gets called when the map happens, and the cb function will receive a handle ID, and then have dedicated functions
> 
> ```js
> void wgpuMemset8ToMappedBuffer(uint destinationMappedBufferHandleId, uint destinationOffset, uint destinationSize, uchar fillValue);
> void wgpuMemset16ToMappedBuffer(uint destinationMappedBufferHandleId, uint destinationOffset, uint destinationSize, ushort fillValue);
> void wgpuMemset32ToMappedBuffer(uint destinationMappedBufferHandleId, uint destinationOffset, uint destinationSize, uint fillValue);
> void wgpuMemcpyToMappedBuffer(uint destinationMappedBufferHandleId, uint destinationOffset, uint destinationSize, void *sourcePtr);
> void wgpuMemcpyFromMappedBuffer(void *destinationPtr, uint destinationOffset, uint destinationSize, uint sourceMappedBufferHandleId);
> ```
> 
> This way the caller will be in control whether there will need to be malloc/free in effect, and avoid any redundant memsets and memcpys?

austinEng:
I think we could add these type of idea with a webgpu-native extension and not need to change the JS API.

Instead of `wgpuBufferGetMappedRange`, we have `wgpuBufferGetMappedRangeHandle`, and add his suggestions. This keeps the mapped range js-side without needing to copy it into the Emscripten heap. Then we can add his helper suggestions which work with the handle.

Having it be callback-based doesn't seem relevant to the extra memory operations. I'm guessing that part is because he's concerned about garbage?

kainino0x:
We already have mapAsync() for asynchronicity and unmap() for memory management; I don't think any more callbacks are needed.

[...]

PR #494, opened 2025-01-11
Add Read/WriteMappedRange

kainino0x:
Issue: #66
(dawn issues: https://crbug.com/389111523, https://crbug.com/389149747)

### DIFF
--- a/doc/articles/BufferMapping.md
+++ b/doc/articles/BufferMapping.md
@@ -6,8 +6,9 @@ The @ref wgpuBufferGetMappedRange, @ref wgpuBufferGetConstMappedRange, @ref wgpu
 
 - Fail (return `NULL` or `WGPUStatus_Error`) with @ref ImplementationDefinedLogging if:
     - There is any content-timeline error, as defined in the WebGPU specification for `getMappedRange()`, given the same buffer, offset, and size (buffer is not mapped, alignment constraints, overlaps, etc.)
-        - All four calls reserve a range which cannot be overlapped by any future call.
-        - Note that implementations must validate overlap rules, even if they do not have such constraints or could avoid them.
+        - **Except** for overlaps between *const* ranges, which are allowed in C *in non-Wasm targets only*.
+            (Wasm does not allow this because const ranges do not exist in JS.
+            All of these calls are implemented on top of `getMappedRange()`.)
     - @ref wgpuBufferGetMappedRange or @ref wgpuBufferWriteMappedRange is called, but the buffer is not mapped with @ref WGPUMapMode_Write.
 
 @ref wgpuBufferGetMappedRange, @ref wgpuBufferGetConstMappedRange additionally:
--- a/doc/articles/BufferMapping.md
+++ b/doc/articles/BufferMapping.md
@@ -1,17 +1,20 @@
 # Buffer Mapping {#BufferMapping}
 
-## GetMappedRange/GetConstMappedRange Behavior {#GetMappedRangeBehavior}
+## Mapped Range Behavior {#MappedRangeBehavior}
 
-@ref wgpuBufferGetMappedRange and @ref wgpuBufferGetConstMappedRange...:
+The @ref wgpuBufferGetMappedRange, @ref wgpuBufferGetConstMappedRange, @ref wgpuBufferReadMappedRange, and @ref wgpuBufferWriteMappedRange methods:
 
-- Return `NULL` with @ref ImplementationDefinedLogging if:
-    - There is any content-timeline error, as defined in the WebGPU specification for `getMappedRange()` (buffer is not mapped, alignment constraints, overlaps, etc.)
-        - **Except** for overlaps between *const* ranges, which are allowed in C *in non-Wasm targets only*.
-            (Wasm does not allow this because const ranges do not exist in JS.)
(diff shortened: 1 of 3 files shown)


doc/articles/BufferMapping.md
```diff
@@ -1,17 +1,20 @@
...
-## GetMappedRange/GetConstMappedRange Behavior {#GetMappedRangeBehavior}
+## Mapped Range Behavior {#MappedRangeBehavior}
 
-@ref wgpuBufferGetMappedRange and @ref wgpuBufferGetConstMappedRange...:
+The @ref wgpuBufferGetMappedRange, @ref wgpuBufferGetConstMappedRange, @ref wgpuBufferReadMappedRange, and @ref wgpuBufferWriteMappedRange methods:
+
+- Fail (return `NULL` or `WGPUStatus_Error`) with @ref ImplementationDefinedLogging if:
+    - There is any content-timeline error, as defined in the WebGPU specification for `getMappedRange()`, given the same buffer, offset, and size (buffer is not mapped, alignment constraints, overlaps, etc.)
+        - All four calls reserve a range which cannot be overlapped by any future call.
+        - Note that implementations must validate overlap rules, even if they do not have such constraints or could avoid them.
```
Kangz, line 10: That's a lot of overhead in native, as well as a breaking change, is there a way we could avoid it?

Kangz: The overhead of validating with a map is much bigger than doing data + offset (with some arithmetic checks) though that's a minor concern for the simple case of a single full GetMappedRange. What I'm worried about are both applications that would:

 1.  Call GetMappedRange for the entire range repeatedly just because they pass the buffer around without the pointer to the mapped range (because it's inconvenient to pass two things). They would get validation errors.
 2. Call GetMappedRange for individual elements that they want to write. The performance would degrade a bunch for them.

What I don't understand is why we need to add this validation in native. Matching the JS spec is nice, but this validation is very much JS-centric because we can't have overlapping ArrayBuffers in JS. What use, even for WASM would it be to add the validation? It's not like we can have multiple non-overlapping JS ArrayBuffers that we put in the WASM heap as 1) we don't have a way to splice an ArrayBuffer in the JS heap yet and 2) when we have, the whole buffer should be spliced in because we are doing splicing at the memory page level.

kainino0x: > 1\. Call GetMappedRange for the entire range repeatedly just because they pass the buffer around without the pointer to the mapped range (because it's inconvenient to pass two things). They would get validation errors.

If we think this is important then we could pretty easily allow this specific case in Wasm. Just have to identify that it's a duplicate and return the same pointer.

> 2\. Call GetMappedRange for individual elements that they want to write. The performance would degrade a bunch for them.

Don't we still need to track these ranges for invalidation on unmap? Doing this seems like it could be terrible for performance (or at least we'd have to do something smart in the implementation to coalesce ranges that sit near each other)

---

If we want applications to be able to do those things then I think we should try to make them possible in Wasm too. In Emscripten/emdawnwebgpu right now we only malloc and memcpy the ranges that have GetMappedRange called on them, not the whole MapAsync region.

I'm sure it's possible to lift some or all of the restrictions in Wasm if we want, but the more we lift the more complicated it gets. For just GetMappedRange/GetConstMappedRange:
- wasm-malloc the entire MapAsync range (possibly wasteful but hopefully applications can deal with this)
- in emscripten, track all ranges that have been accessed (allowing overlaps)
- each time a new range is accessed, copy its data in
- on unmap, flush all non-const ranges

Wasm mmap would fix the allocation problem since we only have to reserve pages; and assuming we have to map entire pages, invalidations will be rounded up to the nearest page, but I think that was always going to be true.

This would also interact with Read/WriteMappedRange.
From [notes](https://github.com/webgpu-native/webgpu-headers/issues/66#issuecomment-2584980077):

> LK: It is possible to make WriteMappedRange/ReadMappedRange work without overlap validation, by breaking up the memcpy into chunks

However if we want to allow overlaps between GetMappedRange and Read/WriteMappedRange we have to make sure the results are consistent, e.g. if there's an overlap, by making Read/WriteMappedRange fall back to modifying the wasm copy instead of the browser copy.

All that to say, we probably _can_ lift restrictions but it's a lot of work. I would like to make it consistent between wasm and native and I'm not sure the use-cases really need us to do all of that.

kainino0x:
Jan 23 meeting:

* LK: Could specify no overlap validation, say wasm tries to do that, but for now is out of spec  
* KN: Definitely can do it if we allocate the whole mapasync region  
* LK: Could warn if we map and unmap without having accessed most of the region  
* LK: Leaning toward allocating whole region  
* KN: Maybe leave spec as is, except say "wasm may validate overlaps", and then deal with this once we actually have someone run into this  
* We'll see what CW says

kainino0x:
Jan 30 meeting:

* CF: about requesting many small GetMappedRanges, can we tell people to not do that? Graphics programming is all about doing less things.  
* … (sorry I didn't take notes)  
* **Conclusion: Leave as it was: no validation in native, always validation in Wasm. Maybe we will add a build option later to Emscripten to lift the validation in wasm (by doing complicated stuff). Update PR to just add Read/WriteMappedRange**

kainino0x:
Revised to just add Read/WriteMappedRange. PTAL

lokokung (approved):

PR #499, opened 2025-01-16
Simplify WGPUWaitStatus "Unsupported" cases

kainino0x:
Mentioned `WGPUWaitStatus` in #401 but I didn't think about simplifying it so we didn't discuss it.

Since we're relying now on "implementation-defined logging" to convey details we don't generally need detailed error codes for things that are just surface-level misuses of the API.

In `WGPUWaitStatus` we have three `Unsupported` cases:
- `WGPUWaitStatus_UnsupportedTimeout`
- `WGPUWaitStatus_UnsupportedCount`
- `WGPUWaitStatus_UnsupportedMixedSources`

I don't think these are cases that a program would actually respond to programmatically. So I'm proposing we simplify into just one `WGPUWaitStatus_Unsupported` or `WGPUWaitStatus_Error`, with implementation-defined logging.

I used `WGPUWaitStatus_Error` here.

### DIFF
--- a/doc/articles/Asynchronous Operations.md
+++ b/doc/articles/Asynchronous Operations.md
@@ -41,7 +41,7 @@ The `callback` function pointer is called when the application _observes complet
 
 Waits on any WGPUFuture in the list of `futures` to complete for `timeoutNS` nanoseconds. Returns when at least one `WGPUFuture` is completed or `timeoutNS` elapses, whichever is first. If `timeoutNS` is zero, all `futures` are polled once, without blocking.
 
-- Returns @ref WGPUWaitStatus_Success if at least one `WGPUFuture` completes. Each future which _just_ completed has its respective callback fired. Each future which _is_ complete has its corresponding @ref WGPUFutureWaitInfo::completed is set to true.
+- Returns @ref WGPUWaitStatus_Success if at least one `WGPUFuture` completes. Each future which _just_ completed has its respective callback fired. Each future which _is_ complete has its corresponding @ref WGPUFutureWaitInfo::completed set to true.
     - Note that the timeout applies only to the "wait" phase, and does not affect the callback firing phase.
 - Returns @ref WGPUWaitStatus_TimedOut if the timeout passed without any futures completing.
 - Returns @ref WGPUWaitStatus_Error if the call was invalid for any of the following reasons:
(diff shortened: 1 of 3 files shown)


doc/articles/Asynchronous Operations.md
```diff
@@ -41,10 +41,13 @@ The `callback` function pointer is called when the application _observes complet
 
 Waits on any WGPUFuture in the list of `futures` to complete for `timeoutNS` nanoseconds. Returns when at least one `WGPUFuture` is completed or `timeoutNS` elapses, whichever is first. If `timeoutNS` is zero, all `futures` are polled once, without blocking.
 
-Returns @ref WGPUWaitStatus_Success if at least one `WGPUFuture` completes. WGPUFutureWaitInfo::completed is set to true for all completed futures. See @ref WGPUWaitStatus for other status codes.
-
-Within this call, for any `WGPUFuture`s that completed, their respective callbacks will fire.
-Note that the timeout applies only to the "wait" phase, and does not affect the callback firing phase.
+- Returns @ref WGPUWaitStatus_Success if at least one `WGPUFuture` completes. Each future which _just_ completed has its respective callback fired. Each future which _is_ complete has its corresponding @ref WGPUFutureWaitInfo::completed is set to true.
```
lokokung, line 44: nit. has its corresponding @ref WGPUFutureWaitInfo::completed set to true.

lokokung (approved):

webgpu.h
```diff
@@ -1125,24 +1125,15 @@ typedef enum WGPUWaitStatus {
...
-     * `0x00000004`.
-     * The number of futures waited on in a @ref Timed-Wait is greater than the supported WGPUInstanceFeatures::timedWaitAnyMaxCount.
-     */
-    WGPUWaitStatus_UnsupportedCount = 0x00000004,
-    /**
-     * `0x00000005`.
-     * An invalid wait was performed with @ref Mixed-Sources.
-     */
-    WGPUWaitStatus_UnsupportedMixedSources = 0x00000005,
+    WGPUWaitStatus_Error = 0x00000003,
```
kainino0x, line 1136: I wonder if we should return this if you try to wait on the null future?

Or on a future that we know is invalid (because it's too big)? (Probably not this one, I don't think we enshrined the monotonically-increasing future id allocation strategy.)

cwfitzgerald (commented):
Like this is fine, but I feel kinda bad removing information from the user

webgpu.h
kainino0x: Filed under #502

kainino0x:
If users need this at runtime I think it's better if they just have a way to check the actual capabilities and not use trial-and-error. Filed #503 about reflecting the instance capabilities.

PR #506, opened 2025-01-25
Add "mixable events" device feature

### LINKED ISSUE #505: Add a way to enable mixed-source WaitAny support
Some backends (at least Metal, D3D12, and Vulkan with some extensions) support getting OS event handles for GPU queue events, which allows mixing their queue events in with CPU events. There should be a way for adapters to say they have support for this, probably simply as a `WGPUFeatureName` that can be enabled in `RequestDevice`. This was alluded to in the old design doc https://docs.google.com/document/d/1qJRTJRY318ZEqhK6ryw4P-91FumYQfOnNP6LpANYy6A/edit?usp=sharing but I think I forgot to actually write anything out about it:

> A timeout was used, and there were two uncompleted futures from different sources (CPU, device1 queue, device2 queue), **but the corresponding feature wasn't enabled on those futures' device.**

> OnSubmittedWorkDone and MapAsync **by default** return futures which can’t be mixed with futures of other types or futures from other devices. See also WGPUWaitStatus_UnsupportedMixedSourcesWithTimeout.

kainino0x:
Fixes #505

### DIFF
--- a/doc/articles/Asynchronous Operations.md
+++ b/doc/articles/Asynchronous Operations.md
@@ -66,16 +66,17 @@ When performing a timed wait, all `WGPUFuture`s must have the same "exclusive de
 Events:
 
 - Events which come from the CPU are always "mixable events".
-    - ::wgpuInstanceRequestAdapter
-    - ::wgpuAdapterRequestDevice
-    - ::wgpuShaderModuleGetCompilationInfo
-    - ::wgpuDeviceCreateRenderPipelineAsync
-    - ::wgpuDeviceCreateComputePipelineAsync
-    - ::wgpuDevicePopErrorScope
+    - @ref wgpuInstanceRequestAdapter
+    - @ref wgpuAdapterRequestDevice
+    - @ref wgpuShaderModuleGetCompilationInfo
+    - @ref wgpuDeviceCreateRenderPipelineAsync
+    - @ref wgpuDeviceCreateComputePipelineAsync
+    - @ref wgpuDevicePopErrorScope
+    - Device loss (@ref WGPUDeviceDescriptor::deviceLostCallbackInfo + @ref wgpuDeviceGetLostFuture)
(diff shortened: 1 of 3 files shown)


kainino0x:
@lokokung WDYT?

doc/articles/Asynchronous Operations.md
```diff
@@ -55,21 +55,27 @@ Use of _timed waits_ (`timeoutNS > 0`), must be enabled on the WGPUInstance in @
...
+
+Events:
+
+- Events which come from the CPU are always "mixable events".
+    - ::wgpuInstanceRequestAdapter
+    - ::wgpuAdapterRequestDevice
+    - ::wgpuShaderModuleGetCompilationInfo
+    - ::wgpuDeviceCreateRenderPipelineAsync
+    - ::wgpuDeviceCreateComputePipelineAsync
+    - ::wgpuDevicePopErrorScope
```
lokokung, line 74: Also device lost.

lokokung (approved):

lokokung (approved):

kainino0x:
Closing in favor of prototyping in Dawn/Emscripten first as a Dawn extension. Not likely wgpu-native will implement it soon.

kainino0x:
Jan 30 meeting:

* LK: Looks good
* CF: I don't expect us to ever implement this. Hard to tell anything right now though because haven't heard from Rajesh in a while. wgpu-py person is kind of maintaining wgpu-native, and I do drive-by stuff, but it's not being maintained really.  
* KN: We can do it as a Dawn extension. Also will want to implement it in Emscripten. Probably.  
* CF: Does Emscripten implement any of the Dawn-specific API?  
* KN: Generally no. Only things like chromium-experimental-subgroups that we need to pass through.  
* **Dawn extension, may implement in Emscripten.**  
* LK: (something about using this in Chromium)
* KN: Chromium may use the fd-exporting extensions. TBD  
* LK: Not sure, might be harder to maintain.

PR #507, opened 2025-01-27
Restore WGPUQueueWorkDoneStatus_Error

kainino0x:
We know this will be useful, and adding it now avoids implementations needing to add it as an extension whenever they prototype multi-queue (which would be annoying because then we'd end up with a bunch of different values for this one status code).

Issue: #401
Partially reverts #464


lokokung (approved):

kainino0x:
Jan 30 meeting:

* KN: summary
* **CF: Sure.**

PR #508, opened 2025-01-29
Add WGPURequestAdapterWebXROptions.xrCompatible

### LINKED ISSUE #501: GPURequestAdapterOptions.xrCompatible
We don't have a way to set GPURequestAdapterOptions.xrCompatible.

This probably should be a Wasm extension because it's not really meaningful outside the web. In the Web there's a known XR API (WebXR), but in native you could be interfacing with anything.

kainino0x:
Fixes #501

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -221,6 +221,7 @@ struct WGPURenderBundleEncoderDescriptor;
 struct WGPURenderPassDepthStencilAttachment;
 struct WGPURenderPassMaxDrawCount;
 struct WGPURequestAdapterOptions;
+struct WGPURequestAdapterWebXROptions;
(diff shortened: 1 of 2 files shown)


lokokung (approved):

kainino0x:
Jan 30 meeting:
* Looks good  
* Why a chain and a boolean?  
* KN: I feel like it's weird to have the presence of an extension actually do something, but no really strong feelings  
* **Looks OK.**

kainino0x:
Oh also:

- KN: Should compatibleSurface be moved to an extension? Is this just for WebGL?  
- **CF: No, it's also used on Vulkan. And might use it in the future for GL.**

Issue #369, opened 2024-10-10
popErrorScope status when stack is empty?


kainino0x:
In JS when the error scope stack is empty, popErrorScope rejects with OperationError. What should we do in C?

Emscripten and Dawn currently translate it to `Unknown` plus a message string. However we also use `Unknown` in Emscripten if the error type is some unexpected object (e.g. if we add one to the JS API in the future).

We should also say what the message string is. I think it **must** be empty if the status is NoError and it **should** be non-empty for any other status. (since message can't be null anymore)

lokokung:
Hmm, that's a good point. In C, there's only `Success` and `InstanceDropped` on `WGPUPopErrorScopeStatus`. Maybe we could add another case there instead of modifying the error type?

kainino0x:
You're right, using the status makes more sense. Also, we shouldn't add it to `WGPUErrorType`, since that's also used in `WGPUUncapturedErrorCallback`.

We have to decide what values you get when the status is something other than `Success`.
- `WGPUStringView message` is empty string
- `WGPUErrorType type` is... `NoError=1`? `Unknown=5`? Add a new `Undefined=0` and return that? Renumber `NoError` to `0`? Rename `NoError=1` to `Undefined=0` and return that?

The first half of this seems pretty trivial so I'm going to just open a PR for it.

kainino0x:
It should not be `Unknown` because that value implies there will be a `message` string.

kainino0x:
Tentatively closing as I did my proposal in #393, but leaving the label to discuss if we want to reopen.

kainino0x:
Nov 21 meeting:

*   KN: This is a promise rejection in JS. I already added `WGPUPopErrorScopeStatus_EmptyStack` (in addition to `Success` and `InstanceDropped`.
    *   (we can't add it to `WGPUErrorType`, that's also used in `WGPUUncapturedErrorCallback`)
*   **Sounds good**
*   When the status is `EmptyStack` or `InstanceDropped`, what should `WGPUErrorType type` be? I tentatively said `NoError` since that matches with `message` being empty string.
*   **NoError because you captured no error (it's always the case that status==Success || type=NoError). But change string to be allowed to be non-empty? (The string is dual purpose: EITHER the error from the status OR the error from the captured error)**
*   **Rename EmptyStack to Error or ValidationError or something**

PR #519, opened 2025-03-10
Renames PopErrorScopeStatus Error to EmptyStack.

lokokung:
Fixes: https://github.com/webgpu-native/webgpu-headers/issues/369

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -634,7 +634,7 @@ typedef enum WGPUPopErrorScopeStatus {
     /**
      * The error scope stack could not be popped, because it was empty.
      */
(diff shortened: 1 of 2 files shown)


cwfitzgerald (approved):
STATUS_OK_STACK_UNDERFLOW

kainino0x:
I thought we decided to rename EmptyStack to Error so that it would be more generic? #369 is already closed.

lokokung:
Oh, hmm, that might be true actually... Lemme dig through the notes to verify then. Could definitely be outdated on Dawn side instead of this one.

Issue #520, opened 2025-03-14
Futures should be individually cancellable


kainino0x:
Futures can currently be cancelled only by dropping the instance. It would be useful to be able cancel them individually.

As an example, if you write a C++ RAII wrapper around a `WGPUFuture` (say, `myapp::WebGPUFuture<wgpu::ComputePipeline>`) and you drop it, there's no way to ensure the callback gets called except to wait for the future. If we had a way to cancel futures like `void wgpuInstanceCancelFuture(WGPUFuture)`[^return], you could do that instead - basically, call the callback immediately and ignore/abort any ongoing work.

Right now we already have the `InstanceDropped` status, I think we could genericize this to `Canceled`[^spelling] and use it for this too.

[^return]: This function could have a return value indicating whether it actually got canceled, though it seems probably not necessary because you'll get the status via the callback.
[^spelling]: though I dislike this word because its spelling is ambiguous, similar to adapter vs adaptor

lokokung:
I like the idea of using `Cancelled` instead of `InstanceDropped`! Do you think we can just move forwards with this, and I'll start implementing it? Or did you want to wait and discuss?

I think the rename of the enum could probably be fine without further discussion, but maybe whether we have `CancelFuture` we can wait and discuss since that would be non-breaking anyways?

kainino0x:
I agree, having a more generic name is better and we should just go ahead with it. Though I'd love if we had word that didn't have ambiguous spelling.

Based on ngrams I'd probably choose `Cancelled` despite many claims that `Canceled` is the American spelling (I don't think that's really true).
https://books.google.com/ngrams/graph?content=canceled%2Ccancelled&year_start=1800&year_end=2022&corpus=en&smoothing=3&case_insensitive=false

Kangz:
Strong +1 on this proposal.

kainino0x:
Mar 20 meeting:

* KN: Cleanup a future without dropping the whole instance.  
* CF: Don't understand why the described RAII wrapper should work like that. This would cause the callback to be called in the destructor. Possibly dangerous if using exceptions.  
* LK: TBH don't understand either.  
* CF: Seems like improper usage of the API. [Don't see when you would actually need to do this.]
* …  
[...]

PR #521, opened 2025-03-20
Rename InstanceDropped to Cancelled for all enums

kainino0x:
Issue: #520

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -432,6 +432,9 @@ typedef enum WGPUCompareFunction {
 
 typedef enum WGPUCompilationInfoRequestStatus {
     WGPUCompilationInfoRequestStatus_Success = 0x00000001,
+    /**
+     * See @ref CallbackStatuses.
+     */
     WGPUCompilationInfoRequestStatus_CallbackCancelled = 0x00000002,
     WGPUCompilationInfoRequestStatus_Force32 = 0x7FFFFFFF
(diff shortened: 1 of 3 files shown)


webgpu.h
```diff
@@ -492,7 +492,7 @@ typedef enum WGPUCullMode {
 typedef enum WGPUDeviceLostReason {
     WGPUDeviceLostReason_Unknown = 0x00000001,
     WGPUDeviceLostReason_Destroyed = 0x00000002,
-    WGPUDeviceLostReason_InstanceDropped = 0x00000003,
+    WGPUDeviceLostReason_Cancelled = 0x00000003,
```
kainino0x, line 495: This one is a little awkward. Did this mean the device was lost because the instance was dropped, or just that we _called the device lost callback_ because the instance was dropped?

kainino0x: Should we call them all CallbackCancelled or something so the meaning is clearer regardless of context?

lokokung: This is for the device lost callback was called because instance was dropped, i.e. we set device lost callback mode to `ProcessEvents` and instance gets dropped. I'm fine with `CallbackCancelled` instead.

kainino0x: Renamed.

lokokung (approved):

lokokung:
Actually, maybe we should add docs for the enum while we are at it?

kainino0x:
Sure, will do before landing.

kainino0x:
Oops, landed with an outdated commit message :)

PR #523, opened 2025-03-28
Update generator to use prefixes.

lokokung:
Fixes: #212

Previous work was done in https://github.com/webgpu-native/webgpu-headers/pull/462, but this change improves on the prototype.

- Updates generator and template to use prefixes for extensions instead of suffixes.
- Adds new `Base` embedded struct in the Go representation of the YAML to facilitate name generation.
- Adds a series of new helpers in `gen.go` to use the new `Base` struct to generate names properly.
- Adds minimal extension YAML along to showcase generator results and updates Makefile for testing.
- Some minor formatting changes to the templates so that newlines are more consistent. This was much easier to verify in the smaller extension header.

### DIFF
--- a/gen/gen.go
+++ b/gen/gen.go
@@ -455,7 +455,7 @@ func (g *Generator) FunctionArgs(f Function, o *Object) string {
 		}
 	}
 	if f.Callback != nil {
-		fmt.Fprintf(sb, ", %s callbackInfo", g.CType(*f.Callback, "")+"CallbackInfo")
+		fmt.Fprintf(sb, ", %sCallbackInfo callbackInfo", g.CType(*f.Callback, ""))
 	}
 	return sb.String()
 }
@@ -605,7 +605,7 @@ func (g *Generator) StructMember(s Struct, memberIndex int) (string, error) {
 		sb.WriteString("WGPU_NULLABLE ")
 	}
 	if strings.HasPrefix(member.Type, "callback.") {
-		fmt.Fprintf(sb, "%s %s;", g.CType(member.Type, "")+"CallbackInfo", CamelCase(member.Name))
+		fmt.Fprintf(sb, "%sCallbackInfo %s;", g.CType(member.Type, ""), CamelCase(member.Name))
 	} else {
 		fmt.Fprintf(sb, "%s %s;", g.CType(member.Type, member.Pointer), CamelCase(member.Name))
 	}
--- a/gen/gen.go
+++ b/gen/gen.go
@@ -5,13 +5,14 @@ import (
 	"io"
 	"math"
 	"reflect"
+	"slices"
 	"strconv"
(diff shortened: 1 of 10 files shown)


tests/extensions/webgpu_extension.h
```diff
@@ -0,0 +1,464 @@
...
+_wgpu_EXTEND_ENUM(WGPUOldEnum, WGPUOldEnum_PrefixNewValue, 0xFFFF0000);
+
+/**
+ * New enum should have the namespace prefix in the name by default.
+ */
+typedef enum WGPUPrefixNewEnum1 {
+    /**
+     * New enum entries for a new enum should not duplicate prefix.
+     */
+    WGPUPrefixNewEnum1_NewValue = 0xFFFF0000,
```
kainino0x, line 109: This is larger than Force32, but we shouldn't allow that. In https://github.com/webgpu-native/webgpu-headers/blob/main/doc/articles/Extensions.md#registry-of-prefixes-and-enum-blocks it's noted enums are 32-bit and we reserve negative values.

(Enums are usually signed, so this might actually make the enum become a uint32_t or signed int64_t? It's implementation-defined.)

gen/gen.go
```diff
@@ -336,31 +433,29 @@ func (g *Generator) FunctionReturns(f Function) string {
...
-			fmt.Fprintf(sb, "%s %s", g.CType(arg.Type, arg.Pointer, typeSuffix), CamelCase(arg.Name))
+			fmt.Fprintf(sb, "%s %s", g.CType(arg.Type, arg.Pointer), CamelCase(arg.Name))
 		}
 		if i != len(f.Args)-1 {
 			sb.WriteString(", ")
 		}
 	}
 	if f.Callback != nil {
-		fmt.Fprintf(sb, ", %s callbackInfo", g.CType(*f.Callback, "", "Info"))
+		fmt.Fprintf(sb, ", %s callbackInfo", g.CType(*f.Callback, "")+"CallbackInfo")
```
kainino0x, line 458: nit: I think I would expect spaces around this `+`. Have you run gofmt? (I forget if we have a presubmit check for gofmt)

kainino0x (approved):
Sorry for the huge delay. This looks great! Just two small comments

tests/extensions/webgpu_extension.h
lokokung: Nice catch. Turned out to be a bug in the schema. The schema allowed the `enum_prefix` to be up to `2^16-1` which effectively allows setting the sign bit. Updated to only allow the maximum to be `2^15-1`.

gen/gen.go
lokokung: I have autoformatting on, so surprisingly, it actually prefers the no space version. I ended up rolling the constant string into the format string for this instead.

PR #530, opened 2025-04-25
[breaking] Add binding arrays to WGPUBindGroupLayoutEntry

### LINKED ISSUE #387: [try for v1] Add binding arrays to API
This doesn't strictly have to be in V1 but it would be nice if it can be.

Upstream proposal: https://github.com/gpuweb/gpuweb/blob/main/proposals/sized-binding-arrays.md

kainino0x:
Based on https://github.com/gpuweb/gpuweb/blob/main/proposals/sized-binding-arrays.md

- [x] but with: https://github.com/gpuweb/gpuweb/pull/5174

Fixes #387
(dawn issue https://crbug.com/415092021)

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -3626,6 +3626,12 @@ typedef struct WGPUBindGroupLayoutEntry {
      */
     uint32_t binding;
     /**
+     * The number of bindings (size of the binding array) defined by this entry.
+     *
+     * The `INIT` macro sets this to `1`.
+     */
+    uint32_t bindingCount;
+    /**
      * The `INIT` macro sets this to @ref WGPUShaderStage_None.
      */
     WGPUShaderStage visibility;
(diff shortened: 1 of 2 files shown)


webgpu.h
```diff
@@ -3625,6 +3625,12 @@ typedef struct WGPUBindGroupLayoutEntry {
      * The `INIT` macro sets this to `0`.
      */
     uint32_t binding;
+    /**
+     * The number of bindings (size of the binding array) defined by this entry.
+     *
+     * The `INIT` macro sets this to `1`.
+     */
+    uint32_t bindingCount;
```
Kangz, line 3633: Could we have this set to 0 meaning not a binding array? The JS spec can decide what 0 means (vs undefined) separately.

Also I don't think the rename is an improvement, see gpuweb comment.

lokokung (approved):

PR #532, opened 2025-04-25
[technically breaking] Auto-destroy Device when the last ref is dropped

### LINKED ISSUE #514: Destroying objects when external ref reaches zero
For the four destroyable objects (device, buffer/texture/queryset) we should explicitly document when destroy will be automatically called on them - whether we do it when the external refcount reaches zero, or when the internal refcount reaches zero.

I think it doesn't make sense to auto-destroy resources (buffer/texture/queryset) because you can bind them into a bindgroup or command buffer and then drop them.

However for device it kind of makes sense. There's not much you can do without an external ref to the device. In Dawn we already do this - we destroy the device when you drop the last external ref.

We should make sure that's what we want the API to be, and document it.

### LINKED ISSUE #531: Device passed to DeviceLostCallback may have already been released
> * KN: Related but orthogonal question, if you call drop the last ref, then (external) refcount is already zero, but then we pass the device to the DeviceLost callback. How does that work?  
> * (discussion)  
>   * The use case was for looking it up in a map. Should we give a uintptr\_t or void\* instead? Would give a strong signal not to try to use it. Though people will probably still use WGPUDevice in their map so would need to cast it back.  
>   * Should we tell people to use the DeviceLost future ID as their key instead, like we considered at some point, and pass that to the callback instead? (Or not pass it to the callback, and tell them to store that themselves?)  
>   * KN: **I need to go back to past discussions and read what we said about this topic.** 

 _Originally posted by @kainino0x in [#514](https://github.com/webgpu-native/webgpu-headers/issues/514#issuecomment-2829183305)_

EDIT: Relevant past discussions:
- #251

kainino0x:
This is _slightly_ different from Dawn's current implementation, in the DeviceLost callback behavior. Not sure about wgpu-native but don't think it will be a significant change there either.

Fixes #514
Fixes #531
(dawn issue https://crbug.com/415094789, obsoletes https://crbug.com/409086866)

### DIFF
--- a/doc/articles/Ownership.md
+++ b/doc/articles/Ownership.md
@@ -28,7 +28,7 @@ Note also that some structs with `FreeMembers` functions may be used as both inp
 
 Unlike other destroyable objects, releasing the last (external) ref to a device causes it to be automatically destroyed (via @ref wgpuDeviceDestroy) if it isn't already lost or destroyed. Though the device object is no longer valid to use after releasing the last ref, this has some notable direct effects:
 
-- It destroy buffers, unmapping them, aborting any pending map requests, and preventing future ones.
+- Destroying a device destroys its buffers, which will unmap them, abort any pending map requests, and prevent future map requests.
 - It fires the DeviceLost event (which will call the registered callback at some point, depending on its @ref WGPUCallbackMode).
     - Because the device's last ref has already been released, DeviceLost callbacks triggered in this way will *not* receive a pointer to the device object. This is still true if the callback was triggered *during* the release of the last ref (via @ref WGPUCallbackMode_AllowSpontaneous).
 
--- a/doc/articles/Ownership.md
+++ b/doc/articles/Ownership.md
@@ -26,10 +26,11 @@ Note also that some structs with `FreeMembers` functions may be used as both inp
 
 ## Releasing a Device object {#DeviceRelease}
 
-Unlike other destroyable objects, releasing the last (external) ref to a device causes it to be automatically destroyed (via @ref wgpuDeviceDestroy). Though the device object is no longer valid to use after dropping the last ref, this has some direct effects:
+Unlike other destroyable objects, releasing the last (external) ref to a device causes it to be automatically destroyed (via @ref wgpuDeviceDestroy) if it isn't already lost or destroyed. Though the device object is no longer valid to use after releasing the last ref, this has some notable direct effects:
 
-- Buffers are destroyed, aborting any pending map requests, and preventing future ones.
-- The @ref WGPUDevice object passed to @ref WGPUDeviceLostCallback is still a valid pointer, but since the refcount has reached 0, the object may not be used for any API operation.
+- It destroy buffers, aborting any pending map requests and preventing future ones.
+- It fires the DeviceLost event (which will call the registered callback at some point, depending on its @ref WGPUCallbackMode).
+    - Because the device's last ref has already been released, DeviceLost callbacks triggered in this way will *not* receive a pointer to the device object. This is still true if the callback was triggered *during* the release of the last ref (via @ref WGPUCallbackMode_AllowSpontaneous).
 
 ## Callback Arguments {#CallbackArgs}
 
--- a/doc/articles/Ownership.md
+++ b/doc/articles/Ownership.md
@@ -24,6 +24,13 @@ Note that such functions will *not* free any previously-allocated data: overwrit
 
 Note also that some structs with `FreeMembers` functions may be used as both inputs and outputs. In this case `FreeMembers` must only be used if the member allocations were made by the `webgpu.h` implementation (as an output), not if they were made by the application (as an input).
 
+## Releasing a Device object {#DeviceRelease}
+
+Unlike other destroyable objects, releasing the last (external) ref to a device causes it to be automatically destroyed (via @ref wgpuDeviceDestroy). Though the device object is no longer valid to use after dropping the last ref, this has some direct effects:
+
+- Buffers are destroyed, aborting any pending map requests, and preventing future ones.
+- The @ref WGPUDevice object passed to @ref WGPUDeviceLostCallback is still a valid pointer, but since the refcount has reached 0, the object may not be used for any API operation.
+
 ## Callback Arguments {#CallbackArgs}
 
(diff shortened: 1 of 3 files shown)


doc/articles/Ownership.md
```diff
@@ -24,6 +24,14 @@ Note that such functions will *not* free any previously-allocated data: overwrit
 
 Note also that some structs with `FreeMembers` functions may be used as both inputs and outputs. In this case `FreeMembers` must only be used if the member allocations were made by the `webgpu.h` implementation (as an output), not if they were made by the application (as an input).
 
+## Releasing a Device object {#DeviceRelease}
+
+Unlike other destroyable objects, releasing the last (external) ref to a device causes it to be automatically destroyed (via @ref wgpuDeviceDestroy) if it isn't already lost or destroyed. Though the device object is no longer valid to use after releasing the last ref, this has some notable direct effects:
+
+- It destroy buffers, unmapping them, aborting any pending map requests, and preventing future ones.
+- It fires the DeviceLost event (which will call the registered callback at some point, depending on its @ref WGPUCallbackMode).
+    - Because the device's last ref has already been released, DeviceLost callbacks triggered in this way will *not* receive a pointer to the device object. This is still true if the callback was triggered *during* the release of the last ref (via @ref WGPUCallbackMode_AllowSpontaneous).
```
kainino0x, line 33: This last bit is a bit funky because it means it doesn't literally behave the same way as calling `wgpuDeviceDestroy` (which, since it's called on a device, means the device must necessarily still have refs). I would choose one of the following, in order of preference:

- Lose the device instead of destroying it. This would have the additional effect not unmapping already-mapped buffers.
- Change this bit this so that it will be non-null if called *during* the Release call.
- Describe this as destroying the device but without literally saying it calls `wgpuDeviceDestroy()`. Handwavy.

doc/articles/Ownership.md
```diff
@@ -24,6 +24,14 @@ Note that such functions will *not* free any previously-allocated data: overwrit
 
 Note also that some structs with `FreeMembers` functions may be used as both inputs and outputs. In this case `FreeMembers` must only be used if the member allocations were made by the `webgpu.h` implementation (as an output), not if they were made by the application (as an input).
 
+## Releasing a Device object {#DeviceRelease}
+
+Unlike other destroyable objects, releasing the last (external) ref to a device causes it to be automatically destroyed (via @ref wgpuDeviceDestroy) if it isn't already lost or destroyed. Though the device object is no longer valid to use after releasing the last ref, this has some notable direct effects:
+
+- It destroy buffers, unmapping them, aborting any pending map requests, and preventing future ones.
```
lokokung, line 31: nit. This reads a bit weird because we are effectively saying it destroys buffers, which means unmapping/aborting/preventing.

Maybe:
"It destroys buffers which will unmap them, abort any pending map requests, and prevent future map requests."

doc/articles/Ownership.md
lokokung: When you say lose the device, would that be a case where objects end up keeping the device alive? Like just the device lost callback gets fired, but everything else just keeps continue working?

kainino0x: No, I mean it would be like a "natural" device loss, like a backend-initiated device loss or unexpected OOM. The device is still lost and nothing works, the only difference in the operation itself is that it doesn't destroy buffers.

I poorly explained why that's relevant here. The behavior I want, for consistency, is for Release to do [drop last ref] [trigger destroy or loss], in that order. The problem with that is that you _can't_ (at the API level) trigger destroy without a ref. Whereas device loss can happen at any time.

lokokung: Sorry, I may have forgotten already if we talked about it, but why can't dropping the last ref just look like it calls `wgpuDeviceDestory` again? It doesn't have to be handwavy because it could just do the same things? The only caveat would be the difference in the device lost callback right?

kainino0x: That's essentially what the current revision of the PR does.

Since Destroy can't be called without a ref, it's implied that Release looks like:
`if (externalRefCount == 1) { Destroy(); externalRefCount--; }`
- If the DeviceLost callback is non-spontaneous, the callback will get called some time later:
  `Release() -> Destroy()`
  `ProcessEvents()/WaitAny() -> callback`
  When the callback is called, externalRefCount==0.
- If it *is* spontaneous, the callback will get called during Release:
  `Release() -> Destroy() -> callback`
  When the callback is called, externalRefCount==1. Hence the special case in the PR to say this still passes null to the DeviceLost callback, even though it still technically has a ref.

lokokung: Ok, yea that sgtm then!

lokokung (approved):

doc/articles/Ownership.md
kainino0x: It's kinda weird though. The implementation would be one of the following:
- Take internal ref; free external ref; if externalRefCount==0, call Destroy(); free internal ref.
  (And when calling the DeviceLost callback, we have to choose whether to pass null based on the externalRefCount of the device, which is also weird)
- Internally use a different signal than Destroy(), so that we know to pass null regardless of any refs on the device.

Is it better to do one of the alternatives I mentioned above?

kainino0x: > * Internally use a different signal than Destroy(), so that we know to pass null regardless of any refs on the device.

Actually this makes me realize there's another weird case: the device is lost naturally (the future becomes Ready), then the last external ref is dropped, then it calls the DeviceLost callback.
- Do we pass null in this case too? (How do we know to do that? I think the only way is by looking at the external refcount.)
- Or does DeviceLost becoming Ready hold an external ref to the device so that this can't happen?

lokokung (approved):

kainino0x:
May 1 meeting:
* CF: Just pass a pointer that you're not allowed to use, like we were saying last week?  
* KN: … problems with that …  
* …  
* CF: Can make the user either hold a ref in the userdata, or hold it in their side map, if they want it.  
* CF: Null sounds good  
* KN: Explanation of PR  
* LK: Could pass an external device ref to the callback. Since if it's not null, it's guaranteed there's already an external ref.  
* KN: In C, that means the callback has to free it.
* **Keep PR as is, approved**

PR #533, opened 2025-04-25
[breaking] Refactor Instance capabilities into features/limits, add SPIRV feature

### LINKED ISSUE #522: SPIRV and WGSL support detection
Two questions regarding checking support for SPIRV and WGSL:
1. How can I check that using a WGPUShaderSourceSPIRV in WGPUShaderModuleDescriptor is supported? Or can I expect all implementations to implement this? I see no feature flags for this in webgpu.h and wgpu-native panics when trying to use it. Dawn works, but it seems like adding a feature flag for this would make it more safer to use.
2. I'm assuming that WGPUShaderSourceWGSL is always is supported by implementations when running in a browser, is this also true when running natively on desktop with something like Dawn or wgpu-native?

---

EDIT: Todo list

- [ ] > I did NOT rename WGPUFeatureName, WGPUSupportedFeatures, and WGPULimits because they're used for both Adapter and Device, so picking one could be confusing. That can be a separate step anyway.

kainino0x:
- [ ] I did NOT rename WGPUFeatureName, WGPUSupportedFeatures, and WGPULimits because they're used for both Adapter and Device, so picking one could be confusing. That can be a separate step anyway.
- [x] `WGPUWGSLLanguageFeatureName` could be flattened into `WGPUInstanceFeatureName` (EDIT: but they're different because they're not disableable, so these would be "auto-enabled"). I haven't done this here.

Issue #522

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -220,7 +220,7 @@ struct WGPUCompilationMessage;
 struct WGPUConstantEntry;
 struct WGPUExtent3D;
 struct WGPUFuture;
-struct WGPUInstanceCapabilities;
+struct WGPUInstanceLimits;
 struct WGPULimits;
 struct WGPUMultisampleState;
 struct WGPUOrigin3D;
@@ -243,6 +243,7 @@ struct WGPUShaderSourceWGSL;
 struct WGPUStencilFaceState;
 struct WGPUStorageTextureBindingLayout;
 struct WGPUSupportedFeatures;
+struct WGPUSupportedInstanceFeatures;
 struct WGPUSupportedWGSLLanguageFeatures;
 struct WGPUSurfaceCapabilities;
 struct WGPUSurfaceColorManagement;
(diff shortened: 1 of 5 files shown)


Kangz (approved):

lokokung (approved):

webgpu.h
```diff
@@ -5501,15 +5553,23 @@ typedef void (*WGPUProcTextureViewRelease)(WGPUTextureView textureView) WGPU_FUN
...
  *
  * @returns
  * Indicates if there was an @ref OutStructChainError.
  */
-WGPU_EXPORT WGPUStatus wgpuGetInstanceCapabilities(WGPUInstanceCapabilities * capabilities) WGPU_FUNCTION_ATTRIBUTE;
+WGPU_EXPORT WGPUStatus wgpuGetInstanceLimits(WGPUInstanceLimits * limits) WGPU_FUNCTION_ATTRIBUTE;
+/**
+ * Check whether a particular @ref WGPUInstanceFeatureName is supported by the instance.
+ */
+WGPU_EXPORT WGPUBool wgpuHasInstanceFeature(WGPUInstanceFeatureName feature) WGPU_FUNCTION_ATTRIBUTE;
```
cwfitzgerald, line 5572: Why do we have both a singular check function and a "return the features so you can check yourself" function?

kainino0x: I copied the pattern from WGPUAdapter. IIRC the motivations for that were (1) convenience, and (2, somewhat theoretical, and besides not at all relevant here, or in GPUDevice for that matter which also has this) to allow people to reduce their fingerprinting footprint in web browsers (via wasm) - avoid loading all the identifiable adapter info just to check for the presence of individual things.

kainino0x:
> * I did NOT rename WGPUFeatureName, WGPUSupportedFeatures, and WGPULimits because they're used for both Adapter and Device, so picking one could be confusing. That can be a separate step anyway.

@cwfitzgerald given this, how strongly do you feel about renaming them, and how would you name them?

kainino0x:
May 1 meeting:
* **Looks good, separate PR for renames**

PR #534, opened 2025-04-25
Remove WGPUFeatureName_Undefined

kainino0x:
We decided that we would not name this:
https://github.com/webgpu-native/webgpu-headers/issues/45#issuecomment-1776319835

But I missed it in #364.

Technically breaking, but since it didn't do anything, any references to it are going to be trivial (e.g. `case` statements).

Issue: #45
(no dawn bug, dawn already doesn't have it)


cwfitzgerald (approved):
0

Issue #537, opened 2025-05-01
Is buffer usage `MapWrite` when setting `mappedAtCreation` on?


eliemichel:
I am surprised that not specifying `MapWrite` does not prevent the use of `mappedAtCreation`. Is this an implementation issue (testing with Dawn 7069) or the expected standard behavior? In the latter case, we should document it in the docstrings of `mappedAtCreation` and `WGPUBufferUsage_MapWrite` because the wording suggests that these are related.

lokokung:
I believe this was intentional. The `mappedAtCreation` bool is essentially to allow a way for users to initialize data via a map. They will be able to `getMappedRange`, and `unmap`, but once they `unmap`, they will not be allowed to map it again.

Issue #473, opened 2024-12-21
Directly reference upstream spec for each item in webgpu.h


kainino0x:
Either with a link to the JS spec, or by embedding documentation from the spec into the generated header.

_Originally posted by @kainino0x in https://github.com/webgpu-native/webgpu-headers/issues/25#issuecomment-2050972751_: 
> Apr 11 meeting:
> 
> *   doc jam - how should we do this?
> *   KN: add links to spec in yaml?
>     *   pull text from it at codegen time? (but the names will differ)
>     *   or just generate links to the spec?
> *   **Generator pulling from spec is worth exploring - from html**
> *   **Explore having the spec repo itself generate some parseable (e.g. json) documentation out of the spec, similar to what types generator does today — then use this for both typescript and C**

PR #538, opened 2025-05-01
Document the buffer usage enum

eliemichel:
In particular this addresses #537 if `mappedAtCreation` indeed does not require `MapWrite` usage.


kainino0x (approved):
I didn't prioritize documenting any of this since it's the same as the JS API (and we have #473), but regardless LGTM!

PR #539, opened 2025-05-01
Rename Feature/Limit to DeviceFeature/DeviceLimit

### LINKED ISSUE #522: SPIRV and WGSL support detection

kainino0x:
As was proposed in the meeting. I'm still not really convinced this is worthwhile, but it is a little clearer (especially in C where you don't get much type safety - you can freely assign a `WGPUFeatureName` value to a `WGPUInstanceFeatureName` variable).

~We could also consider just removing the old names. The migration is probably not so hard that this is necessarily worth keeping around.~ I did add a macro to disable the aliases though.

Also removes WGPU_TRUE/WGPU_FALSE from extension headers. Accidentally added those, they aren't needed since extensions include `webgpu.h`.

Fixes #522

### DIFF
--- a/webgpu.yml
+++ b/webgpu.yml
@@ -42,12 +42,12 @@ constants:
   - name: limit_u32_undefined
     value: uint32_max
     doc: |
-      For `uint32_t` limits, indicates no limit value is specified. For more info,
+      For `uint32_t` limits, indicates no limit value is requested. For more info,
       see @ref SentinelValues and the places that use this sentinel value.
   - name: limit_u64_undefined
     value: uint64_max
     doc: |
-      For `uint64_t` limits, indicates no limit value is specified. For more info,
+      For `uint64_t` limits, indicates no limit value is requested. For more info,
       see @ref SentinelValues and the places that use this sentinel value.
   - name: mip_level_count_undefined
     value: uint32_max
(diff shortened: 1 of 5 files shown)


webgpu.yml
```diff
@@ -434,7 +499,8 @@ enums:
           "Core" profile which can be supported on Vulkan/Metal/D3D12 (at least).
   - name: feature_name
     doc: |
-      TODO
+      @deprecated alias for @ref WGPUDeviceFeatureName.
+      New feature names will NOT be added here.
```
kainino0x, line 503: Best I could come up with for how to keep the aliases. (I could modify the generator so it can make enums alias other enums, but then there's an ordering dependency too, so this is much easier.)

lokokung:
Tbh, I'm not convinced it's worth it to rename these either... Unless we feel strongly about it, I'm heavily leaning towards just leaving it as is. I think that given the `InstanceFeatures` have the explicit name in them it should be clear enough.

J2085isa (approved):
Espero que no se salgan de la misión ni la visión de la colectividad y gestión de humanidad cambio de culturas y reconocimiento de raíz natales saludos

lokokung (approved):

J2085isa (approved):

PR #540, opened 2025-05-06
Add instance feature for reusable adapters

### LINKED ISSUE #535: Reusable adapters?
In the WebGPU spec, adapters are not reusable. Creating a device consumes the adapter. This was done to make sure that people check adapters again if they need to reinitialize.

In https://dawn-review.googlesource.com/c/dawn/+/237274/comment/a9335ef5_172b283b/ it was suggested that this shouldn't be true in webgpu.h (unless the backend is Wasm), or it should be configurable (and we validate it on Wasm).

cc @Kangz

kainino0x:
Fixes #535

### DIFF
--- a/webgpu.yml
+++ b/webgpu.yml
@@ -536,6 +536,11 @@ enums:
         doc: |
           Enable passing SPIR-V shaders to @ref wgpuDeviceCreateShaderModule,
           via @ref WGPUShaderSourceSPIRV.
+      - name: reusable_adapters
+        doc: |
+          Normally, a @ref WGPUAdapter can only create a single device.
+          If this is available and enabled, then adapters won't immediately expire
+          when they create a device. They may still expire for other reasons.
   - name: load_op
     doc: |
       TODO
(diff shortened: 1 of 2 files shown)


webgpu.yml
```diff
@@ -536,6 +536,11 @@ enums:
         doc: |
           Enable passing SPIR-V shaders to @ref wgpuDeviceCreateShaderModule,
           via @ref WGPUShaderSourceSPIRV.
+      - name: reusable_adapters
```
kainino0x, line 539: I just picked a name, but other suggestions welcome.

lokokung (approved):

webgpu.yml
cwfitzgerald: Adapter expiry?

kainino0x: Adapters could still "expire" in the sense that they got unplugged or something, which makes that naming kind of ambiguous. (Actually, the upstream spec uses both "consumed" and "expired" as separate things:
https://gpuweb.github.io/gpuweb/#dom-adapter-state-slot)
Regardless the most precise thing I could think of was `NoConsumeAdapter` or something but then we start getting double negatives ("NoConsumeAdapter is disabled") which is confusing.

lokokung: `MultiDeviceAdapter` maybe?

kainino0x: That's a good direction. Updated with `MultipleDevicesPerAdapter`, I think that gets the point across pretty well.

cwfitzgerald (approved):
Sure!

lokokung (approved):

PR #541, opened 2025-05-07
Minor cleanup to keep inline with Dawn's headers.

lokokung:
- Updates sorting to take into account extensions and ignore type name casing.
- Updates some comments to use helpers and cleans up some whitespace.
- Moves WGPU_STRLEN into a constant.

### DIFF
--- a/gen/cheader.tmpl
+++ b/gen/cheader.tmpl
@@ -131,15 +131,9 @@
  * For info on how this is used in various places, see \ref Strings.
  */
 typedef struct WGPUStringView {
-    char const * WGPU_NULLABLE data;
+    WGPU_NULLABLE char const * data;
     size_t length;
-} WGPUStringView;
-
-/**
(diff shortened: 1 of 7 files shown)


gen/cheader.tmpl
```diff
@@ -131,15 +131,9 @@
  * For info on how this is used in various places, see \ref Strings.
  */
 typedef struct WGPUStringView {
-    char const * WGPU_NULLABLE data;
```
kainino0x, line 134: note: I have this here because it's where we want it to be eventually: http://github.com/webgpu-native/webgpu-headers/issues/190
and since it's not autogenerated it won't get fixed along with everything else when we finally fix it.
But I guess it doesn't matter, once we do that we'll actually test it (set the macro and compile test) and presumably we'll find this again.

kainino0x (approved):

PR #548, opened 2025-06-16
Use Undefined for enums when it's explicitly defined.

lokokung:


gen/gen.go
```diff
@@ -684,15 +684,25 @@ func (g *Generator) DefaultValue(member ParameterType, isDocString bool) string
...
+			// If the enum type has an explicit "Undefined" use it, otherwise, use 0.
+			enumType := g.Enums[idx]
+			undefIdx := slices.IndexFunc(enumType.Entries, func(entry *EnumEntry) bool { return entry != nil && entry.Name == "undefined" })
+			if undefIdx == -1 {
+				if isDocString {
+					return "(@ref " + g.CType(member.Type, "") + ")0"
+				} else {
+					return "_wgpu_ENUM_ZERO_INIT(" + g.CType(member.Type, "") + ")"
+				}
 			} else {
```
kainino0x, line 704: We have four enums where `Undefined` is 1 instead of 0 (because 0 is `BindingNotUsed`).

Maybe instead of special casing the name `Undefined` we should special case the value 0?
Or special case only when the name is `Undefined` AND the value is 0?

kainino0x (commented):
SG, one comment then.

gen/gen.go
lokokung: I don't think it should matter what `Undefined` is actually equal? Either we have an explicit default in the yaml for if we don't want `Undefined` to be the default, or whatever the value of `Undefined` is should be the default. 

FWIW, the enums you're referring to, (`WGPUBufferBindingType`, `WGPUSamplerBindingType`, `WGPUTextureSampleType`, and `WGPUStorageTextureAccess`), all already assume their defaults should be `Undefined` when calling their `INIT` explicitly. The only time that they use `BindingNotUsed` instead is when they are being `INIT` via their wrapping `WGPUBindGroupLayoutEntry`, but that `INIT` macro explicitly uses a zero init for the wrapped structs.

Anyways, I don't think the additional condition is necessary? Especially if the name of the enum is `Undefined`.

kainino0x: > FWIW, the enums you're referring to, (`WGPUBufferBindingType`, `WGPUSamplerBindingType`, `WGPUTextureSampleType`, and `WGPUStorageTextureAccess`), all already assume their defaults should be `Undefined` when calling their `INIT` explicitly. The only time that they use `BindingNotUsed` instead is when they are being `INIT` via their wrapping `WGPUBindGroupLayoutEntry`, but that `INIT` macro explicitly uses a zero init for the wrapped structs.

Ahh true. I forgot how this was set up.

kainino0x (approved):
LGTM!

PR #549, opened 2025-06-18
Rename WGPUInstanceFeatureName_TimedWaitAnyEnable -> TimedWaitAny

kainino0x:
`WGPUInstanceFeature_TimedWaitAnyEnable` (formerly `WGPUInstanceCapabilities.timedWaitAnyEnable`) should be called simply `TimedWaitAny` now that it's a feature name.

Aside, it's arguably redundant with `timedWaitAnyMaxCount`. It would be simpler to use if we didn't have a feature, and just set `timedWaitAnyMaxCount` to 0 or not. The reason this wasn't the case originally is because it has a minimum value of 64, so if we did this then:
- 0 -> disabled with a max of 0
- 1 -> enabled with a max of 64

which is slightly odd. I'm inclined not to bother to change this.

### DIFF
--- a/gen/gen.go
+++ b/gen/gen.go
@@ -684,15 +684,25 @@ func (g *Generator) DefaultValue(member ParameterType, isDocString bool) string 
 	// Cases that may have member.Default
 	case strings.HasPrefix(member.Type, "enum."):
 		if member.Default == nil {
-			if member.Type == "enum.optional_bool" {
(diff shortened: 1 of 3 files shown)


lokokung (approved):

kainino0x:
(Going ahead with landing now, but feel free to comment later and I'll follow up)

PR #553, opened 2025-07-07
[breaking] Add new texture formats from texture-formats-tier1

AlbinBernhardssonARM:
Add the new texture formats from texture-formats-tier1 (16 bits per channel unorm/snorm). Maintain the same ordering as in the JS API. This changes the ABI by changing enum values.


kainino0x (approved):
lgtm, thank you!

BTW, I think we should stop changing enum values about now - for the FeatureName I mostly did it so compat would be first, but it's also nice for these to be in a less confusing order. Since it probably shouldn't affect anyone in practice I say let's go ahead with this.

PR #554, opened 2025-07-11
[breaking] Remove return value from wgpuInstanceGetWGSLLanguageFeatures

kainino0x:
This method has no error conditions, so it should return void like wgpuAdapter/DeviceGetFeatures and wgpuGetInstanceFeatures.


lokokung (approved):

kainino0x:
I don't think this will affect anyone other than Dawn but feel free if you have any comments @cwfitzgerald

PR #560, opened 2025-10-09
Add `"texture-component-swizzle"` feature

beaufortfrancois:
This PR adds the `"texture-component-swizzle feature"`  in parallel of https://github.com/gpuweb/gpuweb/pull/5361

### DIFF
--- a/webgpu.h
+++ b/webgpu.h
@@ -508,11 +508,29 @@ typedef enum WGPUComponentSwizzle {
      * `0`. Indicates no value is passed for this argument. See @ref SentinelValues.
      */
     WGPUComponentSwizzle_Undefined = 0x00000000,
+    /**
+     * Force its value to 0.
+     */
     WGPUComponentSwizzle_Zero = 0x00000001,
+    /**
+     * Force its value to 1.
+     */
     WGPUComponentSwizzle_One = 0x00000002,
+    /**
+     * Take its value from the red channel of the texture.
+     */
     WGPUComponentSwizzle_R = 0x00000003,
+    /**
+     * Take its value from the green channel of the texture.
+     */
     WGPUComponentSwizzle_G = 0x00000004,
+    /**
+     * Take its value from the blue channel of the texture.
+     */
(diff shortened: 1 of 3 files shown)


kainino0x:
Thanks. I guess we should still talk about what we want the C API for this to look like since it's not a direct mapping from JS. I'd really like it to be more ergonomic; I'm not sure if there's a good string-based solution but I still don't like the soup of channel identifiers. (I so wish that C 4-character literals like `'rrrr'` were something we could use, but they're not really.)

Or we could just not worry about the ergonomics and leave it as is. Actually, there could be a macro based helper if we wanted to keep this structure and also make it ergonomic. But kinda unnecessary.

```c
typedef enum CS { CS_R, CS_G, CS_B, CS_A, CS_0, CS_1 } CS;
typedef struct TCS { CS r, g, b, a; } TCS;
typedef struct TCSD { TCS swizzle; } TCSD;

#define MAKE_TCS(cr, cg, cb, ca) (TCS){ \
    .r = CS_##cr, .g = CS_##cg, .b = CS_##cb, .a = CS_##ca }

int main() {
    TCSD desc = { .swizzle = MAKE_TCS(B, 0, 1, R) };
}
```

beaufortfrancois:
I would not worry about the ergonomics. The following code looks good to me.
```cpp
  wgpu::TextureViewDescriptor viewDesc{};
  wgpu::TextureComponentSwizzleDescriptor swizzleDesc{};
  swizzleDesc.swizzle.r = wgpu::ComponentSwizzle::R;
  swizzleDesc.swizzle.g = wgpu::ComponentSwizzle::Zero;
  swizzleDesc.swizzle.b = wgpu::ComponentSwizzle::Zero;
  swizzleDesc.swizzle.a = wgpu::ComponentSwizzle::One;
  viewDesc.nextInChain = &swizzleDesc;
  ```

kainino0x (approved):
Sorry I completely missed your revision somehow!

I do want to avoid duplicating docs into this header but this is fine for now, we can resolve that at a later point when we figure out how to link the docs better.

PR #570, opened 2025-11-21
Add stub ExternalTexture and support for them in BindGroups

Kangz:
PTAL! It doesn't add creation of the external texture as that's still super implementation specific, but this will unblock folks that want to use ExternalTexture in WASM (they can import a C object from a precreated JS GPUExternalTexture).

The webgpu.h diff is slightly messy because the new chained struct force a small reordering. The additions themselves should be pretty trivial and they add a chained struct for the equivalent Rust enum option in wgpu.

I chose `ExternalTextureBindingLayout` to match the `Buffer/Sampler/...BindingLayout` convention that's already used for `BindGroupLayoutEntry` members. `ExternalTextureBindingEntry` is just following that. But I'm not to attached to the exact names.

### DIFF
--- a/webgpu.yml
+++ b/webgpu.yml
@@ -1540,7 +1540,7 @@ bitflags:
           The buffer can be used as an Index buffer when doing indexed drawing in a render pipeline.
       - name: vertex
         doc: |
-          The buffer can be used as an Vertex buffer when using a render pipeline.
+          The buffer can be used as a Vertex buffer when using a render pipeline.
       - name: uniform
         doc: |
           The buffer can be bound to a shader as a uniform buffer.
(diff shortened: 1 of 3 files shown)


webgpu.yml
```diff
@@ -2143,6 +2149,21 @@ structs:
...
+    type: extension
+    extends:
+      - bind_group_entry
+    members:
+      - name: external_texture
+        type: object.external_texture
+        doc: |
+          TODO
+  - name: external_texture_binding_layout
+    doc: Chained in @ref WGPUBindGroupLayoutEntry to specify that the corresponding entries in an @ref WGPUBindGroup will contain an @ref WGPUExternalTexture.
```
kainino0x: note, just the existence of this extension in the chain has an effect - it has no fields. Which means for example if it got flattened into the parent struct (in a v2 api) it would have to be special cased. Generally I've tried to avoid this in the past, but I can't remember offhand for sure if this would be the first one or not.

Kangz: Yeah if we flatten it in the future, I'd expect it to become a boolean. It's a bit of a special case but seems fine. Or we can add a boolean that defaults to false and update Dawn to have that boolean as well. WDYT?

cwfitzgerald: Lets try this as-is, that's fine.

PR #571, opened 2025-11-21
Add missing sType for TextureComponentSwizzleDescriptor

Kangz:
PTAL!


beaufortfrancois:
Sorry for the noise, it was actually caught. See https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/third_party/webgpu-headers/webgpu.h.diff;l=23;drc=2b22aaf232b1207dfbe9d0c48144d04299b06219

PR #586, opened 2026-03-19
Fix linking of TextureViewSwizzle to TextureViewDescriptor.

Kangz:
"extends" shouldn't have the "struct." prefix.

Adds a check in the validator to check that the linking works correctly.


kainino0x (approved):
ohh! Nice find, thanks

PR #591, opened 2026-06-24
Add "subgroup-size-control" feature

Jiawei-Shao:
This patch adds the "subgroup-size-control" feature added in gpuweb/gpuweb#5578.

### DIFF
--- a/webgpu.yml
+++ b/webgpu.yml
@@ -520,9 +520,6 @@ enums:
       - name: subgroups
         doc: |
           TODO
-      - name: subgroup_size_control
-        doc: |
-          TODO
       - name: texture_formats_tier_1
         doc: |
           TODO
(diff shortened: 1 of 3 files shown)


Jiawei-Shao:
@Kangz @kainino0x PTAL, thanks!

webgpu.yml
```diff
@@ -520,6 +520,9 @@ enums:
       - name: subgroups
         doc: |
           TODO
+      - name: subgroup_size_control
```
Kangz: Please add it at the end so it doesn't reorder enums. Enum values are supposed to be stable once added to this file.

Kangz (changes requested):

webgpu.yml
Jiawei-Shao: Done

Kangz (approved):

PR #593, opened 2026-07-02
Adding Typedefs for c_void

jalexcole:
I have added some typedefs that can aid code generators. The generation script does not allow for typedefs in the return type at this moment.


webgpu.yml
```diff
@@ -74,7 +74,47 @@ constants:
     doc: |
       Indicates a size extending to the end of the buffer. For more info,
       see @ref SentinelValues and the places that use this sentinel value.
-typedefs: []
+typedefs:
```
kainino0x: Using typedefs means the types will appear in the C header. Do you need that, or do you just need it in the yaml/json? If just the latter, then instead of adding typedefs here we can add aliases in another way (basically handle all of these strings as aliases of `c_void` in generation).

webgpu.yml
kainino0x, line 1: You'll need to run `make` to check the build and regenerate other files.

webgpu.yml
jalexcole: Aliases in the yaml file would be sufficient.

jalexcole:
I am fine with @kainino0x approach.

lokokung (approved):

PR #604, opened 2026-09-22
Add atomic_vec2u_min_max

MendyBerger:
Add `atomic_vec2u_min_max` WGSL lang feature
https://github.com/gpuweb/gpuweb/pull/5425


webgpu.h
```diff
@@ -1259,6 +1259,7 @@ typedef enum WGPUWGSLLanguageFeatureName {
     WGPUWGSLLanguageFeatureName_BufferView = 0x0000000C,
     WGPUWGSLLanguageFeatureName_SwizzleAssignment = 0x0000000D,
     WGPUWGSLLanguageFeatureName_FragmentDepth = 0x0000000E,
+    WGPUWGSLLanguageFeatureName_AtomicVec2uMinMax = 0x0000000F,
```
Kangz, line 1262: This should be in `WGPUFeatureName` instead. This is where it is in the spec (`GPUFeatureName`), because it is not a pure sugar WGSL feature, but an optional HW feature.

